PR #3, opened 2019-04-18
Tekton CLI rocks

sthaha:
This commit adds a basic cli (`tkn`) for tekton. `tkn` currently supports
the following sub commands.

- `pipeline ls`: to list all pipeline CR in a namespace
- `completion bash`:  to generate interactive completion for bash shell
- `help`

Implementation detail:

- `tkn` use cobra as the cli framework.
- uses `ClientSet` provided by `tektoncd/pipeline/pkg/client` to query
  Pipeline resources

# Changes

# Submitter Checklist

review them:

for more details._

# Release Notes

```
release-note
```


pkg/cmd/completion.go
```diff
@@ -0,0 +1,66 @@
...
+	`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+
+		switch args[0] {
+		case "bash":
+			return rootCmd.GenBashCompletion(os.Stdout)
+			// TODO add zsh completion
+			// case "zsh":
+			// return rootCmd.GenZshCompletion(os.Stdout)
+			// NOTE: cobra v0.0.3 zsh completion seems fail on zsh 5.5.1
```
chmouel: openshift client completion has been failing for me too :(

pkg/cmd/completion.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+		return nil
+	},
+}
+
+func exactValidArgs(n int) cobra.PositionalArgs {
```
chmouel: why not https://github.com/spf13/cobra/blob/master/args.go#L84 is this a custom version?  if yes it may be nice to add a comment why,

pkg/cmd/completion.go
sthaha: oc completion zsh works for me tho

pkg/cmd/completion.go
sthaha: that will available only in the next version on in `v0.0.3`

vdemeester:
/ok-to-test

(even though it doesn't nothing yet :joy_cat: )

cmd/tkn/main.go
```diff
@@ -0,0 +1,21 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import "github.com/tektoncd/cli/pkg/cmd"
+
+func main() {
+	cmd.Execute()
```
vdemeester: nit: I would prefer having the only `os.Exit` in here instead of `pkg/cmd` :angel:

pkg/cmd/completion.go
vdemeester: @sthaha I vote for adding this case. If it fails, we'll wait for upstream to update but that shouldn't keep us to provide completion for zsh <= 5.5.1, should it ?

pkg/cmd/pipeline.go
```diff
@@ -0,0 +1,43 @@
...
+	namespace string
+)
+
+// pipelinesCmd represents the pipelines command
+var pipelinesCmd = &cobra.Command{
+	Use:     "pipelines",
+	Aliases: []string{"p", "pipeline"},
+
+	Short: "Handles pipelines",
+	Long:  `add long description`, // TODO(sthaha)
```
vdemeester: Let's add something like `Manage pipelines` for now :wink:

pkg/cmd/pipeline.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func init() {
+	rootCmd.AddCommand(pipelinesCmd)
+
+	pipelinesCmd.PersistentFlags().StringVarP(
```
vdemeester: - Shouldn't the flag be on the root command even ?
- We should have a default namespace value (can be done in a follow-up for sure).

pkg/cmd/pipeline_list.go
```diff
@@ -0,0 +1,86 @@
+// Copyright © 2019 The Knative Authors.
```
vdemeester: I think we will need to update that at some point :sweat_smile: 
cc @bobcatfish

pkg/cmd/root.go
```diff
@@ -0,0 +1,130 @@
...
+	// has an action associated with it:
+	//	Run: func(cmd *cobra.Command, args []string) { },
+}
+
+// Execute adds all child commands to the root command and sets flags appropriately.
+// This is called by main.main(). It only needs to happen once to the rootCmd.
+func Execute() {
+	if err := rootCmd.Execute(); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
```
vdemeester: As stated above, I would prefer have this `os.Exit` (and thus this whole `Execute` method) in `cmd` :angel:

pkg/cmd/root.go
```diff
@@ -0,0 +1,130 @@
...
+	cobra.OnInitialize(initKubeCfgFile)
+	cobra.OnInitialize(initKubeConfig)
+
+	// Here you will define your flags and configuration settings.
+	// Cobra supports persistent flags, which, if defined here,
+	// will be global for your application.
+	rootCmd.PersistentFlags().StringVarP(
+		&cfgFile, "config", "c",
+		"", "config file (default: $HOME/.tekton.yaml)")
+	rootCmd.PersistentFlags().StringVarP(
```
vdemeester: I wonder if we need that yet :sweat_smile:

pkg/cmd/completion.go
sthaha: what I have on my machine is 5.5.1 :( and it fails there. Could we not address this in different patch?

vdemeester: yep we can, we can :wink:

pkg/cmd/pipeline.go
sthaha: yes, my thoughts on this is to wait a bit until we are sure all commands need `-n ` but that may not be case already with `help` and `completion`

I agree we should find a default namespace based on the context but like you said as a separate patch.

pkg/cmd/pipeline_list.go
sthaha: :) 
I modified YOUR NAME <email address> (default) to The Knative Authors so that the same script that changes tektoncd can be run against `cli` as well.

pkg/cmd/root.go
sthaha: wasn't sure either but knative/cli seems to have one and not sure if it is used. I am happy to remove this.

pkg/cmd/pipeline.go
sthaha: > Few comments, and one question : do we want to vendor files or not ? (`go mod vendor`)

Thanks for the review. 
Is my understanding wrong that `go.sum` would ensure that all of us use the same version of dependencies? I am happy to add `vendor` if you think it is better.

cmd/tkn/main.go
sthaha: did you mean something like `os.Exit(cmd.Execute() )`  ?

pkg/cmd/pipeline.go
vdemeester: > Is my understanding wrong that `go.sum` would ensure that all of us use the same version of dependencies? 

This is correct — the only benefit to add vendor are :
- if something goes down you still have the code (this one is a bit *meh* :sweat_smile:)
- you don't need to download the deps to build the binary (less *meh*) and you can build the binary with older version of go (if it make sense) and anywhere (aka in `GOPATH` without having to set `GO111MODULE` env..)

pkg/cmd/root.go
vdemeester: let's remove for now and see if we need it in the future :wink:

pkg/cmd/pipeline.go
sthaha: ack. I will add it in

bobcatfish:
image

PR #17, opened 2019-05-01
Refactor cli

sthaha:
# Changes
#### What does this solve ? ####
Previously, it was hard to test if a command like `list` works fine as
the command was tighly coupled to a concrete instance of Pipeline clientset
rendering the usage of a fake client for testing impossible.

Previous design also had a drawback that commands got added
magically (`func init`) and encouraged use of package level variables
in root command which the sub commands would depend on (leaking the
internals).

#### How is it solved ? ####

The new refactored code makes use of packages to add sub commands e.g.
`pipeline` package and does not make use of any package level
variables. This design also inverts the dependency between parent command
and child command such that the parent now explicitly need to add a
child command. The patch also makes it possible to test commands as the
commands make use of an interface (`Params`) to obtain the `Clientset`

Signed-off-by: sthaha@redhat.com

# Submitter Checklist

review them:

for more details._

# Release Notes
No real changes to user facing side. This only refactors the old implementation

```
release-note
```


pkg/cmd/pipeline/pipeline.go
```diff
@@ -12,38 +12,36 @@
...
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			return flags.InitParams(p, cmd)
+		},
 
-func init() {
-	rootCmd.AddCommand(pipelinesCmd)
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) == 0 {
+				cmd.Help()
+				os.Exit(1)
```
hrishin: How about returning an `error` with an appropriate message rather calling `Exit()`,
if [root cmd](https://github.com/tektoncd/cli/pull/17/files#diff-6ce6b47ced67718ed4842ac0510704c5R31) calling exit function on `error` and `error` returned from this function could be cascade till root cmd `Executions()` ?

pkg/cli/params.go
```diff
@@ -0,0 +1,91 @@
...
+		os.Exit(1)
+	}
+
+	config, err := clientcmd.BuildConfigFromFlags("", kcPath)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
+	}
+
+	cs, err := versioned.NewForConfig(config)
```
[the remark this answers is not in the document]
sthaha: Yes, we could but I don't feel strongly about it since the package is called `versioned` and does not clash with any other imported package names.

pkg/cli/params.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "failed to create client from config %s  %s", p.kubeConfigPath, err)
```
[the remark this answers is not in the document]
sthaha: good catch! thanks :), I will add that

pkg/cmd/pipeline/pipeline.go
```diff
@@ -12,38 +12,36 @@
...
-	Aliases: []string{"p", "pipeline"},
-	Short:   "Manage pipelines",
-	Run: func(cmd *cobra.Command, args []string) {
-		if len(args) == 0 {
-			cmd.Help()
-			os.Exit(0)
-		}
-	},
-}
+		PreRunE: func(cmd *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
sthaha: Yes, for now, that's how I imagine it would be but we definitely can think of making it "dry" when we get to that point.

pkg/cmd/pipeline/pipeline.go
sthaha: +1 .. will do

pkg/cmd/pipeline/list.go
```diff
@@ -0,0 +1,62 @@
...
+	c := &cobra.Command{
+		Use:     "list",
+		Aliases: []string{"ls"},
+		Short:   "Lists pipelines in a namespace",
+		Long:    ``,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ps, err := List(p)
+
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "failed to list pipelines from namespace %s  %s", p.Namespace(), err)
```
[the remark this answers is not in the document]
sthaha: no, it isn't intentional. I am going to change this a bit to return an error to the caller.

pkg/cmd/pipeline/pipeline.go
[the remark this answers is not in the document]
sthaha: Actually this doesn't seem to work as I thought it would. the PreRunE doesn't seem to get called for all subcommands.

pkg/cli/params.go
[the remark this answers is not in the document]
sthaha: Now handled differently by letting the caller deal with the error

pkg/cmd/pipeline/list.go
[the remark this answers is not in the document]
sthaha: same as above; resolving it

pkg/cmd/pipeline/pipeline.go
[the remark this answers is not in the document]
sthaha: Fixed this to use `PersistentPreRunE`

pkg/cmd/pipeline/pipeline.go
sthaha: done!

pkg/cmd/pipeline/list.go
```diff
@@ -0,0 +1,66 @@
...
+			}
+			return printer.PrintObj(ps, cmd.OutOrStdout())
+		},
+	}
+	f.AddFlags(c)
+
+	return c
+}
+
+func List(p cli.Params) (*v1alpha1.PipelineList, error) {
```
vdemeester: Any reason for being exported as of today ?

sthaha: :thinking: .. no good reason to. I will make that internal.

pkg/flags/flags.go
```diff
@@ -0,0 +1,63 @@
...
+
+const (
+	kubeConfig = "kubeconfig"
+	namespace  = "namespace"
+)
+
+// AddTektonOptions amends command to add flags required to initialise a cli.Param
+func AddTektonOptions(cmd *cobra.Command) {
+	cmd.PersistentFlags().StringP(
+		kubeConfig, "k", "",
```
chmouel: Can we have the long form option?

pkg/flags/flags.go
```diff
@@ -0,0 +1,63 @@
...
+		"kubectl config file (default: $HOME/.kube/config)")
+
+	cmd.PersistentFlags().StringP(
+		namespace, "n", "",
+		"namespace to use (mandatory)")
+
+	cmd.MarkPersistentFlagRequired(namespace)
```
chmouel: Missing an error check ?

pkg/flags/flags.go
sthaha: we do, don't we? `--kubeconfig `

pkg/cmd/pipeline/list.go
```diff
@@ -39,7 +39,7 @@ func ListCommand(p cli.Params) *cobra.Command {
 			ps, err := List(p)
 
 			if err != nil {
-				fmt.Fprintf(os.Stderr, "failed to list pipelines from namespace %s  %s", p.Namespace(), err)
+				fmt.Fprintf(os.Stderr, "Failed to list pipelines from %s namespace\n", p.Namespace())
```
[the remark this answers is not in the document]
sthaha: I tried that and the output isn't as nice as this

pkg/flags/flags.go
sthaha:
```
Flags:
  -h, --help                help for pipeline
  -n, --namespace string    namespace to use (mandatory)

```

pkg/flags/flags.go
sthaha: no, seems fine to me. or may be i didn't follow :thinking:  ... Could you please elaborate?

pkg/cmd/pipeline/list.go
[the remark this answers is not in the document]
sthaha: Since the error is returned, we would first see 
```
Failed to list pipelines from foobar 
<reason>

<help>
```

pkg/flags/flags.go
chmouel: ah nice, i misteaded from the documentation

pkg/flags/flags.go
vdemeester: (I think it's because `cmd.MarkPersistentFlagRequired` does return an `error` ?)

pkg/cmd/pipeline/list.go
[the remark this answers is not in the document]
chmouel: it's what tektoncd/pipelines is going for so I think we may want to keep track of how it's done there?

Issue #4, opened 2019-04-23
List pipelines


siamaksade:
Add a command to list pipelines. For each pipeline, the following details should be displayed in addition to the pipeline details:
* Last pipeline-run name
* Last pipeline-run time started
* Last pipeline-run duration
* Last pipeline-run status

vdemeester:
#20 is merge, we can close this one
/close

PR #20, opened 2019-05-10
Show last pipeline-run details of pipeline

sthaha:
# Changes
Listing pipelines shows additional details about latest run

Earlier, `tkn pipeline list -n foobar` only used to list the names of the
pipelines in the namespace `foobar`. This patch changes the behaviour such
that it also lists details of latest pipeline-run of each pipeline

E.g. Given

```
$ ./tkn pipelineruns ls -n test-tekton

NAME                    STARTED   DURATION     STATUS
output-pipeline-run-2   1 day     26 seconds   Succeeded
output-pipeline-run-3   1 day     26 seconds   Succeeded
output-pipeline-run-5   1 day     27 seconds   Succeeded
```

It now shows

```
$ ./tkn pipeline ls -n test-tekton

NAME              AGE         LAST RUN                STARTED      DURATION     STATUS
output-pipeline   1 day ago   output-pipeline-run-5   1 day ago    27 seconds   Succeeded
```

Changes include:
* Listing pipelines displays details about latest run
* pipeline list prints latest run as well
* pipeline: make listCommand internal
* pipeline: add test for various scenarios
* Improve how Condition is formatted
* Fix PipelineRun tests that broke due to refactor
* Remove dummy_test since there are real tests now

Issue: https://github.com/tektoncd/cli/issues/4
Signed-off-by: sthaha@redhat.com

# Submitter Checklist

review them:

for more details._

# Release Notes

```
Listing pipelines now displays additional details about the latest PipelineRun
```


sthaha:
@vdemeester @hrishin Please note that I have set the PR to `WIP` only because I think i need to improve the commit messages. Otherwise, I am happy with the patch.

pkg/cmd/pipeline/list.go
```diff
@@ -16,51 +16,172 @@ package pipeline
...
+		return err
+	}
+
+	if len(ps.Items) == 0 {
+		fmt.Fprintln(out, emptyMsg)
+		return nil
+	}
+
+	w := tabwriter.NewWriter(out, 0, 5, 3, ' ', tabwriter.TabIndent)
+	defer w.Flush()
```
hrishin: just wondering what would happen if w.Flush() returns an `error`? :thinking:

pkg/cmd/pipeline/list.go
```diff
@@ -16,51 +16,172 @@ package pipeline
...
+			)
+			continue
+		}
+
+		fmt.Fprintf(w, body,
+			pipeline.Name,
+			formatted.Age(pipeline.CreationTimestamp, p.Time()),
+			pr.Name,
+			formatted.Age(*pr.Status.StartTime, p.Time()),
+			formatted.Duration(pr.Status.StartTime, pr.Status.CompletionTime),
```
siamaksade: Duration should say `N ago` instead of just `N`

pkg/cmd/pipeline/list.go
```diff
@@ -16,51 +16,172 @@ package pipeline
...
+
+	fmt.Fprintln(w, header)
+
+	for _, pipeline := range ps.Items {
+		pr, ok := prs[pipeline.Name]
+		if !ok {
+			fmt.Fprintf(w, body,
+				pipeline.Name,
+				formatted.Age(pipeline.CreationTimestamp, p.Time()),
+				"---",
```
hrishin: Maybe we could extract this in constants as well?

pkg/cmd/pipeline/list.go
```diff
@@ -16,51 +16,172 @@ package pipeline
...
+	if len(ps.Items) == 0 {
+		return ps, pipelineruns{}, nil
+	}
+
+	runs, err := cs.TektonV1alpha1().
+		PipelineRuns(p.Namespace()).
+		List(metav1.ListOptions{})
+
+	if err != nil {
+		// TODO: may be just the pipeline details can be print
```
hrishin: good point  :+1: (may be an additional flag could help here?)

pkg/cmd/pipeline/list.go
```diff
@@ -16,51 +16,172 @@ package pipeline
...
+		return nil, nil, err
+	}
+
+	latestRuns := pipelineruns{}
+
+	for _, run := range runs.Items {
+		pipelineName := run.Spec.PipelineRef.Name
+		x, ok := latestRuns[pipelineName]
```
hrishin: How about `r` instead of `x`?

pkg/cmd/pipeline/list.go
sthaha: duration shouldn't say that, start can ... duration is the "time taken" and not how long back.

pkg/cmd/pipeline/list.go
sthaha: :), imho .. it makes no difference, isn't it ?

pkg/cmd/pipeline/list.go
sthaha: if contents can be written to `out`, then isn't it already too late?

pkg/cmd/pipeline/list.go
sthaha: yes, I can although, `"---"` says better than `empty`

pkg/cmd/pipeline/list.go
sthaha: may be better  to return a `specific error` and then check for that and just print the `pipelines`. Can submit that in a different PR.

pkg/cmd/pipeline/list_test.go
```diff
@@ -3,38 +3,205 @@ package pipeline
...
 	if err != nil {
 		t.Errorf("Unexpected error: %v", err)
 	}
 
-	//  account for new line at the end
-	expected := strings.Join(append(names, ""), "\n")
-	if d := cmp.Diff(expected, output); d != "" {
+	expected := []string{
+		"NAME       AGE          LAST RUN   STARTED   DURATION   STATUS",
+		"tomatoes   1 minute     ---        ---       ---        ---",
```
sthaha: @siamaksade , do you feel it would look better with `ago`? 

e.g.

```
$ tkn pipeline ls -n test-tekton

NAME              AGE         LAST RUN                STARTED     DURATION     STATUS
output-pipeline   2 days ago   output-pipeline-run-5   1 day ago   27 seconds   Succeeded
```

pkg/cmd/pipeline/list.go
hrishin: yes however in content writing is fails without printing anything on `out` then it should return an error? so that command would not fail silently?

e.g.
```go
func TestShouldReturnError(t *testing.T) {
		err := hello()
		if  err == nil {
			t.Errorf("error is not returned")
		} 
}

func hello() error {
	defer func() error {
		return errors.New("asds") // escape the error	
	}()	

	return nil
}
```

sthaha: I am not sure if I agree with you on this ...
`Flush` flushes contents to `out` and now if it is unable to `flush` i.e. some issues with out that caused a `panic`; the chances of recovery is low ... 

Have you read : https://github.com/golang/go/blob/eda3401e807be8928eed48bb5fc85ffa8e62ddb4/src/text/tabwriter/tabwriter.go#L496  ?

pkg/cmd/pipeline/list.go
```diff
@@ -16,51 +16,171 @@ package pipeline
...
+func printPipelineListObj(w io.Writer, p cli.Params, f *cliopts.PrintFlags) error {
+	ps, err := listAllPipelines(p)
+
+	if err != nil {
+		return err
+	}
+	return printer.PrintObject(w, ps, f)
+}
+
+func listAllPipelines(p cli.Params) (*v1alpha1.PipelineList, error) {
```
vdemeester: nit: for the future, those "get data from client" shouldn't need the `cli.Params` I feel (here we just need the clients and the namespace)

pkg/cmd/pipeline/list.go
```diff
@@ -16,51 +16,171 @@ package pipeline
...
+		schema.GroupVersionKind{
+			Version: "tekton.dev/v1alpha1",
+			Kind:    "PipelineList",
+		})
+	return ps, nil
+}
+
+type pipelineruns map[string]*v1alpha1.PipelineRun
+
+func listPipelineDetails(p cli.Params) (*v1alpha1.PipelineList, pipelineruns, error) {
```
vdemeester: nit: same here

pkg/testutil/tekton.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"time"
+
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
+	tb "github.com/tektoncd/pipeline/test/builder"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// NOTE: methods in this file must be submitted upstream
```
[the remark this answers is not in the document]
sthaha: yeah, found it really well designed :heart:
hmm .. should have used `should` instead of `must` :D . I will change it to should

pkg/cmd/pipeline/list.go
```diff
@@ -16,51 +16,171 @@ package pipeline
...
+		List(metav1.ListOptions{})
+
+	if err != nil {
+		// TODO: may be just the pipeline details can be print
+		return nil, nil, err
+	}
+
+	latestRuns := pipelineruns{}
+
+	for _, run := range runs.Items {
```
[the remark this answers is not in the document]
sthaha: Could you elaborate please? How will the value in the latestRun change? What I am trying to store is the `pipeline name` and the `pointer` to the latest run. How will the `PipelineRun` Object (the pointer of which is stored in map) change?

pkg/cmd/pipeline/list_test.go
```diff
@@ -3,38 +3,205 @@ package pipeline
...
+		runDuration     = 5 * time.Minute
+
+		firstRunStarted   = clock.Now().Add(10 * time.Minute)
+		firstRunCompleted = firstRunStarted.Add(runDuration)
+
+		secondRunStarted   = firstRunCompleted.Add(5 * time.Minute)
+		secondRunCompleted = secondRunStarted.Add(runDuration)
+	)
+
+	cs, _ := test.SeedTestData(test.Data{
```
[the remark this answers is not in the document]
sthaha: that is done in 137, isn't it?

pkg/cmd/pipeline/list_test.go
```diff
@@ -3,38 +3,205 @@ package pipeline
...
+		Pipelines: []*v1alpha1.Pipeline{
+			tb.Pipeline("pipeline", "ns",
+				// created  5 minutes back
+				testutil.PipelineCreationTimestamp(pipelineCreated),
```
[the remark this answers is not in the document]
sthaha: Well, both pipeline's age  (based on creation time) and the lastest pipeline-runs's start-time is shown in the output. 
Are you suggesting to remove the "age" of the `pipeline` ? IMHO, I find it useful.

pkg/formatted/k8s.go
```diff
@@ -0,0 +1,18 @@
+package formatted
+
+import (
+	"github.com/knative/pkg/apis"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func Condition(c apis.Condition) string {
```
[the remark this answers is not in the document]
sthaha: :hugs:

pkg/cmd/pipeline/list.go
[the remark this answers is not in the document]
sthaha: Items is defined here: https://github.com/tektoncd/cli/blob/2a58e089fe3942efe56d719bef730f5f09dc2a46/vendor/github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1/pipelinerun_types.go#L198:2 as `[] PipelineRun` and not `[] *PipelineRun` 

So the map is storing a `pointer` to the `object` and not a `pointer to a pointer`

pkg/cmd/pipeline/list_test.go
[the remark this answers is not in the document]
vdemeester: Yep my bad :sweat_smile:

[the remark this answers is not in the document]
vdemeester: not for the `PipelineRun` but yeah, this isn't where I should have commented :angel:

pkg/cmd/pipeline/list_test.go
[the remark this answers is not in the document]
vdemeester: @sthaha right, we need it for the `Age`, my bad… So we care for both :angel:

pkg/cmd/pipeline/list.go
sthaha: Do you have reason `why` it is better to do it that way i.e pass in `client, ns` opposed to `cli.Params` ? 
Also, I don't have a strong opinion why `p cli.Param` should be passed, so I will make that change in this PR itself.

pkg/cmd/pipeline/list_test.go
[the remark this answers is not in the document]
sthaha: Actually, we do .. as in the test does since the "age" of the pipeline is shown in the output.

pkg/formatted/k8s.go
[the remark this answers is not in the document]
sthaha: thank you :) :hugs:

pkg/cmd/pipeline/list.go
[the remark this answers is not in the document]
vdemeester: Using the pointer to a `loop` variable is really really error-prone (the variable is re-assigned and thus the pointer **always** refers to the current value). A simple example of what it is not working :  https://play.golang.org/p/Pt0vJj81xZU.

pkg/cmd/pipeline/list_test.go
```diff
@@ -3,38 +3,205 @@ package pipeline
...
+			),
+		},
+		PipelineRuns: []*v1alpha1.PipelineRun{
+
+			tb.PipelineRun("pipeline-run-1", "ns",
```
vdemeester: This PipelineRun needs a creation date.

pkg/cmd/pipeline/list_test.go
```diff
@@ -3,38 +3,205 @@ package pipeline
...
+				tb.PipelineRunStatus(
+					tb.PipelineRunStatusCondition(apis.Condition{
+						Status: corev1.ConditionTrue,
+						Reason: resources.ReasonSucceeded,
+					}),
+					tb.PipelineRunStartTime(firstRunStarted),
+					testutil.PipelineRunCompletionTime(firstRunCompleted),
+				),
+			),
+			tb.PipelineRun("pipeline-run-2", "ns",
```
vdemeester: This PipelineRun needs a creation date too :angel: 
(and if you put a creation date "before" `pipeline-run-1` you'll see it succeeds although `pipeline-run-1` is the *last run* — and thus should have failed)

pkg/cmd/pipeline/list_test.go
```diff
@@ -3,38 +3,205 @@ package pipeline
...
 		t.Errorf("Unexpected output mismatch: %s", d)
 	}
 }
+
+func TestPipelinesList_with_single_run(t *testing.T) {
+	clock := clockwork.NewFakeClock()
+
+	cs, _ := test.SeedTestData(test.Data{
+		Pipelines: []*v1alpha1.Pipeline{
+			tb.Pipeline("pipeline", "ns",
```
vdemeester: Shouldn't this use `seedPipelines` function ?

pkg/cmd/pipeline/list_test.go
vdemeester: See https://github.com/tektoncd/cli/compare/master...vdemeester:tasks#diff-acd75ed12ce8066b2e1a5f9a1f25c427R150

pkg/testutil/tekton.go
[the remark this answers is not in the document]
vdemeester: Upstream PR with them is merged :wink:

pkg/cmd/pipeline/list.go
vdemeester: As of today no *important* reason **but** we may/will want to list pipelines in multiple case. This function might get extracted elsewhere *and* in that case, it shouldn't be tied to any `cli` code.
I tend to like having the `cli` related code constrained to only a really small subset of packages (aka, in our case `pkg/cmd`).

But it's a nit, no worries for that PR :wink:

sthaha: I have implemented the change anyway to see how decoupling feels like.

pkg/cmd/pipeline/list.go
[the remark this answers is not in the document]
sthaha: omg! :open_mouth: 

I will copy the object :D  ... I did find that you can set ` latest[name] = &run.Items[index] ` instead but seems too hackish.

pkg/cmd/pipeline/list_test.go
sthaha: no, seedPipelines seeds only the pipelines and not any pipeline-runs which is why I kept all pipeline and runs in one place.

pkg/cmd/pipeline/list_test.go
sthaha: Should we sort by creationTimestamp or the `Status.StartTime` ?

pkg/cmd/pipeline/list_test.go
sthaha: yes, I found that when I fixed the `listPipelineDetails` function. Do you think sorting by `Status.StartTime` instead of creationTimestamp better / expected behaviour ?

pkg/cmd/pipeline/list.go
sthaha: I am marking it as resolved. Please reopen if you disagree

pkg/cmd/pipeline/list.go
sthaha: I am marking it as resolved. Please reopen if you disagree

pkg/cmd/pipeline/list.go
sthaha: Lets go ahead with the current implementation and modify it when we really feel the need. 

I am marking it as resolved. Please reopen if you disagree

pkg/cmd/pipeline/list.go
```diff
@@ -16,51 +16,169 @@ package pipeline
...
+		pipelineName := run.Spec.PipelineRef.Name
+		r, ok := latestRuns[pipelineName]
+		if !ok {
+			latestRuns[pipelineName] = run
+			continue
+		}
+		if run.Status.StartTime.After(r.Status.StartTime.Time) {
```
sthaha: @vdemeester does this make more sense than `creationTimestamp` ?

vdemeester: :thinking: I tend to prefer `creationTimestamp` — for me *last run* means the last one that got created (not the last one that started), but this might "change" in the future, we'll see :wink:)

vdemeester:
@sthaha can you squash some commits ? (there is too much commits for my little eyes :stuck_out_tongue_closed_eyes:)

pkg/cmd/pipeline/list.go
sthaha: cool, will change to use creationTimestamp

sthaha:
> @sthaha can you squash some commits ? (there is too much commits for my little eyes )

Agreed :) 
This is my workflow. I make a commit per review comment and squash those once all review comments are addressed/settled. This helps me fix the review comments as atomic commits, revert them without rewriting the git history. E.g.

 I can now `git revert` sorting by `status.StartTime` and use the `creationTimestamp`.

pkg/cmd/pipeline/list_test.go
sthaha: Fixed ...

Issue #5, opened 2019-04-23
Display pipeline-run logs


siamaksade:
Add a command to display pipeline runs for a specified pipeline. 

**Inputs (Required):** pipeline-run name
**Input (Optional):** task name
**Output:** show logs from the pipeline-run. If the task name is specified, then only logs from that specific task are displayed

It should also be possible to follow logs in real time

PR #24, opened 2019-05-14
Shows pipelinerun logs

hrishin:
# Changes
Adds subcommand, to show the pipelinerun logs

```
09:26 $ ./tkn pr logs output-pipeline-run -n test
[first-create-file : git-source-skaffold-git-qn42p] {"level":"warn","ts":1557744960.2213447,"logger":"fallback-logger","caller":"logging/config.go:65","msg":"Fetch GitHub commit ID from kodata failed: \"ref: refs/heads/master\" is not a valid GitHub commit ID"}
[first-create-file : git-source-skaffold-git-qn42p] {"level":"info","ts":1557744966.9072747,"logger":"fallback-logger","caller":"git/git.go:105","msg":"Successfully cloned https://github.com/GoogleContainerTools/skaffold @ master in path /workspace/damnworkspace"}

.....................
```

- pipelinerun log command refactoring:
  * extract logging pieces into pipelinerun, taskrun and pod
    packages
- param interface refactoring:
  * Clientset()(Clients()) returns both tekton and kubernets
    in one object
  * Replaces clientset() usage accross all commands

# Submitter Checklist

review them:

# Release Notes

```
To fetch the logs from pipelinerun "foo-pipelinerun-1" from the namespace "bar"
./tkn pr logs <pipeline-run-name> -n <namepsace>
```

### DIFF
--- a/pkg/cli/interface.go
+++ b/pkg/cli/interface.go
@@ -17,14 +17,20 @@ package cli
 import (
 	"github.com/jonboulle/clockwork"
 	"github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
+	k8s "k8s.io/client-go/kubernetes"
 )
+
+type Clients struct {
+	Tekton versioned.Interface
+	Kube   k8s.Interface
+}
 
 // Params interface provides
 type Params interface {
-	// SetKubeConfigPath uses the kubeconfig path to instantiate clientset
+	// SetKubeConfigPath uses the kubeconfig path to instantiate tekton
 	// returned by Clientset function
 	SetKubeConfigPath(string)
-	Clientset() (versioned.Interface, error)
+	Clients() (*Clients, error)
 
 	// SetNamespace can be used to store the namespace parameter that is needed
 	// by most commands
--- a/pkg/cli/params.go
+++ b/pkg/cli/params.go
@@ -15,21 +15,27 @@
 package cli
 
 import (
+	"k8s.io/client-go/rest"
 	"os"
 	"path/filepath"
 
 	"github.com/jonboulle/clockwork"
-	homedir "github.com/mitchellh/go-homedir"
+	"github.com/mitchellh/go-homedir"
 	"github.com/pkg/errors"
 	"github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
+	k8s "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
 )
 
-type TektonParams struct {
-	kubeConfigPath string
-	clientset      *versioned.Clientset
+const (
+	DEFAUL_CONFIG_DIR   = ".kube"
+	DEFAULT_CONFIG_FILE = "config"
+)
 
-	namespace string
+type TektonParams struct {
+	clients        *Clients
+	kubeConfigPath string
+	namespace      string
 }
 
 // ensure that TektonParams complies with cli.Params interface
@@ -39,45 +45,81 @@ func (p *TektonParams) SetKubeConfigPath(path string) {
 	p.kubeConfigPath = path
 }
 
-func (p *TektonParams) Clientset() (versioned.Interface, error) {
-	if p.clientset != nil {
-		return p.clientset, nil
+func (p *TektonParams) tektonClient(config *rest.Config) (versioned.Interface, error) {
+	cs, err := versioned.NewForConfig(config)
+	if err != nil {
+		return nil, err
 	}
 
-	kcPath, err := resolveKubeConfigPath(p.kubeConfigPath)
+	return cs, nil
+}
+
+func (p *TektonParams) kubeClient(config *rest.Config) (k8s.Interface, error) {
+	k8scs, err := k8s.NewForConfig(config)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to resolve kubeconfig path")
+		return nil, errors.Wrapf(err, "Failed to create ks8 client from config")
+	}
+
+	return k8scs, nil
+}
+
+func (p *TektonParams) Clients() (*Clients, error) {
+	if p.clients != nil {
+		return p.clients, nil
+	}
+
+	config, err := p.config()
+	if err != nil {
+		return nil, err
+	}
+
+	tekton, err := p.tektonClient(config)
+	if err != nil {
+		return nil, err
+	}
+
+	kube, err := p.kubeClient(config)
+	if err != nil {
+		return nil, err
+	}
+
+	p.clients = &Clients{
+		Tekton: tekton,
+		Kube:   kube,
+	}
+
+	return p.clients, nil
+}
+
+func (p *TektonParams) config() (*rest.Config, error) {
+	kcPath, err := p.resolveKubeConfigPath()
+
+	if err != nil {
+		return nil, errors.Wrap(err, "Failed to resolve kubeconfig path")
 	}
 
 	config, err := clientcmd.BuildConfigFromFlags("", kcPath)
 	if err != nil {
 		return nil, errors.Wrapf(err, "Parsing kubeconfig failed")
 	}
-
-	cs, err := versioned.NewForConfig(config)
-	if err != nil {
-		return nil, err
-	}
-
-	p.clientset = cs
-	return p.clientset, nil
+	return config, err
 }
 
-func resolveKubeConfigPath(kcPath string) (string, error) {
-	if kcPath != "" {
-		return kcPath, nil
+func (p *TektonParams) resolveKubeConfigPath() (string, error) {
+	if p.kubeConfigPath != "" {
+		return p.kubeConfigPath, nil
 	}
 
 	if kubeEnvConf, ok := os.LookupEnv("KUBECONFIG"); ok {
 		return kubeEnvConf, nil
 	}
 
-	// deduce from ~/.kube/config
 	home, err := homedir.Dir()
 	if err != nil {
 		return "", err
 	}
-	return filepath.Join(home, ".kube", "config"), nil
+
+	return filepath.Join(home, DEFAUL_CONFIG_DIR, DEFAULT_CONFIG_FILE), nil
 }
 
 func (p *TektonParams) SetNamespace(ns string) {
@@ -87,7 +129,6 @@ func (p *TektonParams) SetNamespace(ns string) {
 func (p *TektonParams) Namespace() string {
 	return p.namespace
 }
-
 
 func (p *TektonParams) Time() clockwork.Clock {
 	return clockwork.NewRealClock()
--- a/pkg/cmd/pipelinerun/log_test.go
+++ b/pkg/cmd/pipelinerun/log_test.go
@@ -1,0 +1,409 @@
+// Copyright © 2019 The Tekton Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pipelinerun
+
+import (
+	"bytes"
+	"github.com/tektoncd/cli/pkg/cmd/taskrun"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/jonboulle/clockwork"
+	"github.com/knative/pkg/apis"
+	"github.com/tektoncd/cli/pkg/cli"
+	"github.com/tektoncd/cli/pkg/logs"
+	tu "github.com/tektoncd/cli/pkg/test"
+	cb "github.com/tektoncd/cli/pkg/test/builder"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
+	"github.com/tektoncd/pipeline/pkg/reconciler/v1alpha1/pipelinerun/resources"
+	"github.com/tektoncd/pipeline/test"
+	tb "github.com/tektoncd/pipeline/test/builder"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func Test_command_has_pipelinerun_arg(t *testing.T) {
+	c := Command(&tu.Params{})
+
+	_, err := tu.ExecuteCommand(c, "logs", "-n", "ns")
+
+	if err == nil {
+		t.Error("Expecting an error but it's empty")
+	}
+}
+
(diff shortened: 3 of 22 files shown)


pkg/cli/interface.go
```diff
@@ -17,14 +17,22 @@ package cli
...
+	Tekton versioned.Interface
+	Kube   k8s.Interface
+}
+
 // Params interface provides
 type Params interface {
 	// SetKubeConfigPath uses the kubeconfig path to instantiate clientset
 	// returned by Clientset function
 	SetKubeConfigPath(string)
 	Clientset() (versioned.Interface, error)
```
vdemeester: "there can be only one" :sweat_smile: As `Clients` holds both `versions.Interface` and `k8s.Interface`, `Clientset()` and `KubeClient()` should go away :angel:

pkg/cli/params.go
```diff
@@ -22,12 +22,15 @@ import (
 	homedir "github.com/mitchellh/go-homedir"
 	"github.com/pkg/errors"
 	"github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
+	k8s "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
 )
 
 type TektonParams struct {
 	kubeConfigPath string
 	clientset      *versioned.Clientset
```
vdemeester: Same here :angel:

pkg/cli/params.go
```diff
@@ -63,6 +66,53 @@ func (p *TektonParams) Clientset() (versioned.Interface, error) {
...
+	k8scs, err := k8s.NewForConfig(config)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to create ks8 client from config %s", err)
+	}
+
+	p.kubeClient = k8scs
+	return p.kubeClient, nil
+}
+
+func (p *TektonParams) Clients() (*Clients, error) {
```
vdemeester: There is too much duplicated code here, no need to do the `resolveKubeConfigPath` & co multplie times :angel: If you go for having only `Clients`, you can do something like the following

```go
func (p *TektonParams) Clients() (*Clients, error) {
	if p.clients != nil {
		return p.clients, nil
	}
	kcPath, err := resolveKubeConfigPath(p.kubeConfigPath)
	if err != nil {
		return nil, errors.Wrap(err, "failed to resolve kubeconfig path")
	}
	config, err := clientcmd.BuildConfigFromFlags("", kcPath)
	if err != nil {
		return nil, errors.Wrapf(err, "Parsing kubeconfig failed")
	}
	cs, err := versioned.NewForConfig(config)
	if err != nil {
		return nil, err
	}
	kcs, err := corev1.NewForConfig(config)
	if err != nil {
		return nil, err
	}
	clients := &Client{
		Tekton: cs,
		Kube:   kcs,
	}
	p.clients = clients
	return p.clients, nil
}
```

pkg/cmd/pipelinerun/logs.go
```diff
@@ -0,0 +1,269 @@
...
+)
+
+type taskRunMap map[string]*v1alpha1.PipelineRunTaskRunStatus
+
+type taskRun struct {
+	name   string
+	status v1alpha1.TaskRunStatus
+}
+
+type PipelineRunLogs struct {
```
vdemeester: I think I would split this struct into several pieces : the options (initsteps, follow, …), the i/o (out, err) and the client (on it's own)

pkg/cmd/pipelinerun/logs.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+
+	c.Flags().BoolVarP(&follow, "follow", "f", false, "stream the live pipeline run logs")
+	c.Flags().BoolVarP(&allLogs, "all", "a", false, "show all logs including init steps injected by tekton")
+	c.Flags().StringSliceVarP(&tasks, "only-task", "t", []string{}, "show the logs for mentioned task only")
+
+	return c
+}
+
+func (p *PipelineRunLogs) Fetch() error {
```
vdemeester: I would make that a *standard* func that takes the struct described above as parameters, sthg like

```
func fetchLogs(cs *cli.Clients, opts logOptions, io ioStruct) error {
```

pkg/cmd/pipelinerun/pipelinerun.go
```diff
@@ -26,6 +26,7 @@ import (
 func Command(p cli.Params) *cobra.Command {
 	c := &cobra.Command{
 		Use:     "pipelineruns",
+		DisableFlagsInUseLine: true,
```
vdemeester: Why is it required ? :thinking:

pkg/cmd/root.go
```diff
@@ -26,8 +26,7 @@ func Root(p cli.Params) *cobra.Command {
 	var cmd = &cobra.Command{
 		Use:   "tkn",
 		Short: "CLI for tekton pipelines",
-		Long: `
-	`,
+		Long:  ``,
```
vdemeester: nit for future: we need to add descriptions here :angel:

pkg/cli/interface.go
hrishin: yeah, I was also wondering. just want to confirm, is there any case where it would be good to create k8s client only? otherwise, let me make those changes

vdemeester: I don’t think there is any case :) and if it turns out there is, we can think about it then :wink:

pkg/cmd/pipelinerun/pipelinerun.go
hrishin: To disable the additional flag note in the usage of `tkn pr -h`.
It doesn't help the user

pkg/cli/interface.go
sthaha: Agree it is better to Clientset of return `*Client` which holds all clients.

pkg/cmd/pipelinerun/pipelinerun.go
sthaha: FYI: There are 2 other settings that I want to add to `root` command 
`SilenceUsage, SilenceErrors`  so that the usage isn't displayed unless there is good reason to.

pkg/cli/interface.go
hrishin: resolved :bowing_man: , please confirm if its the right

pkg/cli/params.go
hrishin: resolved :bowing_man:

pkg/cli/params.go
hrishin: :sweat_smile: resolved

pkg/cmd/pipelinerun/logs.go
hrishin: Definitely :+1: :metal:

pkg/cli/interface.go
```diff
@@ -17,14 +17,20 @@ package cli
...
+	Kube   k8s.Interface
+}
+
 // Params interface provides
 type Params interface {
-	// SetKubeConfigPath uses the kubeconfig path to instantiate clientset
+	// SetKubeConfigPath uses the kubeconfig path to instantiate tekton
 	// returned by Clientset function
 	SetKubeConfigPath(string)
-	Clientset() (versioned.Interface, error)
```
hrishin: I wonder if this change could have gone into a different PR? :thinking:

vdemeester: That's fine here :wink:

pkg/cli/params.go
```diff
@@ -53,31 +85,47 @@ func (p *TektonParams) Clientset() (versioned.Interface, error) {
...
+	return config, err
+}
 
-	cs, err := versioned.NewForConfig(config)
+func (p *TektonParams) Clients() (*Clients, error) {
+	if p.clients != nil {
+		return p.clients, nil
+	}
+
+	tekton, err := p.tektonClient()
```
vdemeester: I don't think we need to duplicate the `p.config()` call.

```go
func (p *TektonParams) Clients() (*Clients, error) {
	if p.clients != nil {
		return p.clients, nil
	}
	config, err := p.config()
	if err != nil {
		return nil, err
	}
	tekton, err := p.tektonClient(config)
	// […]
	kube, err := p.kubeClient()
	// […]
}
```

pkg/cmd/pipelinerun/log_test.go
```diff
@@ -0,0 +1,251 @@
...
+	expected := msgPRNotFoundErr + " : pipelineruns.tekton.dev \"output-pipeline-2\" not found \n"
+
+	if d := cmp.Diff(expected, got); d != "" {
+		t.Errorf("Unexpected output mismatch: \n%s\n", d)
+	}
+}
+
+func Test_print_pipelinerun_logs(t *testing.T) {
+	var (
+		pipelineName = "output-pipeline"
```
vdemeester: Same comment as the other PR, we can inline those :wink:

pkg/cmd/pipelinerun/log_test.go
```diff
@@ -0,0 +1,251 @@
...
+			logs.Step(tr2Step1Name, taskrun.ContainerNameForStep(tr2Step1Name),
+				"able to read a file",
+			),
+			logs.Step("nop", "nop",
+				"Build successful",
+			),
+		),
+	)
+
+	t.Run("all task logs", func(t *testing.T) {
```
vdemeester: We should do a table test here to reduce the duplication

```go
tests := []struct{
	name string
	follow bool
	allSteps bool
	expectedLogs []string
}
for _, td := range tests {
	t.Run(td.name, func(t *testing.T) {
		// […]
	})
}
```

pkg/cmd/pipelinerun/logs.go
```diff
@@ -0,0 +1,220 @@
...
+		Ns:      namespace,
+		Clients: clients,
+	}
+}
+
+func logCommand(p cli.Params) *cobra.Command {
+	var (
+		follow    bool
+		allSteps  bool
+		onlyTasks []string
```
vdemeester: Shouldn't this just be `tasks`, as the tasks I want to show the log for ? (and if it's empty, it means all)

pkg/cmd/taskrun/logs.go
```diff
@@ -0,0 +1,110 @@
+package taskrun
+
+import (
+	"fmt"
+	"github.com/tektoncd/cli/pkg/cli"
```
vdemeester: nit: needs an empty line :angel:

pkg/cmd/taskrun/logs.go
```diff
@@ -0,0 +1,110 @@
...
+	return stepState.ContainerState.Waiting == nil
+}
+
+//TODO: annonymous steps?
+func ContainerNameForStep(name string) string {
+	switch name {
+	case "nop":
+		return "nop"
+	default:
+		return "build-step-" + name
```
vdemeester: I'm not sure if we need to prepend the `step` names with `build-step`, it feels weird :sweat_smile:

pkg/logs/fakelog.go
```diff
@@ -0,0 +1,86 @@
+package logs
```
vdemeester: As it is only for testing purpose, should this file under a package in `pkg/test` ? :thinking:

pkg/test/builder/taskrun.go
```diff
@@ -14,3 +15,23 @@ func TaskRunCompletionTime(ct time.Time) tb.TaskRunStatusOp {
...
 	}
 }
+
+// StepState adds a StepState to the TaskRunStatus.
+func StepState(ops ...tb.StepStateOp) tb.TaskRunStatusOp {
+	return func(s *v1alpha1.TaskRunStatus) {
+		state := &v1alpha1.StepState{}
+		for _, op := range ops {
+			op(state)
+			s.Steps = append(s.Steps, *state)
```
vdemeester: This should  be out of the loop — you "build" the state, then you add it.

pkg/test/builder/taskrun.go
```diff
@@ -14,3 +15,23 @@ func TaskRunCompletionTime(ct time.Time) tb.TaskRunStatusOp {
...
+		}
+	}
+}
+
+func StateTerminated(name string, exitcode int) tb.StepStateOp {
+	return func(s *v1alpha1.StepState) {
+		s.Name = name
```
vdemeester: Shouldn't this be split into 2 operation : `StepStateTerminated` and `StepStateName` or something ?

pkg/cmd/taskrun/logs.go
hrishin: yeah, that's how container name as are formed here. we need container name at the end to query the logs.

pkg/logs/fakelog.go
hrishin: good question :), just wondering if a test should hold reusable things, while this `fakelogs` specifically for logs only, I felt its better to be close to the package :)

WDYT?

pkg/cli/interface.go
hrishin: resolving this, please free to open it again :sweat_smile:

pkg/cmd/root.go
hrishin: In which PR?

pkg/cmd/taskrun/logs.go
hrishin: resolved, please feel free to re-open it :sweat_smile:

pkg/cmd/root.go
vdemeester: later on :wink:

pkg/logs/fakelog.go
vdemeester: ok let's discuss that later, ain't a big deal :wink:

pkg/cmd/pipelinerun/log_test.go
hrishin: I tried removing it later working on tests I realized indeed it helped me to figure what values are being passed actually. :sweat_smile: 

I do understand it may appear as extra noise but it helps I think. Shall we please keep this :smiley: ?

vdemeester: yes we can :wink:

pkg/test/builder/taskrun.go
hrishin: That's a good suggestion, let's see if we could.

pkg/test/builder/taskrun.go
hrishin: good catch :+1:  , resolved it  :sweat_smile:

pkg/test/builder/taskrun.go
hrishin: good catch :+1:  , resolved it  :sweat_smile:

pkg/cmd/root.go
hrishin: :bowing_man:

vdemeester:
(ah @hrishin you're gonna need to squash some of those commits too :angel:)

hrishin:
yes @vdemeester will squash it at the end :smile:

sthaha:
looks good to me but please squash the commits while keeping the message as the one in PR description

sthaha:
looks good to me but please squash the commits while keeping the message as the one in PR description

hrishin:
this fixes https://github.com/tektoncd/cli/issues/5

Issue #25, opened 2019-05-14
[refactor] Improve list pipelinerun command 


hrishin:
At present, the tests for list pipelinerun command are hard to understand. Also, it should to use the new packages and some tets utils in order to bring consistency with other list commands.

PR #26, opened 2019-05-14
Improves list pipelinerun command tests

hrishin:
# Changes
Test cases were not organized well and kind of
hard to understand it.

This patch improves the test cases readability. Also,
refactors code to use the printers package.

fixes https://github.com/tektoncd/cli/issues/25

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/pkg/cmd/pipelinerun/list_test.go
+++ b/pkg/cmd/pipelinerun/list_test.go
@@ -23,17 +23,77 @@ import (
 	"github.com/knative/pkg/apis"
 	"github.com/spf13/cobra"
 	"github.com/tektoncd/cli/pkg/test"
+	tu "github.com/tektoncd/cli/pkg/test"
+	cb "github.com/tektoncd/cli/pkg/test/builder"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
 	"github.com/tektoncd/pipeline/pkg/reconciler/v1alpha1/pipelinerun/resources"
 	pipelinetest "github.com/tektoncd/pipeline/test"
 	tb "github.com/tektoncd/pipeline/test/builder"
 	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func TestListPipelineRuns(t *testing.T) {
-	now := time.Now()
-	prs := pipelineRuns(now)
+	clock := clockwork.NewFakeClock()
+
+	var (
+		ns          = "namespace"
+		runDuration = 1 * time.Minute
+
+		pr1name         = "pr1-1"
+		pr1Pipeline     = "pipeline"
+		pr1Status       = corev1.ConditionTrue
+		pr1StatusReason = resources.ReasonSucceeded
+		pr1Started      = clock.Now().Add(10 * time.Second)
+		pr1Finished     = pr1Started.Add(runDuration)
+
+		pr2name         = "pr2-1"
+		pr2Pipeline     = "random"
+		pr2Status       = corev1.ConditionTrue
+		pr2StatusReason = resources.ReasonRunning
+		pr2Started      = clock.Now().Add(-2 * time.Hour)
+
+		pr3name         = "pr2-2"
+		pr3Pipeline     = "random"
+		pr3Status       = corev1.ConditionFalse
+		pr3StatusReason = resources.ReasonFailed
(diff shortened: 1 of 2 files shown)


pkg/cmd/pipelinerun/list_test.go
```diff
@@ -22,18 +22,76 @@ import (
 	"github.com/jonboulle/clockwork"
 	"github.com/knative/pkg/apis"
 	"github.com/spf13/cobra"
-	"github.com/tektoncd/cli/pkg/testutil"
+	tu "github.com/tektoncd/cli/pkg/testutil"
```
vdemeester: This one is not needed anymore as https://github.com/tektoncd/pipeline/pull/854 is merged :angel:

sthaha: FYI, it won't work (for now ) as we have pinned down the version to the latest release of pipeline (0.3.1). Should master track pipeline/master instead and make `releases` in sync with pipeline?

sthaha:
> * I am not a huge fan of function names like `TestListPipelineRuns_empty`, `TestListPipelineRunsEmpty` is fine to me (that's probably a nit, but `_` bothers me joy_cat).

@vdemeester 

> The naming convention to declare examples for the package, a function F, a type T and method M on type T are:
```
func Example() { ... }
func ExampleF() { ... }
func ExampleT() { ... }
func ExampleT_M() { ... }
```

>Multiple example functions for a package/type/function/method may be provided by appending a distinct suffix to the name. The suffix must start with a lower-case letter.
```
func Example_suffix() { ... }
func ExampleF_suffix() { ... }
func ExampleT_suffix() { ... }
func ExampleT_M_suffix() { ... }
```
ref: https://golang.org/pkg/testing/#examples

sthaha:
> I am also not a huge fan of testutil package, it doesn't convey any meaning (as an example, the builder part in tektoncd/pipeline is is test/builder package 

`builder` is a  much better name and we could reuse if we agree to call `cli/test/builder` cb` and `pipeline/test/builder` `tb`.

pkg/cmd/pipelinerun/list_test.go
vdemeester: ah good point, I didn't thought of that :sweat_smile:. I think we can follow master and make releases in sync with pipeline indeed (or at least, when we release the cli, we pin pipeline)

vdemeester:
> > I am also not a huge fan of testutil package, it doesn't convey any meaning (as an example, the builder part in tektoncd/pipeline is is test/builder package
> 
> `builder` is a much better name and we could reuse if we agree to call `cli/test/builder` cb`and`pipeline/test/builder` `tb`.

I like that idea :angel:

pkg/cmd/pipelinerun/list_test.go
hrishin: that sounds viable as well

sthaha: @hrishin would you be making that change in this PR?

hrishin: I could @sthaha, if we all agreed :+1:

vdemeester: Let's do the dep/vendor bump in another PR :angel:

pkg/cmd/pipelinerun/list_test.go
```diff
@@ -22,18 +22,76 @@ import (
...
 func TestListPipelineRuns(t *testing.T) {
-	now := time.Now()
-	prs := pipelineRuns(now)
+	clock := clockwork.NewFakeClock()
+
+	var (
+		ns          = "namespace"
+		runDuration = 1 * time.Minute
+
+		pr1name         = "pr1-1"
```
vdemeester: nit: Those could even be inlined

hrishin: `go fmt` formatting it like that only :joy_cat:

vdemeester: By inline I meant no var and using it in the function call directly, it would reduce the noise 😊

hrishin: We could :thinking: , however, it makes the test input more readable, inst it? 
I mean all test data appears first rather than in the `tb` noise? :sweat_smile:

hrishin: fixed the vars section @vdemeester . however, the test needs 3 vars so I kept it there :sweat_smile:

PR #79, opened 2019-06-14
Improve CLI E2E test

pradeepitm12:
This improves cli e2e test by running cli commands through script
- without the presence of pipeline where all the commands
  should fail
- With pipeline where all the commands should pass

# Changes

# Submitter Checklist

review them:

for more details._

# Release Notes

```
release-note
```

### DIFF
--- a/test/e2e-common.sh
+++ b/test/e2e-common.sh
@@ -124,7 +124,7 @@ function run_yaml_tests() {
 
 function install_pipeline_crd() {
   echo ">> Deploying Tekton Pipelines"
-  ko apply -f config/ || fail_test "Build pipeline installation failed"
+  kubectl apply -f https://github.com/tektoncd/pipeline/releases/download/v0.4.0/release.yaml || fail_test "Build pipeline installation failed"
 
   # Make sure thateveything is cleaned up in the current namespace.
   for res in pipelineresources tasks pipelines taskruns pipelineruns; do
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -17,22 +17,127 @@
 # This script calls out to scripts in tektoncd/plumbing to setup a cluster
 # and deploy Tekton Pipelines to it for running integration tests.
 
+# TODO: fix common.sh and  enable set  -euo pipefail
+#set -euxo pipefail
+
 source $(dirname $0)/e2e-common.sh
 
 # Script entry point.
 
-initialize $@
+#set -euo pipefail
 
-header "Setting up environment"
+# ci_run allows certain steps to be skipped when running locally
+# usage: 
+# ci_run && {
+#   commands to be run in CI
+# }
+# 
+ci_run()  {
+  ${LOCAL_CI_RUN:-false} || return 0
+  return 1
+}
+
+# skips the test if it fails unless NO_SKIP is set to true
+# to skip a test: 
+#   run_test "list pipelines" skip ./tkn pipeline list -n default
+# to turn off skipping:
+#   NO_SKIP=true ./e2e-tests.sh ...
+SKIP() { 
+  ${NO_SKIP:-false} && {
+      $@
+      return $?
+  }
+
+  $@ || { 
+    echo "SKIPPING: $@ returned $?"
+    return 0
+  } 
+}
+
+run_test() {
+  local desc="$1"; shift
+
+  echo "Running $@"
+  $@ || fail_test "failed to $desc"
+  echo 
+}
+
+ci_run && {
+  header "Setting up environment"
+  initialize $@
+}
 
 # FIXME(vdemeester) To Do later
 # install_pipeline_crd
 
-failed=0
 
 # Run the integration tests
-header "Running Go e2e tests"
-go_test_e2e ./test || failed=1
+ci_run && {
+  header "Running Go e2e tests"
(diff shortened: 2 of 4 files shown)


pradeepitm12:
Work in progress, do not merge.

test/e2e-tests.sh
```diff
@@ -35,4 +35,19 @@ header "Running Go e2e tests"
 go_test_e2e ./test || failed=1
 
 (( failed )) && fail_test
+
```
vdemeester: `install_pipeline_crd` needs to be uncommented above :angel:

pradeepitm12: We have planned to run cli command without installing the pipeline where all the commands should fail. Further, we will run install_pipeline_crd and create task, taskrun, pipeline, and pipelinerun then run the cli commands which should execute without any error.

vdemeester: ah ok :+1:

chmouel:
I don't know if this should be part of this but it would be nice that the script start with :

`
set -euxo pipefail
`

we have been bitten a lot on tektoncd-pipeline by flakyness due of not failign fast enough

(https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/)

test/e2e-common.sh
```diff
@@ -124,7 +124,7 @@ function run_yaml_tests() {
 
 function install_pipeline_crd() {
   echo ">> Deploying Tekton Pipelines"
-  ko apply -f config/ || fail_test "Build pipeline installation failed"
+  kubectl apply -f config/ || fail_test "Build pipeline installation failed"
```
vdemeester: should be `kubectl apply -f https://github.com/tektoncd/pipeline/releases/download/v0.4.0/release.yaml` :angel:

pradeepitm12: Copy paste error :)

test/e2e-tests.sh
```diff
@@ -35,4 +37,30 @@ header "Running Go e2e tests"
...
+#./tkn pipelinerun logs foo && fail_test "Failed to show logs"
+
+install_pipeline_crd
+
+# change default namespace to tektoncd
+kubectl create namespace tektoncd
+kubectl config set-context $(kubectl config current-context) --namespace=tektoncd
+
+# create pipeline, pipelinerun, task, and taskrun
+kubectl apply -f ./resources/output-pipelinerun.yaml
```
vdemeester: I am guessing the working dir is the git root, so it should be `./test/resource/…` for those

test/e2e-tests.sh
```diff
@@ -35,4 +37,37 @@ header "Running Go e2e tests"
...
+install_pipeline_crd
+
+# change default namespace to tektoncd
+kubectl create namespace tektoncd
+kubectl config set-context $(kubectl config current-context) --namespace=tektoncd
+
+# create pipeline, pipelinerun, task, and taskrun
+kubectl apply -f ./test/resources/output-pipelinerun.yaml
+kubectl apply -f ./test/resources/task-volume.yaml
+sleep 10
```
hrishin: instead of waiting for the arbitrary seconds, should it wait until respective resources state reaches to a certain state?
e.g. for ref. https://github.com/k8spatterns/examples/blob/master/advanced/Controller/config-watcher-controller.sh#L19
:thinking:

hrishin: probably we can also use `kubectl get  -w pr/<name> -o json`

pradeepitm12: @hrishin Thanks for this suggestion. 
As it has to be in a loop so `kubectl get pr/<name> -o json` works

test/e2e-tests.sh
```diff
@@ -35,4 +37,46 @@ header "Running Go e2e tests"
...
+# create pipeline, pipelinerun, task, and taskrun
+kubectl apply -f ./test/resources/output-pipelinerun.yaml
+kubectl apply -f ./test/resources/task-volume.yaml
+
+notready=true
+while $notready
+  do
+    sleep 2 
+    pr=$(kubectl get pipelinerun/output-pipeline-run -o json)
+    if [ "$(echo "$pr" | jq -r .status.conditions[0].status)" = "True" ]; then
```
hrishin: Nice!!

Just one more change?
If the `status` is false its better to exit with `non-zero ` and log the right failure message. So in this way, it would also prevent never ending loop in case of failure.

hrishin: Also, it would be better to add similar checks for the TaskRun status?

vdemeester: indeed :angel:

test/e2e-tests.sh
```diff
@@ -35,4 +37,46 @@ header "Running Go e2e tests"
 go_test_e2e ./test || failed=1
 
 (( failed )) && fail_test
+
+go build github.com/tektoncd/cli/cmd/tkn
+
+./tkn pipeline list && fail_test "Failed to list pipelines"
```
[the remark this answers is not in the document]
sthaha: we have made a change to covert this to
 
```shell
run_test "list pipelines" tkn pipeline list
run_test "list pipeline in an empty namespace" \
    SKIP tkn pipeline list -n empty-ns
```

Issue #93, opened 2019-06-18
Object completions for shell completion


chmouel:
# Expected Behavior

```
% tkn pipelinerun describe [TAB]
pipelinerun1 pipelinerun2e ....
```

# Actual Behavior

completion system don't offer anything else than completion for arguments and keywords, not objects (as which pipelineruns)

chmouel:
Merged in #96

PR #96, opened 2019-06-19
Shell completion on objects

chmouel:
Add completion for objects when doing for example :

```
tkn pipelinerun describe [TAB]
```

it will complete with the available pipelineruns.

It will do this for the describe and logs command

# Release Notes

```
Add shell completion for object, you can now do `tkn pipelinerun describe [TAB]` and you have your available pipelineruns getting completed
```

Closes bug #93


pkg/cmd/completion/bash_completion.go
```diff
@@ -0,0 +1,188 @@
...
+
+package completion
+
+const (
+	// BashCompletionFunc the custom bash completion function to complete object.
+	// The bash completion mechanism will launch the __custom_func func if it cannot
+	// find any completion and pass the object in the ${last_command}
+	// variable. Which then are used to get the pipelineruns/taskruns etc..
+	BashCompletionFunc = `
+__kubectl_get_object()
```
vdemeester: Shouldn't we use `tkn` only here ? :angel: (I'm guessing this is inspired from `kubectl` that's why we have the `kubectl` call)

chmouel: it use kubectl to get the object not tkn 😃

vdemeester: yes, that's what I think too :angel: we'll need an issue to track that :+1:

hrishin: yeah, no issues with `kubectl`. However `tkn` does support the template output.

chmouel: wdym @hrishin where does it support the template output ?

@vdemeester fyi https://github.com/tektoncd/cli/issues/103

sthaha: @chmouel what do you think of handing over the completion back to `tkn` so that we don't have to rely on `kubectl` or `oc ` to be present?

sthaha: @chmouel try: `tkn t ls -o jsonpath='{.items[].metadata.name}'`

chmouel: ah thanks i didn't know we have this,

chmouel: althoouuughhhhh, we need kubectl still for namespace completion (PR to come)?

chmouel: I have changed it to use tkn instead of kubectl, btu for `-n` `--namespace` i will have to change it :(

PR #134, opened 2019-07-02
Add the latest version check to the version command

shuheiktgw:
# Changes

I personally think it would be nice if one can tell whether the newer version of `cli` is available or not, so added a version check to the version command! It simply tells the user to update `cli` if the newer version exists.

Thanks for your review! 👍 

# Submitter Checklist

review them:

for more details._

# Release Notes

```
Add `--check` flag to the version command to check if a newer version is available. 
```


pkg/cmd/version/version.go
```diff
@@ -15,22 +15,85 @@
...
 func Command() *cobra.Command {
 	var cmd = &cobra.Command{
 		Use:   "version",
 		Short: "Prints version information",
 		Run: func(cmd *cobra.Command, args []string) {
 			fmt.Fprintf(cmd.OutOrStdout(), "Client version: %s\n", clientVersion)
+
+			current, latest, err := getVersions()
+			if err != nil {
+				// Suppress the error
```
shuheiktgw: I know this part can be a little bit controversial. 

I want the version command to succeed even if the latest version check fails since, after all, the information about the latest version is not critical. Also, the latest version check can fail relatively easily if, for example, the user does not have an internet access. 

However, I feel ignoring the error is a little bit too extreme, so probably need to print out the error at least with the current version info?（Or if there is a debug flag, I should probably print out the error as a debug info）.

go.sum
```diff
@@ -185,8 +187,6 @@ github.com/stretchr/testify v1.2.2/go.mod h1:a8OnRcib4nhh0OaRAV+Yts87kKdq0PP7pXf
 github.com/tarm/serial v0.0.0-20180830185346-98f6abe2eb07/go.mod h1:kDXzergiv9cbyO7IOYJZWg1U88JhDg3PB6klq9Hg2pA=
 github.com/tektoncd/pipeline v0.3.1 h1:LZ0U+HJ3jg3o7e5A7+8nRDEt0G5yJ5seb/uoTxHtBR0=
 github.com/tektoncd/pipeline v0.3.1/go.mod h1:IZzJdiX9EqEMuUcgdnElozdYYRh0/ZRC+NKMLj1K3Yw=
-github.com/tektoncd/plumbing v0.0.0-20190515135757-19ac801cfcf2 h1:F2DzJ7AOA5y6tDBKPL/vwsDgAWQGI67ph+aKThl8rkI=
```
sthaha: why was this deleted? 
Seems like unrelated to this patch.

shuheiktgw: Thanks for your comment! Sorry and you are right. My editor seemed to run `go mod tidy` implicitly. I'll remove the change above and create an another PR to remove those unnecessary dependencies! 👍

vdemeester: @sthaha @shuheiktgw also it's on the `go.sum`, ain't a big deal that those are removed :angel:

pkg/cmd/version/version.go
vdemeester: Yes, I would vote for printing the error in a warning message on `stderr` or something :+1:

vdemeester:
@shuheiktgw oh you need to do a `go mod vendor` and check out the `vendor` folder :angel:

chmouel:
This is nice feature to be able to see if the CLI used is outdated but I think I am -1 on this.

I don't think we should connect outside (i.e: github) when running the command line version. 

For example think about those big enteprise or gov folks,  that don't allow network connection to the outside 

Moreover IMHO the preferred medium should be the distros packages manager/brew etc.... which are the ones owning the version shipping lifecycles.

This could be seen as a phone home anti-feature too :(

what about instead a `--check` flag?

pkg/cmd/version/version.go
shuheiktgw: Thanks for your comment! Now I fixed to print out the error to the stderr, so please check the commit: https://github.com/tektoncd/cli/pull/134/commits/58ca5b2ec076cf54a19277559eead620463d666d

shuheiktgw:
@chmouel Thanks for your comment! Would you let me know where your comment came from and give me a little bit more information about what kind of change I should make?

chmouel:
@shuheiktgw there was an issue with my comment which was edited instead of commented on, i reverted it https://github.com/tektoncd/cli/pull/134#issuecomment-507973540 does it make sense now? adding a --check option ?

shuheiktgw:
@chmouel Thanks for your explanation and now I get your point! 👍  
I'll close this PR and open a new one to implement `--check` option!

shuheiktgw:
@chmouel I added `--check` flag to the version command! 👍 
Would you review the PR again?

pkg/cmd/version/version.go
```diff
@@ -15,22 +15,92 @@
...
 		Use:   "version",
 		Short: "Prints version information",
 		Run: func(cmd *cobra.Command, args []string) {
 			fmt.Fprintf(cmd.OutOrStdout(), "Client version: %s\n", clientVersion)
+
+			if check {
+				current, latest, err := getVersions()
+				if err != nil {
+					fmt.Fprintf(cmd.OutOrStderr(), "Failed to get the latest version: %s\n", err)
+					return
```
chmouel: is there a way to exit 1 when we have an error?

pkg/cmd/version/version.go
```diff
@@ -15,22 +15,92 @@
...
+				}
+
+				if current.LT(*latest) {
+					fmt.Fprintf(cmd.OutOrStdout(), "The latest version v%s is avairable. Please update.\n", latest)
```
chmouel: perhaps we can print the `html_url` field so the user can easily go to the webpage of the version ? i.e: 

image

shuheiktgw:
@chmouel Thanks for your comments! 👍 I've fixed the code again, so please check the commits above! 😄

pkg/cmd/version/version.go
```diff
@@ -15,22 +15,97 @@
...
+	var check bool
+
 	var cmd = &cobra.Command{
 		Use:   "version",
 		Short: "Prints version information",
-		Run: func(cmd *cobra.Command, args []string) {
+		RunE: func(cmd *cobra.Command, args []string) error {
 			fmt.Fprintf(cmd.OutOrStdout(), "Client version: %s\n", clientVersion)
+
+			if check {
```
hrishin: How about?

```
if !check {
   return nil 
}
// version chcek code
```
Minor change but simplifies code readbility?

pkg/cmd/version/version.go
```diff
@@ -15,22 +15,97 @@
...
+
+			return nil
 		},
 	}
+
+	cmd.Flags().BoolVarP(&check, "check", "c", false, "check if a newer version is available")
 	return cmd
 }
+
+func getVersions() (*semver.Version, *semver.Version, string, error) {
```
hrishin: Its returning 4 values.
Could we improve it?

shuheiktgw:
@hrishin Thanks for your comments! 👍 
I've simplified the code flow, so please check the commit! https://github.com/tektoncd/cli/pull/134/commits/28764b2a2d0152127c29498b05152fb5452c51b2

hrishin:
@shuheiktgw  Thank you! :+1:

hrishin:
/lgtm 

@chmouel could you review it once?

shuheiktgw:
@vdemeester Thank you for your comment! I've fixed to skip the version check if the current version is `dev`, so would you let me know what you think about the change? 😄 

0a1dc6b

pkg/cmd/version/version.go
```diff
@@ -15,22 +15,96 @@
...
+	var check bool
+
 	var cmd = &cobra.Command{
 		Use:   "version",
 		Short: "Prints version information",
-		Run: func(cmd *cobra.Command, args []string) {
+		RunE: func(cmd *cobra.Command, args []string) error {
 			fmt.Fprintf(cmd.OutOrStdout(), "Client version: %s\n", clientVersion)
+
+			if !check || clientVersion == devVersion {
```
vdemeester: I wonder if we should skip silently or display something like "You are running a dev version, you rock!" :man_singer: :joy:

chmouel:
It would be nicer to have the commits squashed! Thanks for addressing it,

shuheiktgw:
Thank you all for reviewing the PR! 
Now I've squashed most of the commits, so would you check them please? 😄

pkg/cmd/version/version.go
```diff
@@ -15,22 +15,96 @@
...
+		return err
+	}
+
+	current, err := parseVersion(clientVersion)
+	if err != nil {
+		return err
+	}
+
+	if current.LT(*latest) {
+		fmt.Fprintf(outStream, "The latest version v%s is avairable. Please check %s.\n", latest, response.HtmlUrl)
```
sthaha: typo: available

sthaha: How about: 
`A newer version (v%s) of  Tekton CLI is available, please check %s` 
else 
 `You are running the latest version (v%s) of Tekton CLI`

shuheiktgw:
@sthaha Thanks for your comment! I've fixed the message and squashed the commits again! 👍

PR #176, opened 2019-07-19
Improve sort feature in tr and pr list

piyush-garg:
This will improve the output experience for user in case of
tr and pr list

Now it shows tr and pr in sorted order in case of tr or pr ls
in standard format but in case of output in json or yaml
format or others it shows different order

It should be consistent for all

### DIFF
--- a/pkg/cmd/pipelinerun/list.go
+++ b/pkg/cmd/pipelinerun/list.go
@@ -17,8 +17,8 @@ package pipelinerun
 import (
 	"fmt"
 	"os"
-	"text/tabwriter"
 	"sort"
+	"text/tabwriter"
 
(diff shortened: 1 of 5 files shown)


pkg/cmd/pipelinerun/list.go
```diff
@@ -29,6 +25,9 @@ import (
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	cliopts "k8s.io/cli-runtime/pkg/genericclioptions"
+	"os"
```
hrishin: need `gofmt` ?

piyush-garg: addressed

Issue #163, opened 2019-07-12
Limit pipelineruns and taskruns Returned by List Subcommand


danielhelfand:
/kind feature

There should be a limit option for both the `pipelinerun` and `taskrun` commands to allow a user to limit how many `pipelinerun`s and `taskrun`s are returned as output. For example:

**Retruns the most recent 10 pipelineruns**

```
tkn pipelinerun ls --limit 10 
```

There should also be a default amount of `pipelineruns` and `taskruns` returned when `tkn pr ls` and `tkn taskrun ls` are run. However, a user specifying more than the default amount should return more than the default value.

girishramnani:
@danielhelfand I think this issue should be close as there is a PR https://github.com/tektoncd/cli/pull/185 which resolves this has been merged

danielhelfand:
@girishramnani The issue is for pipelineruns and taskruns. This has only been implemented for pipelineruns in #185. Was going to have a discussion with others on whether it made sense for taskruns before implementing.

PR #185, opened 2019-07-24
Add Limit Option for PipelineRuns

danielhelfand:
This pull request partially addresses #163. It adds a local flag for `--limit` for `pipelineruns`. The `--limit` flag will allow a user to specify how many `pipelineruns` are returned from `tkn pr ls`. 

If this pull request is accepted, I can implement this same approach for other list commands where `--limit` would be a good use case. 

### Usage

**Returns 5 pipelineruns with most recent start time**

`tkn pr ls --limit 5`

### Acceptable Input

`--limit` must be less than the total number of `pipelineruns` for a particular namespace and also greater than 0

### Default Case

The default case will return all `pipelineruns` for a particular namespace

### Local Results

image

# Changes

* A `Limit` field was added to the cli interface
* `Limit` added as flag under `flags.go`
* Logic was implemented under `pipelinerun/list.go` to slice the list of `pipelineruns` returned if `--limit` is passed in by user
* Tests were added to validate `--limit` logic

# Submitter Checklist

review them:

# Release Notes

```
Adds a limit flag to limit the amount of pipelineruns returned by the pipelinerun list command
```

### DIFF
--- a/pkg/cmd/pipelinerun/list_test.go
+++ b/pkg/cmd/pipelinerun/list_test.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 	"testing"
 	"time"
+	"fmt"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/jonboulle/clockwork"
@@ -112,6 +113,42 @@ func TestListPipelineRuns(t *testing.T) {
 				"pr1-1",
 				"pr2-2",
 				"pr2-1",
+				"",
+			},
+		},
+		{
+			name:    "limit pipelineruns returned to 1",
+			command: command(t, prs, clock.Now()),
+			args:    []string{"list", "-n", "namespace", "-l", fmt.Sprintf("%d", 1)},
+			expected: []string{
+				"NAME    STARTED          DURATION   STATUS      ",
+				"pr1-1   59 minutes ago   1 minute   Succeeded   ",
+				"",
+			},
+		},
+		{
+			name:    "limit pipelineruns negative case",
+			command: command(t, prs, clock.Now()),
+			args:    []string{"list", "-n", "namespace", "-l", fmt.Sprintf("%d", -1)},
+			expected: []string{
+				"",
+			},
+		},
+		{
+			name:    "limit pipelineruns greater than maximum case",
+			command: command(t, prs, clock.Now()),
+			args:    []string{"list", "-n", "namespace", "-l", fmt.Sprintf("%d", 4)},
+			expected: []string{
+				"",
+			},
+		},
+		{
+			name:    "limit pipelineruns with output flag set",
+			command: command(t, prs, clock.Now()),
+			args:    []string{"list", "-n", "namespace", "-o", "jsonpath={range .items[*]}{.metadata.name}{\"\\n\"}{end}", "-l", fmt.Sprintf("%d", 2)},
+			expected: []string{
+				"pr1-1",
+				"pr2-2",
 				"",
 			},
 		},
--- a/docs/cmd/tkn_pipelinerun_list.md
+++ b/docs/cmd/tkn_pipelinerun_list.md
@@ -29,6 +29,7 @@ tkn pr list -n foo",
 ```
       --allow-missing-template-keys   If true, ignore any errors in templates when a field or map key is missing in the template. Only applies to golang and jsonpath output formats. (default true)
   -h, --help                          help for list
+  -l, --limit int                     limit pipelineruns listed (default: return all pipelineruns)
   -o, --output string                 Output format. One of: json|yaml|name|go-template|go-template-file|template|templatefile|jsonpath|jsonpath-file.
       --template string               Template string or path to template file to use when -o=go-template, -o=go-template-file. The template format is golang templates [http://golang.org/pkg/text/template/#pkg-overview].
 ```
--- a/pkg/cmd/pipelinerun/list.go
+++ b/pkg/cmd/pipelinerun/list.go
@@ -35,7 +35,13 @@ const (
 	emptyMsg = "No pipelineruns found"
 )
 
+type ListOptions struct {
+	Limit int
+}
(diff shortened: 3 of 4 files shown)


danielhelfand:
Closing this and will re-implement as non global

danielhelfand:
Reopening with `--limit` as local flag

danielhelfand:
Reopening this with a local option as opposed to global flag approach

chmouel:
would not be better to ask for limit [server wise ](https://kubernetes.io/docs/reference/using-api/api-concepts/#retrieving-large-results-sets-in-chunks) instead of getting everything and filter out ?

danielhelfand:
@chmouel My intention here was to be able to view the last x amount of `pipelineruns`. The intention here was regarding if you didn't necessarily want every `pipelinerun` entry returned in case you regularly use a `pipeline` and have many `pipelineruns`.  

The issue that we run into with this is that `pipelineruns` are sorted by start time, so the issue with requesting server side is trying to figure out which `pipelineruns` will be returned back. If there's a way to pull back the most recent `pipelineruns`, then I agree with what you're saying.

chmouel:
@danielhelfand thanks for the explanation, I guess the `limit`  option in this case would not be that much different than using a | tail -n$NUM  or head $NUM ?

vdemeester:
> It's interesting, I wonder what is the use case? I think from an implementation optimisation, would not be better to ask for limit [from the server ](https://kubernetes.io/docs/reference/using-api/api-concepts/#retrieving-large-results-sets-in-chunks) instead of getting everything and filter out ?

Agreeing with @chmouel I think it should be done server side (using `client-go` I feel it's possible to ask for it).

danielhelfand:
@chmouel @vdemeester Sounds good. I'll look into server side handling of `limit`.

danielhelfand:
Would it make sense to actually change the [v1alpha1 List method](https://github.com/tektoncd/pipeline/blob/master/pkg/client/clientset/versioned/typed/pipeline/v1alpha1/pipelinerun.go#L78) to support a sort option?

This way, we could support a user specifying how they would like `pipelineruns` to be sorted and it could take the sorting process out of `list.go`.

danielhelfand:
@vdemeester @chmouel Following up here, the issue is that the kubernetes API [does not support server-side sorting](https://github.com/kubernetes/kubernetes/issues/80602#issuecomment-515301618) at this time. 

There is the ability to Limit what data is pulled back using the [Limit field of LimitOptions]((https://github.com/kubernetes/kubernetes/blob/78915a10949e74feb77586f71ee035292462ea56/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L377), but objects are sorted by name by default, so using `Limit` server side would not pull back the correct data needed. 

So, with regards to this pr, here's a couple of options:

1. The current implementation could be used until some sort of approach is outlined by the kubernetes API to address server side sorting

2. This feature will not be implemented until there is a way to do server side sorting with the kubernetes API

I think this feature would be useful as is and should not alter what is currently being done with list in terms of pulling all `pipelineruns`. If there are workarounds for server side sorting, that I am not aware of, I would definitely be willing to try and improve this pr with those approaches.

sthaha:
looks good to me but will wait for @vdemeester or @chmouel to approve since they had comments earlier.

danielhelfand:
@sthaha Appreciate you looking it over. Thanks!

vdemeester:
/ok-to-test
(I wonder why the CI did not run…)

danielhelfand:
It looks like there needs to be a generic way to pass arguments for testing. Currently only `[]string` is valid.

danielhelfand:
It appears there isn't the most straight forward way of passing in an int as an argument to a `command` in cobra. If I am mistaken, would be happy to try implementing tests again. 

However, the issue is that [SetArgs](https://github.com/danielhelfand/cli/blob/master/pkg/test/cobra.go#L35) appears to only accept strings. Would be happy to work on adding tests in another pr for this.

danielhelfand:
Figured out how to use `fmt.Sprintf("%d", -1)` to pass int values. Should be good for another look.

pkg/cmd/pipelinerun/list_test.go
```diff
@@ -115,6 +116,42 @@ func TestListPipelineRuns(t *testing.T) {
...
+			expected: []string{
+				"",
+			},
+		},
+		{
+			name:    "limit pipelineruns greater than maximum case",
+			command: command(t, prs, clock.Now()),
+			args:    []string{"list", "-n", "namespace", "-l", fmt.Sprintf("%d", 4)},
+			expected: []string{
+				"",
```
danielhelfand: Test result set to `""` since error message is sent as Stderr and result is returning `nil`.

```golang
fmt.Fprintf(os.Stderr, "Limit was %d but cannot be greater than %d\n", limit, prslen)
```

pkg/cmd/pipelinerun/list_test.go
```diff
@@ -115,6 +116,42 @@ func TestListPipelineRuns(t *testing.T) {
...
+				"pr1-1   59 minutes ago   1 minute   Succeeded   ",
+				"",
+			},
+		},
+		{
+			name:    "limit pipelineruns negative case",
+			command: command(t, prs, clock.Now()),
+			args:    []string{"list", "-n", "namespace", "-l", fmt.Sprintf("%d", -1)},
+			expected: []string{
+				"",
```
danielhelfand: Test result set to `""` since error message is sent as Stderr and result is returning `nil` : 

```golang
fmt.Fprintf(os.Stderr, "Limit was %d but must be a positive number\n", opts.Limit)
return nil
```

danielhelfand:
@vdemeester Can you give this another review when you have a moment?

danielhelfand:
@vdemeester Yup, just squashed and pushed it up. Thanks!

Issue #195, opened 2019-08-01
`pipeline start` does not use namespace flag if no namespace has been set in current context


Conky5:
# Expected Behavior

`tkn --namespace <namespace> pipeline start <pipeline>` should always look for pipelines in the given `namespace`.

# Actual Behavior

`tkn --namespace <namespace> pipeline start <pipeline>` looks in the `default` namespace if no namespace is set for the current context.

# Steps to Reproduce the Problem

1. Have an unset namespace for your current cluster context:

```
# unset the namespace for the current context
$ CURRENT_CONTEXT=$(kubectl config current-context)
$ kubectl config unset contexts.${CURRENT_CONTEXT}.namespace
...
# verify it is unset
$ kubectl config view --minify --output 'jsonpath={..namespace}'
```

2. Have a pipeline in the namespace:
```
$ tkn --namespace my-namespace pipeline list
NAME   AGE              LAST RUN   STARTED   DURATION   STATUS
test   10 minutes ago   ---        ---       ---        ---
```

3. Attempt to start an existing pipeline with a specified namespace:
```

$ tkn --namespace my-namespace pipeline start test
Error: invalid pipeline name
```

# Additional Info

I encountered this issue in a fresh environment where the namespace had not been set with `kubectl config set-context --current --namespace=my-namespace`.

PR #196, opened 2019-08-01
InitParams before arg validation in pipeline start

Conky5:
Fixes https://github.com/tektoncd/cli/issues/195

# Changes

Invoke `InitParams` so that the namespace will get set before trying to
validate the pipeline exists. Without this the `default` namespace will
be used even if `--namespace` is set and cause confusion for users in a
fresh cli environment that does not have a namespace set for the current
context.

# Submitter Checklist

review them:

for more details._

# Release Notes

```
Ensure namespace flag is honored when detecting pipelines when using pipeline start.
```

### DIFF
--- a/pkg/cmd/pipeline/start.go
+++ b/pkg/cmd/pipeline/start.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/tektoncd/cli/pkg/cli"
+	"github.com/tektoncd/cli/pkg/flags"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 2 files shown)


pkg/cmd/pipeline/start.go
```diff
@@ -80,11 +77,17 @@ func startCommand(p cli.Params) *cobra.Command {
...
-# start pipeline foo by creating a pipelienrun named "foo-run-xyz123" from the namespace "bar"
+# start pipeline foo by creating a pipelinerun named "foo-run-xyz123" from the namespace "bar"
 tkn pipeline start foo --param NAME=VALUE --resource source=scaffold-git  -s ServiceAccountName  -n bar
 `,
 		SilenceUsage: true,
-		Args:         NameArg(p),
+		Args: func(cmd *cobra.Command, args []string) error {
+			initResult := flags.InitParams(p, cmd)
+			if initResult != nil {
+				return initResult
```
vdemeester: We still need to call `NameArg` to get the Pipeline, don't we ?

Conky5: This is intended to return an error if `InitParams` has an error. I've renamed the variable to `err` to make this clear. `NameArg` is getting returned after this check is passed.

PR #198, opened 2019-08-05
Improve pipeline logs command

pradeepitm12:
Address a few review comments from last PR
Rephrase example message for pipeline log command
Improves output messages if pipeline or runs not found
Verify if the given input pipeline exist
Adds more test cases

# Changes

# Submitter Checklist

review them:

for more details._

# Release Notes

```
release-note
```

### DIFF
--- a/pkg/cmd/pipeline/logs.go
+++ b/pkg/cmd/pipeline/logs.go
@@ -53,17 +53,17 @@ func logCommand(p cli.Params) *cobra.Command {
 	}
 
 	eg := `
-  # show logs interactively for no inputs
+  # interactive mode: shows logs of the selected pipeline run
     tkn pipeline logs -n namespace
 
-  # show logs interactively for given pipeline
-    tkn pipeline logs pipeline_name -n namespace
+  # interactive mode: shows logs of the selected pipelinerun of the given pipeline
+    tkn pipeline logs pipeline -n namespace
 
   # show logs of given pipeline for last run
-    tkn pipeline logs pipeline_name -n namespace --last
+    tkn pipeline logs pipeline -n namespace --last
 
   # show logs for given pipeline and pipelinerun
-    tkn pipeline logs pipeline_name pipelinerun_name -n namespace
+    tkn pipeline logs pipeline run -n namespace
   
    `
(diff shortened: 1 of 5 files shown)


pkg/cmd/pipeline/logs.go
```diff
@@ -129,11 +137,16 @@ func (opts *logOptions) init(args []string) error {
 }
 
 func (opts *logOptions) getAllInputs() error {
-	ps, err := allPipelines(opts.params)
+	ps, err := opts.allPipelines()
```
vdemeester: This is not consistent with `verify`. Why this change ?

pradeepitm12: `verify()` checks whether the input which is pipeline given by user exist or not.
`getAllInputs()` takes no input from the user and given options to select from.

`tkn pipeline logs foo` showed message no pipelinerun found for the pipeline foo even if the pipeline itself is not there, `verify()` checks if the pipeline exists given by user.

vdemeester: my point is on the `go` level. somewhere we do `opts.allPipelines()` and elsewhere we do `verify(opts.pipelineName, opts.param)`. Why not keep `allPipelines(opts.params)` then ?

pradeepitm12: Make sense, I will make it consistent and update this.

pkg/cmd/pipelinerun/log_reader.go
```diff
@@ -50,7 +50,7 @@ func (lr *LogReader) Read() (<-chan Log, <-chan error, error) {
 	tkn := lr.Clients.Tekton
 	pr, err := tkn.TektonV1alpha1().PipelineRuns(lr.Ns).Get(lr.Run, metav1.GetOptions{})
 	if err != nil {
-		return nil, nil, fmt.Errorf("%s : %s", msgPRNotFoundErr, err)
+		return nil, nil, fmt.Errorf(err.Error())
```
vdemeester: nit: for the future, we should think of using `golang.org/x/xerrors` to wrap errors :angel:

PR #212, opened 2019-08-12
Add Limit Option for TaskRuns

### LINKED ISSUE #163: Limit pipelineruns and taskruns Returned by List Subcommand

danielhelfand:
Closes #163. 

This pull request adds a local flag (`--limit`) for `taskruns`. The `--limit` flag will allow a user to specify how many `taskruns` are returned from `tkn tr ls`. 

It follows the same pattern used in #185 for `pipelineruns`, but also changes the behavior for this feature for `pipelineruns`. Now, instead of returning an error when `limit` is greater than the max amount of `pipelineruns`, it will return all `pipelineruns` available.

### Usage

**Returns 5 taskruns with most recent start time**

`tkn tr ls --limit 5`

### Acceptable Input

`--limit` must be less than the total number of `taskruns` for a particular namespace and also greater than 0

### Default Case

The default case will return all `taskruns` for a particular namespace

### Local Results

image

# Submitter Checklist

review them:

# Release Notes

```
Adds a limit flag to limit the amount of taskruns returned by the taskrun list command
```

### DIFF
--- a/pkg/cmd/taskrun/list.go
+++ b/pkg/cmd/taskrun/list.go
@@ -35,7 +35,13 @@ const (
 	emptyMsg = "No taskruns found"
 )
 
+type ListOptions struct {
+	Limit int
+}
+
 func listCommand(p cli.Params) *cobra.Command {
+
+	opts := &ListOptions{Limit: 0}
 	f := cliopts.NewPrintFlags("list")
 	eg := `
 # List all TaskRuns of Task name 'foo'
@@ -57,7 +63,12 @@ tkn pr list -n foo \n",
 				task = args[0]
 			}
 
-			trs, err := list(p, task)
+			if opts.Limit < 0 {
+				fmt.Fprintf(os.Stderr, "Limit was %d but must be a positive number\n", opts.Limit)
+				return nil
+			}
+
+			trs, err := list(p, task, opts.Limit)
 			if err != nil {
 				fmt.Fprintf(os.Stderr, "Failed to list taskruns from %s namespace \n", p.Namespace())
 				return err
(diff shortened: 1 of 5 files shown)


pkg/cmd/taskrun/list.go
```diff
@@ -109,10 +127,23 @@ func list(p cli.Params, task string) (*v1alpha1.TaskRunList, error) {
...
 
-	if len(trs.Items) != 0 {
+	trslen := len(trs.Items)
+
+	if trslen != 0 {
 		sort.Sort(byStartTime(trs.Items))
 	}
 
+	// Limit after sort since taskruns are not in correct order until sorted
+	if limit > trslen {
```
hrishin: Instead of returning an error, would it be nice to just return whatever objects found? :thinking:

pkg/cmd/taskrun/list_test.go
```diff
@@ -143,6 +144,42 @@ func TestListTaskRuns(t *testing.T) {
 			args:     []string{"list", "-n", "random"},
 			expected: []string{emptyMsg, ""},
 		},
+		{
+			name:    "limit taskruns returned to 1",
+			command: command(t, trs, now),
+			args:    []string{"list", "-n", "foo", "-l", fmt.Sprintf("%d", 1)},
```
hrishin: How about replacing `fmt.Sprintf("%d", 1)` -> `"1"`?

danielhelfand: I tried using `"1"` previously in #185 and it did not work. The argument needs to be an int, so this is the way I found to pass it as a String and have it eventually interpreted as an int.

pkg/cmd/taskrun/list.go
danielhelfand: My personal opinion is that it would be confusing to return something with invalid input, but, if you feel strongly about it, I can set up a conditional that will display the error message and set `limit` to 0 so it will display all `taskruns` found in the namespace.

danielhelfand: If you would like me to update it, it would make sense to also do this for `pipelineruns` as well.

pkg/cmd/taskrun/list_test.go
hrishin: :+1:

pkg/cmd/taskrun/list.go
hrishin: Well, I also felt to return an error initially and can relate the thought. However, if we step back for a moment, the user's intention is to limit the number of object to Y. So having a number of objects <=  Y anyways serves the users purpose? isn't it?  I'm missing something on this? 
 :thinking:

However, if you are not convinced then no worries at all :smile:  (Hakuna Matata)

danielhelfand: Haha, I was thinking about this and came to the same conclusion. It makes sense to just return them all instead of sending an error message. 

Let me touch this up for `taskruns` and `pipelineruns`, and I'll have you give it another look.

danielhelfand: Pushed up a change that will return all values when greater than max amount in a namespace and updated tests.

danielhelfand:
@hrishin Just let me know if you would like me to squash commits.

pkg/cmd/taskrun/list.go
hrishin: Thank you @danielhelfand :smile:  :+1:

danielhelfand:
@vdemeester @hrishin Thanks for reviewing. Fixed the CI issue and squashed commits. Just waiting on LGTM label. Let me know if there's anything else you need from me.

PR #218, opened 2019-08-17
Add the information regarding docs

piyush-garg:
This will add the information regarding generating
and updating docs(markdown and man pages) in the PR

Also updated the docs according to current master

# Submitter Checklist

review them:


DEVELOPMENT.md
```diff
@@ -76,7 +76,8 @@ While iterating on the project, you may need to:
 
 1. [Install Tekton pipeline](#install-pipeline)
 2. [Building Tekton client](#building-tekton-client)
-1. [Add and run tests](./test/README.md#tests)
+3. [Add and run tests](./test/README.md#tests)
```
sthaha: fyi: markdown simplifies ordered list e.g. 

```md
1. foo 
1. bar
1. baz
```

docs/cmd/tkn_pipeline_logs.md
```diff
@@ -15,17 +15,17 @@ Show pipeline logs
 ### Examples
 
 
-  # show logs interactively for no inputs
+  # interactive mode: shows logs of the selected pipeline run
```
sthaha: Won't this be rendered as H1? 

# interactive mode: ...

docs/cmd/tkn_task_list.md
```diff
@@ -19,7 +19,6 @@ Lists tasks in a namespace
 ```
       --allow-missing-template-keys   If true, ignore any errors in templates when a field or map key is missing in the template. Only applies to golang and jsonpath output formats. (default true)
   -h, --help                          help for list
-  -l, --limit int                     limit taskruns listed (default: return all taskruns)
```
sthaha: why is this removed?

docs/cmd/tkn_pipeline_logs.md
piyush-garg: yes, need to change example in the respective file

DEVELOPMENT.md
piyush-garg: oh nice, I was not aware of this, so corrected the order.

docs/cmd/tkn_pipeline_logs.md
vdemeester: Yeah, those need to be updated in the source code (maybe we can remove some and add those in a md file so it only appears on man pages and reference docs)

piyush-garg:
@hrishin PR has been updated with the current master.

PR #227, opened 2019-08-20
Auto generate zsh completion

chmouel:
# Changes

We were using a hand written one before, which was buggy (global variable not
exposed when installed) and we had to syncronize it every time.

With the cobra fork from https://github.com/chmouel/cobra we can now use cobra
generated function.

We just need to Mark a command with the custom completion, for example :

`c.MarkZshCompPositionalArgumentCustom(1, "__tkn_get_pipelinerun")`

will complete pipelinerun to that command for the `1`/first argument.

The `__tkn_get_pipelinerun` is a shell function referenced in the
`zsh_completion.go` file.

# Submitter Checklist

review them:

- [ 🤷‍♂  ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ 🤷‍♂ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [ 🌮 ] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)


go.mod
```diff
@@ -62,3 +62,5 @@ require (
 )
 
 replace github.com/kr/pty => github.com/creack/pty v1.1.7
+
+replace github.com/spf13/cobra => github.com/chmouel/cobra v0.0.0-20190820110723-8f09bb39b20d
```
hrishin: shall we file a follow up issue to host the fork at right place?

chmouel: where is a right place you thihnk? i am hoping this is just temporary

hrishin: then lets not worry about it, lets fight other battles

PR #228, opened 2019-08-21
Sort PipelineRuns for Pipeline Logs by Start Time and Add Limit Flag

### LINKED ISSUE #224: Sort pipelineruns based on time started with configurable limit
The `tkn pipeline logs` command should sort the listed pipelineruns based on time started. Currently it's sorted alphabetically which means the last pipelineruns might not be even in the 5 runs that are displayed.

Furthermore, it should be possible to list a greater number of runs based on a flag for example `--n 10` would list last 10 pipelineruns.

danielhelfand:
Closes #224 

# Changes

This pull request adds sorting of `pipelineruns` by their start times. It also adds a `limit` flag that allows more or less than 5 `pipelineruns` to be returned. The current implementation only allows a hard coded 5 `pipelineruns` to be returned.

### Local Testing

Start times of some `pipelineruns`:

image

Correct order of `pipelineruns` with default of 5 `pipelineruns`:

image

Limit flag with `L` set to 2:

image

Error messaging for negative number:

image

Specify greater number than 5 for `pipelineruns`:

image

# Submitter Checklist

review them:

for more details._

# Release Notes

```
Sort pipelineruns by start time for tkn pipeline logs command and add limit flag to limit pipelineruns returned from tkn pipeline logs
```

### DIFF
--- a/docs/cmd/tkn_pipeline_logs.md
+++ b/docs/cmd/tkn_pipeline_logs.md
@@ -26,16 +26,17 @@ Show pipeline logs
 
   # show logs for given pipeline and pipelinerun
     tkn pipeline logs pipeline run -n namespace
-  
+
    
 
 ### Options
 
 ```
-  -a, --all      show all logs including init steps injected by tekton
-  -f, --follow   stream live logs
-  -h, --help     help for logs
-  -l, --last     show logs for last run
+  -a, --all         show all logs including init steps injected by tekton
+  -f, --follow      stream live logs
+  -h, --help        help for logs
+  -l, --last        show logs for last run
+  -L, --limit int   lists number of pipelineruns (default 5)
 ```
 
 ### Options inherited from parent commands
--- a/pkg/cmd/pipeline/logs.go
+++ b/pkg/cmd/pipeline/logs.go
@@ -17,6 +17,7 @@ package pipeline
 import (
 	"fmt"
 	"os"
+	"sort"
 	"strings"
 
 	"github.com/AlecAivazis/survey/v2"
@@ -38,6 +39,7 @@ type logOptions struct {
 	follow       bool
 	pipelineName string
 	runName      string
+	limit        int
 }
 
 func nameArg(args []string, p cli.Params) error {
@@ -100,6 +102,7 @@ func logCommand(p cli.Params) *cobra.Command {
 	c.Flags().BoolVarP(&opts.last, "last", "l", false, "show logs for last run")
 	c.Flags().BoolVarP(&opts.allSteps, "all", "a", false, "show all logs including init steps injected by tekton")
 	c.Flags().BoolVarP(&opts.follow, "follow", "f", false, "stream live logs")
+	c.Flags().IntVarP(&opts.limit, "limit", "L", 5, "lists number of pipelineruns")
 
 	_ = c.MarkZshCompPositionalArgumentCustom(1, "__tkn_get_pipeline")
 	return c
@@ -154,6 +157,12 @@ func (opts *logOptions) init(args []string) error {
 }
 
 func (opts *logOptions) getAllInputs() error {
+	err := validate(opts)
+
+	if err != nil {
+		return err
+	}
+
 	ps, err := allPipelines(opts)
 	if err != nil {
 		return err
@@ -182,9 +191,14 @@ func (opts *logOptions) getAllInputs() error {
 }
 
 func (opts *logOptions) askRunName() error {
+	err := validate(opts)
+	if err != nil {
+		return err
+	}
+
 	var ans string
 
-	prs, err := allRuns(opts.params, opts.pipelineName)
+	prs, err := allRuns(opts.params, opts.pipelineName, opts.limit)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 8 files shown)


sthaha:
@pradeepitm12 fyi

pkg/cmd/pipeline/logs.go
```diff
@@ -288,3 +312,17 @@ func lastRun(cs *cli.Clients, ns, pName string) (string, error) {
...
 }
+
+type byStartTime []v1alpha1.PipelineRun
+
+func (s byStartTime) Len() int      { return len(s) }
+func (s byStartTime) Swap(i, j int) { s[i], s[j] = s[j], s[i] }
+func (s byStartTime) Less(i, j int) bool {
+	if s[j].Status.StartTime == nil {
+		return false
+	} else if s[i].Status.StartTime == nil {
```
sthaha: nit: else isn't needed (only a nit)

pkg/cmd/pipeline/logs.go
```diff
@@ -182,9 +191,14 @@ func (opts *logOptions) getAllInputs() error {
 }
 
 func (opts *logOptions) askRunName() error {
+	if opts.limit <= 0 {
```
sthaha: Validation here is fine but cobra already supports https://github.com/spf13/cobra#positional-and-custom-arguments 

Since the validation of options repeat, it may be better to have a `func (opts *logOptions validate() error`

pkg/cmd/pipeline/logs.go
khrm: It's needed because it's checking for nil pointer. Otherwise, it panics in certain conditions(similar to a bug which was fixed recently).

docs/cmd/tkn_pipeline_logs.md
```diff
@@ -32,10 +32,11 @@ Show pipeline logs
...
 ```
-  -a, --all      show all logs including init steps injected by tekton
-  -f, --follow   stream live logs
-  -h, --help     help for logs
-  -l, --last     show logs for last run
+  -a, --all         show all logs including init steps injected by tekton
+  -f, --follow      stream live logs
+  -h, --help        help for logs
+  -l, --last        show logs for last run
+  -L, --limit int   lists an amount of pipelineruns (default 5)
```
hrishin: minor: `List number of pipelineruns` instead of `amount`? :)

pkg/cmd/pipelinerun/logs.go
```diff
@@ -33,6 +33,7 @@ type LogOptions struct {
 	Stream          *cli.Stream
 	Streamer        stream.NewStreamerFunc
 	Tasks           []string
+	Limit           int
```
pradeepitm12: are we using Limit in pipeline run logs as well?
if not then we can limit its scope to local logOptions of pipeline logs

pkg/cmd/pipeline/logs.go
sthaha: @khrm 
I meant only the `else` wasn't needed :D 

```go
if foo {
  ___
} else if bar {
  ___
}
```
becomes
```go
if foo {
  ___
} 
if bar {
  ___
}
```
but again it was a `nit` so wasn't a big deal

pkg/cmd/pipelinerun/logs.go
danielhelfand: No, you're right. It's not needed. I added to `runLogOption` in `pipeline.logs` when first getting started on this thinking it would be, but I will remove it.

docs/cmd/tkn_pipeline_logs.md
danielhelfand: Changed to this in latest push

pkg/cmd/pipeline/logs.go
danielhelfand: I added a `validate()` function in my latest push. I agree with you though that this should eventually be switched to cobra's validation. If it's all right with you, I would like to do this in a separate pr as I would also like to refactor some things around how sorting is handled for `pipelineruns` across the CLI. The switch over to better validation can be part of that, and I can outline it in an issue.

pkg/cmd/pipeline/logs.go
danielhelfand: Changed to your suggestion. I also went ahead and changed for `pipelineruns`/`taskruns` as they also used the same sorting approach.

pkg/cmd/pipelinerun/logs.go
danielhelfand: Removed. Thanks for pointing this out.

danielhelfand:
> Looks good to me overall. ❤️
> 
> This PR introduces behavioural changes and should documented under `docs` directory. see: `make docs man`

Doc changes in latest push.

pradeepitm12:
@danielhelfand 
Could you please try revert Less() as it was earlier 
func (s byStartTime) Less(i, j int) bool {
	if s[j].Status.StartTime == nil {
		return false
	} else if s[i].Status.StartTime == nil {
		return true
	}

	return s[j].Status.StartTime.Before(s[i].Status.StartTime)
}

danielhelfand:
> @danielhelfand
> Could you please try revert Less() as it was earlier
> 
> ```
>     func (s byStartTime) Less(i, j int) bool {
>         if s[j].Status.StartTime == nil {
> 	      return false
>        } else if s[i].Status.StartTime == nil {
> 	      return true
> }
> 
> return s[j].Status.StartTime.Before(s[i].Status.StartTime)
>     }
> ```
> 
> Not sure but can give a try

The issue with the most recent build was a linting issue. Other times it appears `pull-tekton-cli-build-tests` wasn't picking up my latest changes. Hoping this push will resolve.

danielhelfand:
Looks like latest push resolved issues. Ready for another review whenever reviewers have time.

PR #229, opened 2019-08-21
Refactor shell completion

chmouel:
Refactor how we do shell completion,

* Commonise between bash and zsh shell functions so we have less chance to forget one or another,
* Add utility function with a test for adding cusotm completion to a flag,
* Add completion on serviceAccount for `tkn start`


chmouel:
oh fun : 

`I0821 19:54:49.908] 2019/08/21 19:54:49 Error listing files: GET https://api.github.com/repos/tektoncd/cli/pulls/229/files?page=1: 403 API rate limit exceeded for 35.184.145.65. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.) [rate reset in 59m59s]`

PR #230, opened 2019-08-22
Fix homebrew formula test

chmouel:
We were using --version for test which was wrong, it now exists with 1 so this
would fail (before that would exit 0 which was buggy). Let's use tkn version
properly!

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -54,7 +54,7 @@ brews:
     homepage: "https://github.com/tektoncd/cli"
     description: Tekton CLI - The command line interface for interacting with Tekton
     test: |
-      system "#{bin}/tkn", "--version"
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -54,7 +54,7 @@ brews:
     homepage: "https://github.com/tektoncd/cli"
     description: Tekton CLI - The command line interface for interacting with Tekton
     test: |
-      system "#{bin}/tkn", "--version"
+      system "#{bin}/tkn", "version"
```
hrishin: oops that how it supposed to be. why it was not failing then :thinking: during release

chmouel: i don't think brew test was launched,

PR #239, opened 2019-08-26
Display sub commands types only if there is any.

vdemeester:
# Changes

- Won't display "other commands" if there is no utility sub commands
- Won't display "other commands" if there is no sub commands

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -38,13 +38,13 @@ Aliases:
   {{.NameAndAliases}}{{end}}{{if .HasExample}}
 
 Examples:
-  {{.Example}}{{end}}{{if .HasAvailableSubCommands}}
+  {{.Example}}{{end}}{{if .HasAvailableSubCommands}}{{if HasMainSubCommands .}}
 
 Available Commands:{{range .Commands}}{{if (eq .Annotations.commandType "main")}}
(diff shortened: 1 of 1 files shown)


pkg/cmd/root.go
```diff
@@ -83,3 +85,20 @@ func Root(p cli.Params) *cobra.Command {
...
+}
+func hasUtilitySubCommands(cmd *cobra.Command) bool {
+	return len(subCommands(cmd, "utility")) > 0
+}
+
+func subCommands(cmd *cobra.Command, annotation string) []*cobra.Command {
+	cmds := []*cobra.Command{}
+	for _, sub := range cmd.Commands() {
+		if sub.IsAvailableCommand() && sub.Annotations["commandType"] == annotation {
+			cmds = append(cmds, sub)
```
hrishin: can return `true` instead of populating any array, given the fact `cmds` are not used any where?

vdemeester: @hrishin maybe :angel: I was thinking we may want to use this later on and need to array but…

hrishin: Ahh was also wondering on the same if we can use it later.

Cool, nothing much to worry about it :smile:

PR #251, opened 2019-08-29
Add yaml lint makefile target

chmouel:
Need to find how to add yamllint binary to the unittest container images  before
enabling it by default in lint: makefile target.


chmouel:
@vdemeester would you know by any chace how can we add yamllint into the images?

PR #252, opened 2019-08-29
Mock github API calls

chmouel:
we were querying github api on every testrun and were hitting ratelimit
issues. the tests are not totally perfect but they are better that what they
were before.


chmouel:
/cc @vdemeester @hrishin 

(some tests has been failing due of the gh ratelimit we need to get this in)

Issue #13, opened 2019-04-23
Cancel task run


siamaksade:
**Input (Required):** task-run name

**Output:** Cancel the running task-run

16yuki0702:
I thinks this is completed.
Could I close it??

chmouel:
yep we can close this, thanks @16yuki0702 👍 

/close
/meow

PR #281, opened 2019-09-04
Add canceling feature to taskrun command

16yuki0702:
# Changes
Add canceling feature to taskrun command for this https://github.com/tektoncd/cli/issues/13

# Submitter Checklist

review them:

for more details._

# Release Notes

```
you can cancel taskrun from tkn
```


pkg/cmd/taskrun/cancel_test.go
```diff
@@ -0,0 +1,107 @@
...
+						Status: corev1.ConditionTrue,
+						Reason: resources.ReasonSucceeded,
+					}),
+				),
+			),
+		}
+		cs, _ := test.SeedTestData(t, pipelinetest.Data{TaskRuns: trs})
+		cs2, _ := test.SeedTestData(t, pipelinetest.Data{TaskRuns: trs2})
+
+		cs2.Pipeline.PrependReactor("update", "taskruns", func(action k8stest.Action) (bool, runtime.Object, error) {
```
hrishin: Wonders why we need this? :thinking:

pkg/cmd/taskrun/cancel_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	k8stest "k8s.io/client-go/testing"
+	"knative.dev/pkg/apis"
+)
+
+func TestTaskRunCancel(t *testing.T) {
+	seeds := make([]pipelinetest.Clients, 0)
+	failures := make([]pipelinetest.Clients, 0)
+
+	for i := 0; i < 1; i++ {
```
hrishin: wonders what we are trying to archive with a loop? :thinking: 

I can see it iterates for once if so we can assign the values directly

pkg/cmd/taskrun/cancel_test.go
16yuki0702: Thank you for your review!!
It is for making update error and just improve coverage.
It's test following code.
```
tr.Spec.Status = v1alpha1.TaskRunSpecStatusCancelled
if _, err := cs.Tekton.TektonV1alpha1().TaskRuns(p.Namespace()).Update(tr); err != nil {
	return fmt.Errorf("failed to cancel taskrun %q: %s", trName, err)
}
```

pkg/cmd/taskrun/cancel_test.go
16yuki0702: Thank you for your review!!
It takes increasing test cases into account for future.
(for needing more test datas)

but for now as you mentioned it is no need these test cases :sweat_smile:
I'll delete this loop for avoiding confuse, thanks!!

pkg/cmd/taskrun/cancel_test.go
hrishin: :+1: got it thanks!

Gave me the impression that every time update call would lead to an error state?
How about? 

```
testParams := []struct {
		name      string
		command   []string
		input     []v1alpha1.TaskRun
		wantError bool
		want      string
	}{
		{
			name:      "Canceling taskrun successfully",
			command:   []string{"cancel", "taskrun-1", "-n", "ns"},
			input:     seeds[0],
			wantError: false,
			want:      "TaskRun cancelled: taskrun-1\n",
		},
                ....
	}

       for _, t := range testParams {
                cs := seedData()
                if wantErr {
                       prependReactor()
                }
       }

```

16yuki0702: PrependReactor is called on only cs2, and cs2 is added to failures slices.
successful case is using seeds slice as a input, and failure case is using failure slice as a input.
so update call would be error only failure test cases :thumbsup: !

```
seeds = append(seeds, cs)
	failures = append(failures, cs2)

	testParams := []struct {
		name      string
		command   []string
		input     pipelinetest.Clients
		wantError bool
		want      string
	}{
		{
			name:      "Canceling taskrun successfully",
			command:   []string{"cancel", "taskrun-1", "-n", "ns"},
			input:     seeds[0],
			wantError: false,
			want:      "TaskRun cancelled: taskrun-1\n",
		},
		{
			name:      "Not found taskrun",
			command:   []string{"cancel", "nonexistent", "-n", "ns"},
			input:     seeds[0],
			wantError: true,
			want:      "failed to find taskrun: nonexistent",
		},
		{
			name:      "Failed canceling taskrun",
			command:   []string{"cancel", "failure-taskrun-1", "-n", "ns"},
			input:     failures[0],
			wantError: true,
			want:      "failed to cancel taskrun \"failure-taskrun-1\": test error",
		},
```

hrishin: cool no worries :+1:

16yuki0702: Thanks !!

pkg/cmd/taskrun/cancel_test.go
16yuki0702: I fixed it !!

chmouel:
/lgtm

i haven't tested it but this looks fine to me,

chmouel:
May be not trivial but would be nice to add an E2E test for this,

16yuki0702:
@chmouel 
> May be not trivial but would be nice to add an E2E test for this,

Thanks !! I'll add later :thumbsup:

PR #290, opened 2019-09-05
🎨 Add rainbow colouring of steps label

### LINKED ISSUE #285: rainbow colours with "tkn * logs" command
We have some blue colours highlighting the task name which is nice and dandy but would be nicer if we had different colours for the different tasks, 

the tiny little things that make people happier when digging into logs 💘

chmouel:
We were showing the same blue color for every steps, let's now cycle of a list of
colors.

Demo: https://asciinema.org/a/hbJk1PqMFE8hleRpDsraPw0WB

Closes #285 

# Release Notes
```
Steps are now labelled with different colours instead of the same one 🎨
```

### DIFF
--- a/pkg/formatted/color.go
+++ b/pkg/formatted/color.go
@@ -16,12 +16,69 @@ package formatted
 
 import (
 	"io"
+	"sync"
+	"sync/atomic"
 
 	"github.com/fatih/color"
 )
 
+var (
+	// Red is avoided as keeping it for errors
+	palette = []color.Attribute{
+		color.FgHiGreen,
+		color.FgHiYellow,
+		color.FgHiBlue,
+		color.FgHiMagenta,
+		color.FgHiCyan,
+	}
+)
+
+type atomicCounter struct {
+	value     uint32
+	threshold int
+}
+
+func (c *atomicCounter) next() int {
+	v := atomic.AddUint32(&c.value, 1)
+	next := int(v-1) % c.threshold
+	atomic.CompareAndSwapUint32(&c.value, uint32(c.threshold), 0)
+	return next
+
+}
+
+type rainbow struct {
+	cache   sync.Map
+	counter atomicCounter
+}
+
+func newRainbow() *rainbow {
+	return &rainbow{
+		counter: atomicCounter{threshold: len(palette)},
+	}
+}
+
+func (r *rainbow) get(x string) color.Attribute {
+	if value, ok := r.cache.Load(x); ok {
+		return value.(color.Attribute)
+	}
+
+	clr := palette[r.counter.next()]
+	r.cache.Store(x, clr)
+	return clr
+}
+
+// Fprintf formats according to a format specifier and writes to w.
+// the first argument is a label to keep the same colour on.
+func (r *rainbow) Fprintf(label string, w io.Writer, format string, args ...interface{}) {
+	attribute := r.get(label)
+	crainbow := color.Set(attribute).Add(color.Bold)
+	crainbow.Fprintf(w, format, args...)
+}
+
 //Color formatter to print the colored output on streams
 type Color struct {
+	Rainbow *rainbow
+
 	red  *color.Color
 	blue *color.Color
 }
@@ -29,6 +86,8 @@ type Color struct {
 //NewColor returns a new instance color formatter
 func NewColor() *Color {
 	return &Color{
+		Rainbow: newRainbow(),
+
 		red:  color.New(color.FgRed),
 		blue: color.New(color.FgBlue),
 	}
@@ -37,11 +96,6 @@ func NewColor() *Color {
 //PrintBlue prints the formatted content to given destination in blue color
 func (c *Color) PrintBlue(w io.Writer, format string, args ...interface{}) {
 	c.blue.Fprintf(w, format, args...)
-}
-
-//Header prints the formatted content to given destination in blue color
-func (c *Color) Header(w io.Writer, format string, args ...interface{}) {
-	c.PrintBlue(w, format, args...)
 }
 
 //PrintBlue prints the formatted content to given destination in red color
--- a/pkg/cmd/pipelinerun/log_writer.go
+++ b/pkg/cmd/pipelinerun/log_writer.go
@@ -45,7 +45,8 @@ func (lw *LogWriter) Write(s *cli.Stream, logC <-chan Log, errC <-chan error) {
 				fmt.Fprintf(s.Out, "\n")
 				continue
 			}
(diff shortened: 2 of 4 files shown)


pkg/formatted/color.go
```diff
@@ -20,15 +20,55 @@ import (
...
+		current: 0,
+		cached:  make(map[string]color.Attribute),
+		all:     allColors,
+	}
+}
+
+func (rc *RainbowColors) Get(label string) *color.Color {
+	if _, ok := rc.cached[label]; !ok {
+		rc.cached[label] = rc.all[rc.current]
+		rc.current++
```
vdemeester: There might be a race here if this is called in different goroutine 🤔

chmouel: yeah i tough about that but since we werent using it i didn't follow tough, what do you suggest? external lib for locking etc...

chmouel: and fyi i did test it with `-race` and it was running well but i'll just sync/lock/release it in case off,

pkg/formatted/color.go
```diff
@@ -16,19 +16,64 @@ package formatted
...
+}
+
+func (rc *RainbowColors) Get(label string) *color.Color {
+	mux.Lock()
+	defer mux.Unlock()
+
+	if _, ok := rc.cached[label]; !ok {
+		rc.cached[label] = rc.all[rc.current]
+		rc.current++
+		if rc.current >= len(rc.all) {
```
sthaha: nit: usually this is done using `rc.current = (rc.current +1) % len(rc.all)` without having the need for an `if`

pkg/formatted/color.go
```diff
@@ -16,19 +16,64 @@ package formatted
...
+func NewRainbowColors() *RainbowColors {
+	return &RainbowColors{
+		current: 0,
+		cached:  make(map[string]color.Attribute),
+		all:     allColors,
+	}
+}
+
+func (rc *RainbowColors) Get(label string) *color.Color {
+	mux.Lock()
```
sthaha: mux must be part of the `rc` and not a package level variable; this allows for mulitple RainbowColors to be used without having one instance to wait for another instance.

pkg/formatted/color.go
```diff
@@ -16,19 +16,64 @@ package formatted
...
+	defer mux.Unlock()
```
sthaha: it is better to use `Rlock` and `RWlocks`

pkg/formatted/color.go
```diff
@@ -16,19 +16,64 @@ package formatted
 
 import (
 	"io"
+	"sync"
 
 	"github.com/fatih/color"
 )
 
+var (
+	mux       sync.Mutex
```
sthaha: mux should be instance variable and not package var since it only locks the `rc.cached` and not a package level variable.

pkg/formatted/color.go
vdemeester: @chmouel right, we might not be in the case of race condition in the cli, but if somebody uses this might hit it

pkg/formatted/color.go
```diff
@@ -16,19 +16,74 @@ package formatted
...
+
+func (c *counter) Get() int {
+	c.mu.RLock()
+	n := c.n
+	c.mu.RUnlock()
+	return n
+}
+
+var (
+	allColors = []color.Attribute{
```
chmouel: red is avoided since we keep it for error messages, (probably should have put this in the comment instead of here)

pkg/formatted/color.go
```diff
@@ -16,19 +16,74 @@ package formatted
 
 import (
 	"io"
+	"sync"
 
 	"github.com/fatih/color"
 )
 
+type counter struct {
```
sthaha: when I mentione RLock and Lock() I was hinting to having a single mutex in the rainbow. which will be RLocked for checking if the item is in cache and if not, Lock() to add the item in cache. but i like your implementation to create a round robin counter but for the counter itself, `Get` and `increase` isn't quite as useful as just a `next` which returns the next value. 

please consider one of these implemention for counter. 1. that uses atomic, the other that uses a mutex

```
type atomicCounter struct {
  value     uint32
  threshold int
}

func (c *atomicCounter) next() int {
  v := atomic.AddUint32(&c.value, 1)
  next := int(v-1) % c.threshold
  atomic.CompareAndSwapUint32(&c.value, uint32(c.threshold), 0)
  return next

}

type counter struct {
  value     int
  threshold int
  m         sync.Mutex
}

func (c *counter) next() int {
  c.m.Lock()
  defer c.m.Unlock()

  v := c.value
  c.value = (v + 1) % c.threshold
  return v

}
```

and the rainbow would be some what like this 
```
type rainbow struct {
  cache   sync.Map
  counter counter
}

func newRainbow() *rainbow {
  return &rainbow{
    counter: counter{threshold: len(palette)},
  }

}

func (r *rainbow) colorFor(x string) color {
  if value, ok := r.cache.Load(x); ok {
    return value.(color)
  }

  clr := palette[r.counter.next()]
  r.cache.Store(x, clr)
  return clr
}
```

chmouel: I am not really sure why we would want to prematurely optimizing  this on a CLI with a very low requirement for scaling but since you wrote the code i made the change.

pkg/formatted/color.go
```diff
@@ -16,19 +16,78 @@ package formatted
...
+	}
+
+	clr := palette[r.counter.next()]
+	r.cache.Store(x, clr)
+	return clr
+}
+
+// Fprintf formats according to a format specifier and writes to w.
+// the first argument is a label to keep the same colour on.
+func (r *rainbow) Fprintf(label string, w io.Writer, format string, args ...interface{}) {
```
sthaha: imho, should follow the same signature/convention as Fprintf which is to accept `io.Writer` as the first argument.

pkg/formatted/color.go
```diff
@@ -16,19 +16,78 @@ package formatted
...
+	palette = []color.Attribute{
+		color.FgHiGreen,
+		color.FgHiYellow,
+		color.FgHiBlue,
+		color.FgHiMagenta,
+		color.FgHiCyan,
+	}
+)
+
+type atomicCounter struct {
```
sthaha: I liked your earlier name: counter but I am not opposed to this (internal implementation)

pkg/formatted/color.go
sthaha: I think this change can be a follow up patch. 
hence lgtm.

chmouel: it would never be the same as Fprintf since it need to be 'labelled', i can rename it to LabelFrprintf if that cause for confusion,

Issue #297, opened 2019-09-09
Pipeline start should not ask for interactive when specified --last


piyush-garg:
# Expected Behavior

`tkn p start {pipeline name} -l` should not ask for resource and param details.

But it should override the given resource and params if specified by flag `-r` or `-p`.

# Actual Behavior

`tkn p start {pipeline name} -l` asks for resource and param details interactively.

# Steps to Reproduce the Problem

1. Create a pipeline with resource/param
2. Create a run
3. Execute `tkn p start pipelinename -l`

# Additional Info

PR #302, opened 2019-09-09
Fix interactive prompt with --last

piyush-garg:
This will fix the issue of interactive prompt getting
started with --last flag in scenario of not using
resource or param flag. It should take the value from
last pipelinerun in case of no use of resource and
param flag. In case of using the flag, it will override
the value of last run but should never start
interactive prompt.

Change the resource selection question in interactive
prompt for better user experience.

Adds tests for the given scenario

Fixes https://github.com/tektoncd/cli/issues/297

# Submitter Checklist

review them:

### DIFF
--- a/pkg/cmd/pipeline/start.go
+++ b/pkg/cmd/pipeline/start.go
@@ -158,7 +158,7 @@ func (opt *startOptions) getInput(pname string) error {
 		return err
 	}
 
-	if len(opt.Resources) == 0 {
+	if len(opt.Resources) == 0 && !opt.Last {
 		pres, err := getPipelineResources(cs.Tekton, opt.cliparams.Namespace())
 		if err != nil {
 			fmt.Fprintf(opt.stream.Err, "failed to list pipelineresources from %s namespace \n", opt.cliparams.Namespace())
@@ -173,7 +173,7 @@ func (opt *startOptions) getInput(pname string) error {
 		}
 	}
 
-	if len(opt.Params) == 0 {
+	if len(opt.Params) == 0 && !opt.Last {
 		err = opt.getInputParams(pipeline)
 		if err != nil {
 			return err
@@ -196,7 +196,7 @@ func (opt *startOptions) getInputResources(resources resourceOptionsFilter, pipe
 			{
 				Name: "pipelineresource",
 				Prompt: &survey.Select{
-					Message: fmt.Sprintf("Which %s resource to use for %s ?", res.Type, res.Name),
+					Message: fmt.Sprintf("Choose the %s resource to use for %s:", res.Type, res.Name),
 					Options: options,
 				},
 			},
--- a/pkg/cmd/pipeline/start_test.go
+++ b/pkg/cmd/pipeline/start_test.go
@@ -148,7 +148,6 @@ func Test_start_pipeline(t *testing.T) {
 		"-r=source=scaffold-git",
 		"-p=key1=value1",
 		"-s=svc1",
-		"-serviceaccounts=task1=svc1",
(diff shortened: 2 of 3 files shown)


pkg/cmd/pipeline/start_test.go
```diff
@@ -356,6 +356,80 @@ func Test_start_pipeline_last(t *testing.T) {
...
+
+	cs := pipelinetest.Clients{
+		Pipeline: pClient,
+		Kube:     fakekubeclientset.NewSimpleClientset(),
+	}
+
+	p := &test.Params{Tekton: cs.Pipeline, Kube: cs.Kube}
+
+	pipeline := Command(p)
+	got, _ := test.ExecuteCommand(pipeline, "start", pipelineName,
```
sthaha: are we testing the simplest case that  `-l` starts a new pipelinerun using the params, resources, sa from the latest `pipelinerun` ?

pkg/cmd/pipeline/start.go
```diff
@@ -158,25 +158,27 @@ func (opt *startOptions) getInput(pname string) error {
 		return err
 	}
 
-	if len(opt.Resources) == 0 {
-		pres, err := getPipelineResources(cs.Tekton, opt.cliparams.Namespace())
-		if err != nil {
-			fmt.Fprintf(opt.stream.Err, "failed to list pipelineresources from %s namespace \n", opt.cliparams.Namespace())
-			return err
-		}
+	if opt.Last != true {
```
sthaha: `!= true` => ` == false`

sthaha: I think this creates a bit too much nesting. can this be simplified?

pkg/cmd/pipeline/start_test.go
piyush-garg: yes, I have added test for this basic scenario now. Resource and Param from the previous run.

pkg/cmd/pipeline/start.go
piyush-garg: Improved a bit

piyush-garg:
Go.mod is coming from master

pkg/cmd/pipeline/start_test.go
```diff
@@ -356,6 +356,77 @@ func Test_start_pipeline_last(t *testing.T) {
...
+		"--last",
+		"-nns")
```
sthaha: nit: -n ns

pkg/cmd/pipeline/start_test.go
```diff
@@ -356,6 +356,77 @@ func Test_start_pipeline_last(t *testing.T) {
...
+	expected := "Pipelinerun started: random\n\nIn order to track the pipelinerun progress run:\ntkn pipelinerun logs random -f\n"
+	test.AssertOutput(t, expected, got)
+
+	pr, err := cs.Pipeline.TektonV1alpha1().PipelineRuns(p.Namespace()).Get("random", v1.GetOptions{})
+
+	if err != nil {
+		t.Errorf("Error getting pipelineruns %s", err.Error())
+	}
+
+	for _, v := range pr.Spec.Resources {
```
sthaha: nit: pretty sure there will be a better way to obtain the resource `"git-repo"` but that can be a different patch. 
@vdemeester

pkg/cmd/pipeline/start_test.go
piyush-garg: resolved this.

piyush-garg:
@sthaha These `go.mod` changes are coming from the latest master. My branch is up to date with the master.

Issue #311, opened 2019-09-12
Error Message if Namespace Does Not Exist With -n Flag


danielhelfand:
# Expected Behavior

If a namespace is provided that does not exist for any command that uses `-n`, the cli should notify the user the namespace does not exist. Simply listing that no items were found is misleading and may make the user think there are no items in the namespace. 

# Actual Behavior

```
$ tkn task ls -n doesnotexist
No tasks found
```

```
$ tkn pr ls -n doesnotexist
No pipelineruns found
```

# Steps to Reproduce the Problem

1. Run any command that uses `-n` (e.g. `tkn tr ls`) and enter a namespace that does not exist

chmouel:
I believe this is fixed in 2efd4e7b2baacc28910ecbdcba3e9634bdffd604 

@danielhelfand  feel free to reopen if there was more to it

/assign @danielhelfand 
/closes

danielhelfand:
/reopen

`2efd4e7` addressed #308, but did not address this issue. I can pick it up and look into it later this week or next.

chmouel:
Sounds good! thanks

afrittoli:
To be fair this is the same behaviour kubectl exposes - it just applies a filter and returns what matches. For instance in my cluster, if I filter by a non-existing namespace, I get cluster-wide resources:
```
$ kubectl get all -n foobar


NAME                                                               READY   REASON
clusterchannelprovisioner.eventing.knative.dev/in-memory           True
clusterchannelprovisioner.eventing.knative.dev/in-memory-channel   True

NAME                                                                                        READY   REASON
clusteringress.networking.internal.knative.dev/route-4bdb4e0b-6129-11e9-b22c-165770ee2c86   True
clusteringress.networking.internal.knative.dev/route-696e0dff-5afe-11e9-91a2-42406a53c343   True
clusteringress.networking.internal.knative.dev/route-6a3fb797-5afe-11e9-91a2-42406a53c343   True
clusteringress.networking.internal.knative.dev/route-7b36407a-50b4-11e9-8ed0-7a2460b12a0e   True
clusteringress.networking.internal.knative.dev/route-a981cb46-50b3-11e9-8ed0-7a2460b12a0e   True
clusteringress.networking.internal.knative.dev/route-aaca4ee9-50b3-11e9-8ed0-7a2460b12a0e   True
```

danielhelfand:
[...]

PR #316, opened 2019-09-17
Improve the error message on pipeline create

afrittoli:
# Changes

If the pipeline is not found on pipeline create, it may be due to
a misspelled pipeline name or namespace name. Include both in the
error message to help troubleshooting.

# Submitter Checklist

review them:

- [n/a] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)

for more details._

### DIFF
--- a/pkg/cmd/pipeline/start.go
+++ b/pkg/cmd/pipeline/start.go
@@ -34,7 +34,7 @@ import (
 
 var (
 	errNoPipeline      = errors.New("missing pipeline name")
-	errInvalidPipeline = "invalid pipeline name %s in namespace %s"
+	errInvalidPipeline = "pipeline name %s does not exist in namespace %s"
 )
 
 const (
@@ -368,7 +368,7 @@ func (opt *startOptions) startPipeline(pName string) error {
 	}
 
 	fmt.Fprintf(opt.stream.Out, "Pipelinerun started: %s\n\n"+
-		"In order to track the pipelinerun progress run:\ntkn pipelinerun logs -n %s %s -f\n", prCreated.Name, prCreated.Namespace, prCreated.Name)
+		"In order to track the pipelinerun progress run:\ntkn pipelinerun logs %s -f -n %s\n", prCreated.Name, prCreated.Name, prCreated.Namespace)
 	return nil
 }
 
--- a/pkg/cmd/pipeline/start.go
+++ b/pkg/cmd/pipeline/start.go
@@ -34,7 +34,7 @@ import (
 
 var (
 	errNoPipeline      = errors.New("missing pipeline name")
-	errInvalidPipeline = errors.New("invalid pipeline name")
+	errInvalidPipeline = "invalid pipeline name %s in namespace %s"
 )
 
 const (
@@ -76,7 +76,7 @@ func NameArg(args []string, p cli.Params) error {
 	name, ns := args[0], p.Namespace()
 	_, err = c.Tekton.TektonV1alpha1().Pipelines(ns).Get(name, metav1.GetOptions{})
 	if err != nil {
-		return errInvalidPipeline
+		return fmt.Errorf(errInvalidPipeline, name, ns)
 	}
 
 	return nil
--- a/pkg/cmd/pipeline/start_test.go
+++ b/pkg/cmd/pipeline/start_test.go
@@ -119,7 +119,7 @@ func Test_start_pipeline_not_found(t *testing.T) {
 
 	pipeline := Command(p)
 	got, _ := test.ExecuteCommand(pipeline, "start", "test-pipeline-2", "-n", "ns")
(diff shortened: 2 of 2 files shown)


chmouel:
I think #311 is related would be nice if we get this fixed everywhere,

cc @danielhelfand

afrittoli:
> I think #311 is related would be nice if we get this fixed everywhere,

I agree that it would be nice to fix other error messages, I was a bit lazy - only fixed the one I encountered :P 

About #311, I think the list and create commands are a bit different use cases. For the create, `tkn` looks for a specific pipeline specified via CLI parameters - and it may not find it - thus the *error* message.

For the list case, we don't expect anything specific, so if we don't get anything we can get a message that nothing was found, but that's for information only. To be honest my preference there would be to get no message at all, or if we do get a message get it on stderr - so that it's possible to reliably parse the stdout if needed.

pkg/cmd/pipeline/start.go
```diff
@@ -33,7 +33,7 @@ import (
 
 var (
 	errNoPipeline      = errors.New("missing pipeline name")
-	errInvalidPipeline = errors.New("invalid pipeline name")
+	errInvalidPipeline = "invalid pipeline name %s in namespace %s"
```
danielhelfand: Would this read better as `pipeline name %s does not exist in namespace %s`?

danielhelfand:
I think #311 should have a complete different error message that is specific to the namespace. This is more specific to the pipeline name not existing in a valid namespace, but I do think it more clearly communicates the issue to the user.

chmouel:
let me know if that's good to go for you guys ?

pkg/cmd/pipeline/start.go
afrittoli: Fixed

afrittoli:
I've been looking at different commands, and the errors returned a bit mixed up in style:

On pipeline commands:
* Start: only custom error (the one fixed in this PR)
* Delete: custom error + original client error
* Describe: original client error only
* Log: original client error only
* List: log a custom error and return the original error

On task commands:
* Delete: original client error
* List: log a custom error and return the original error

There are more commands for other resources too, I did not check them all; but it looks to me as if it would be best to agree on a common approach for the error message returned by commands, and then align them all. We could have one shared function used for logging for instance.

My proposal would be:
* normally only print a custom error message, which includes name and namespace of the resource
* in verbose mode (--verbose|-v) option passed, also print the original tekton client error

pkg/cmd/pipeline/start.go
```diff
@@ -33,7 +33,7 @@ import (
 
 var (
 	errNoPipeline      = errors.New("missing pipeline name")
-	errInvalidPipeline = errors.New("invalid pipeline name")
+	errInvalidPipeline = "pipeline name %s does not exist in namespace %"
```
piyush-garg: nit: namespace %s

pkg/cmd/pipeline/start_test.go
```diff
@@ -117,7 +117,7 @@ func Test_start_pipeline_not_found(t *testing.T) {
 
 	pipeline := Command(p)
 	got, _ := test.ExecuteCommand(pipeline, "start", "test-pipeline-2", "-n", "ns")
-	expected := "Error: " + errInvalidPipeline.Error() + "\n"
+	expected := "Error: " + fmt.Sprintf(errInvalidPipeline, "test-pipeline-2", "ns") + "\n"
```
piyush-garg: its better to use exact message, not fmt.Sprintf, you may have test failed for the typo here if dont this way

pkg/cmd/pipeline/start.go
```diff
@@ -367,7 +367,7 @@ func (opt *startOptions) startPipeline(pName string) error {
 	}
 
 	fmt.Fprintf(opt.stream.Out, "Pipelinerun started: %s\n\n"+
-		"In order to track the pipelinerun progress run:\ntkn pipelinerun logs %s -f\n", prCreated.Name, prCreated.Name)
+		"In order to track the pipelinerun progress run:\ntkn pipelinerun logs -n %s %s -f\n", prCreated.Name, prCreated.Namespace, prCreated.Name)
```
piyush-garg: not sure, but I think `tkn pipelinerun logs name -f -n namespace` will give better readability

pkg/cmd/pipeline/start.go
afrittoli: upz, nice catch, thank you!

pkg/cmd/pipeline/start_test.go
afrittoli: heh, indeed

pkg/cmd/pipeline/start.go
afrittoli: done

PR #340, opened 2019-09-30
Cleanup -l flag across commands

chmouel:
# Changes

We has an inconsistentie across commands for the `-l` flag, it will sometime
mean for some commands as `--limit` or sometime behaving for some other command
as `--last`.

I have moved last-log to `-L` and removed the short option for limits.

We keep `-l` for a future work with the --label filter to be consistent on
kubectl and other kubernetes based cli as `-l`  see tektoncd/cli#339.

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [X] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [X] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

# Release Notes

```
* Renamed the last log short flag to -L instead of -l
* Removed the short flag -l as --limit
```

PR #342, opened 2019-10-02
Allow passing labels to start

### LINKED ISSUE #339: Add a flag to pass labels when starting a Pipeline
# Expected Behavior

Users should be able to pass **[labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/)** when starting a `Pipeline` with the `tkn pipeline start pipeline` command.

# Actual Behavior

It is not possible to specify any label.
```
tkn pipeline start --help                                                                                                                                                                    
Start pipelines

Usage:
tkn pipeline start pipeline [RESOURCES...] [PARAMS...] [SERVICEACCOUNT] [flags]

Aliases:
  start, trigger

Examples:

# start pipeline foo by creating a pipelinerun named "foo-run-xyz123" from the namespace "bar"
tkn pipeline start foo -s ServiceAccountName -n bar

For params value, if you want to provide multiple values, provide them comma separated
like cat,foo.bar


Flags:
  -h, --help                          help for start
  -l, --last                          re-run the pipeline using last pipelinerun values
  -p, --param strings                 pass the param as key=value
  -r, --resource strings              pass the resource name and ref as name=ref
  -s, --serviceaccount string         pass the serviceaccount name
      --task-serviceaccount strings   pass the service account corresponding to the task

Global Flags:
  -n, --namespace string    namespace to use (default: from $KUBECONFIG)
  -C, --nocolour            disable colouring (default: false)
```
# Additional Info

In the same idea as the `-n` flag, I think it would have been great to use the `-l` flag like[ the `kubectl` command](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#api). But this flag is already used for the `last` feature.

chmouel:
# Changes

Add -l flag for start to pass labels to pipelineruns (and then to pods)

Closes #339

# Submitter Checklist

review them:

- [X] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)

# Release Notes

```
Allow passing labels when starting a pipeline with the `-l` option
```

### DIFF
--- a/pkg/cmd/pipeline/start.go
+++ b/pkg/cmd/pipeline/start.go
@@ -40,6 +40,7 @@ var (
 const (
 	invalidResource = "invalid input format for resource parameter : "
 	invalidParam    = "invalid input format for param parameter : "
+	invalidLabel    = "invalid input format for label parameter : "
 	invalidSvc      = "invalid service account parameter: "
 )
 
@@ -52,6 +53,7 @@ type startOptions struct {
 	ServiceAccountName string
 	ServiceAccounts    []string
 	Last               bool
+	Labels             []string
 }
 
 type resourceOptionsFilter struct {
@@ -132,6 +134,7 @@ like cat,foo.bar
 	c.Flags().StringSliceVar(&opt.ServiceAccounts, "task-serviceaccount", []string{}, "pass the service account corresponding to the task")
 	flags.AddShellCompletion(c.Flags().Lookup("task-serviceaccount"), "__kubectl_get_serviceaccount")
 	c.Flags().BoolVarP(&opt.Last, "last", "L", false, "re-run the pipeline using last pipelinerun values")
+	c.Flags().StringSliceVarP(&opt.Labels, "labels", "l", []string{}, "pass labels as label=value.")
 
 	_ = c.MarkZshCompPositionalArgumentCustom(1, "__tkn_get_pipeline")
 
@@ -350,6 +353,10 @@ func (opt *startOptions) startPipeline(pName string) error {
 		return err
 	}
 
+	if err := mergeLabels(pr, opt.Labels); err != nil {
+		return err
+	}
+
 	if err := mergeParam(pr, opt.Params); err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


pkg/cmd/pipeline/start.go
```diff
@@ -394,6 +401,19 @@ func mergeRes(pr *v1alpha1.PipelineRun, optRes []string) error {
...
+func mergeLabels(pr *v1alpha1.PipelineRun, labelPar []string) error {
+	labels, err := parseLabels(labelPar)
+	if err != nil {
+		return err
+	}
+	if len(labels) == 0 {
+		return nil
+	}
+
+	pr.ObjectMeta.Labels = labels
```
piyush-garg: Just a thought, this will remove the old labels. I think we should keep the old labels and add new ones. In the case of repeat, we keep the new one.

chmouel: you are right, and i was just being lazy! thanks mister senior principal to keep me straight! 👍

pkg/cmd/pipeline/start.go
```diff
@@ -132,6 +134,7 @@ like cat,foo.bar
 	c.Flags().StringSliceVar(&opt.ServiceAccounts, "task-serviceaccount", []string{}, "pass the service account corresponding to the task")
 	flags.AddShellCompletion(c.Flags().Lookup("task-serviceaccount"), "__kubectl_get_serviceaccount")
 	c.Flags().BoolVarP(&opt.Last, "last", "L", false, "re-run the pipeline using last pipelinerun values")
```
vdemeester: Didn't we want to remove the shorthand flag for `--last` ?

chmouel: yeah we moved it to -L  https://github.com/tektoncd/cli/pull/340 and unified labels to -l as kubectl and other kube tools,

vdemeester: fair enough :stuck_out_tongue:

pkg/cmd/pipeline/start.go
piyush-garg: You troll nicely :smile:

PR #349, opened 2019-10-04
Check if Namespace Exists for Task List Command

danielhelfand:
This pull request is a proposal to help deal with #311. It adds an error message to the CLI to check whether the namespace specified by the user exists on his or her cluster. 

Assuming this is a valid proposal, I can begin implementing this same change for other locations in the CLI that use `-n`.

# Changes

Adds a check to `task/list.go` to see whether a namespace exists. If it does not, an error message is returned: `"Error: namespaces \"foo\" not found\n"`. The use of `namespaces` versus non-plural `namespace` in the error message is [not considered an issue](https://github.com/kubernetes/client-go/issues/696) for client-go. 

# Submitter Checklist

review them:

for more details._

# Release Notes

```
Adds error message to tkn task ls when a namespace does not exist
```

### DIFF
--- a/pkg/cmd/task/list.go
+++ b/pkg/cmd/task/list.go
@@ -76,6 +76,12 @@ func printTaskDetails(s *cli.Stream, p cli.Params) error {
 		return err
 	}
 
+	// Check if namespace exists. Return error if namespace specified with -n doesn't exist or if user doesn't have permissions to view.
+	_, err = cs.Kube.CoreV1().Namespaces().Get(p.Namespace(), metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
 	tasks, err := listAllTasks(cs.Tekton, p.Namespace())
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Failed to list tasks from %s namespace \n", p.Namespace())
--- a/pkg/test/cobra.go
+++ b/pkg/test/cobra.go
@@ -33,6 +33,7 @@ func ExecuteCommandC(c *cobra.Command, args ...string) (*cobra.Command, string, 
 	buf := new(bytes.Buffer)
 	c.SetOutput(buf)
 	c.SetArgs(args)
+	c.SilenceUsage = true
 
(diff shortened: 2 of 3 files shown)


pkg/cmd/task/list.go
```diff
@@ -76,6 +76,13 @@ func printTaskDetails(s *cli.Stream, p cli.Params) error {
 		return err
 	}
 
+	// Check if namespace exists. Return error if namespace specified with -n doesn't exist.
+	_, err = cs.Kube.CoreV1().Namespaces().Get(p.Namespace(), metav1.GetOptions{})
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error: Namespace \"%s\" does not exist. Please enter a valid namespace.\n", p.Namespace())
```
piyush-garg: I think we need to handle the case of `if the user doesn't have access to that namespace`. In that case, this error message may be misleading.

danielhelfand: It may be better to just return `err` here as what ever is returned will be displayed to the user. I ran into issues with testing by returning `err` originally since it appears to display the usage as part of the output. I'll take another shot at that approach.

pkg/test/cobra.go
```diff
@@ -33,6 +33,7 @@ func ExecuteCommandC(c *cobra.Command, args ...string) (*cobra.Command, string,
 	buf := new(bytes.Buffer)
 	c.SetOutput(buf)
 	c.SetArgs(args)
+	c.SilenceUsage = true
```
danielhelfand: Adding this since `tkn` already has `SilenceUsage` set to true in [`root.go`](https://github.com/tektoncd/cli/blob/604f6cf20f2a7fad2ecbe523ac57b0ea4826b2ab/pkg/cmd/root.go#L71). Tests should be the same output.

pkg/cmd/task/list.go
danielhelfand: This should now display RBAC errors as well with most recent push.

PR #350, opened 2019-10-08
Add Namespace Check to Task Delete Command

danielhelfand:
This pull request follows up #349 and also helps to address #311. The goal is to check whether a namespace exists to alert the user if the namespace provided is invalid. 

# Submitter Checklist

review them:

# Release Notes

```
Adds error message to tkn task delete when a namespace does not exist
```

### DIFF
--- a/pkg/cmd/task/delete.go
+++ b/pkg/cmd/task/delete.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/tektoncd/cli/pkg/cli"
+	validate "github.com/tektoncd/cli/pkg/helper/validate"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	cliopts "k8s.io/cli-runtime/pkg/genericclioptions"
 )
@@ -58,6 +59,15 @@ tkn t rm foo -n bar
 				In:  cmd.InOrStdin(),
 				Out: cmd.OutOrStdout(),
 				Err: cmd.OutOrStderr(),
+			}
+
+			cs, err := p.Clients()
+			if err != nil {
+				return fmt.Errorf("failed to create tekton client")
+			}
+
+			if err := validate.NamespaceExists(cs.Kube, p.Namespace()); err != nil {
+				return err
 			}
 
 			if err := checkOptions(opts, s, p, args[0]); err != nil {
--- a/pkg/cmd/task/list.go
+++ b/pkg/cmd/task/list.go
@@ -23,6 +23,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/tektoncd/cli/pkg/cli"
 	"github.com/tektoncd/cli/pkg/formatted"
+	validate "github.com/tektoncd/cli/pkg/helper/validate"
 	"github.com/tektoncd/cli/pkg/printer"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
 	"github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
@@ -48,6 +49,15 @@ func listCommand(p cli.Params) *cobra.Command {
 			"commandType": "main",
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
+			cs, err := p.Clients()
+			if err != nil {
+				return fmt.Errorf("failed to create tekton client")
+			}
+
+			if err := validate.NamespaceExists(cs.Kube, p.Namespace()); err != nil {
+				return err
+			}
+
 			output, err := cmd.LocalFlags().GetString("output")
 			if err != nil {
 				fmt.Fprint(os.Stderr, "Error: output option not set properly \n")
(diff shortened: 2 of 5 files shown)


pkg/cmd/task/delete.go
```diff
@@ -81,6 +81,12 @@ func deleteTask(opts *deleteOptions, s *cli.Stream, p cli.Params, tName string)
 		return fmt.Errorf("failed to create tekton client")
 	}
 
+	// Check if namespace exists. Return error if namespace specified with -n doesn't exist or if user doesn't have permissions to view.
```
sthaha: imho, these validations should be done before calling `checkOptions` (asks for user input) otherwise the UX will be that they issue a `task delete -n foobar task-name` and `checkOptions` would ask for confirmation and then fail saying there is no namespace `foobar` where as we we could do the validation before invoking `checkOptions` it would fail soon after they run the `task delete -n foobar ....` .
 
I am "fine" with this implementation but if you feel motivated, could you consider moving the validations to a different function?

danielhelfand: @sthaha Yeah sure. I see your point and think it makes sense. So basically move the validation up to before `checkOptions`? 

I don't think that should change the behavior here, and I can also correct this then for [`list.go`](https://github.com/tektoncd/cli/blob/aea08692f3b6d4fdc81866d456157ea17247e491/pkg/cmd/task/list.go#L80) since we already merged in this validation for `tkn task list`.

danielhelfand: I pushed the validation before `checkOptions` in latest push, but feel free to still recommend a better location.

pkg/cmd/task/delete.go
```diff
@@ -75,11 +86,7 @@ tkn t rm foo -n bar
 	return c
 }
 
-func deleteTask(opts *deleteOptions, s *cli.Stream, p cli.Params, tName string) error {
-	cs, err := p.Clients()
-	if err != nil {
-		return fmt.Errorf("failed to create tekton client")
-	}
+func deleteTask(opts *deleteOptions, s *cli.Stream, p cli.Params, cs *cli.Clients, tName string) error {
```
sthaha: isn't cs already in `p cli.Params` ?

pkg/cmd/task/delete.go
```diff
@@ -60,11 +60,22 @@ tkn t rm foo -n bar
 				Err: cmd.OutOrStderr(),
 			}
 
+			cs, err := p.Clients()
+			if err != nil {
+				return fmt.Errorf("failed to create tekton client")
+			}
+
+			// Check if namespace exists. Return error if namespace specified with -n doesn't exist or if user doesn't have permissions to view.
```
sthaha: it may be better to move these validations to a package `validate`  so that the call appears

```
if err := validate.NamespaceExists(kubeClient, ns); err != nil{
  return err
}
```
WDYT?

pkg/cmd/task/delete.go
danielhelfand: Yes it is. I'll point on in a review comment why I went with this approach, but the gist of it was I was trying to avoid checking for the `"failed to create tekton client"` error twice when using `p.Clients()`.

pkg/cmd/task/delete.go
danielhelfand: I have moved the validation code under `pkg/helper/validate`. Let me know what you think.

pkg/helper/validate/validate.go
```diff
@@ -0,0 +1,32 @@
+// Copyright © 2019 The Tekton Authors.
```
danielhelfand: Idea here would be to use this class for common validation methods across the CLI.

pkg/cmd/task/list.go
```diff
@@ -70,16 +80,9 @@ func listCommand(p cli.Params) *cobra.Command {
 }
 
 func printTaskDetails(s *cli.Stream, p cli.Params) error {
-
 	cs, err := p.Clients()
 	if err != nil {
```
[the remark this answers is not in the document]
sthaha: Agree, I was also thinking of 
```go 

validate.All( validate.Clients(p), validate.NamespaceExists(p)) 
```
where `validate.KubeClient(p)` returns a `validator function`  like: ` func validator() error ` and `validate.All` just checks none of the functions return an error` 

That way all commands would just have to invoke just

``` go
if   err := validate.All( validate.Clients(p), validate.Foobar(x, y, z)); err != nil {
   return err
}
```
 WDYT? 

BTW, I am not suggesting this change in this PR :)

pkg/cmd/task/delete.go
sthaha: IMHO,  ... it is okay for internal (non exported methods) to make assumptions about arguments it receives ; i.e. the caller should have ensured the arguments are valid before passing it to callee so that certain step need to be performed only once ... (premature optimisation 🤷‍♂️ ; may be ...)

pkg/cmd/task/list.go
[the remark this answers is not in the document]
danielhelfand: Sure, I think that's really clever. I'll keep this in mind a try to refactor it.

PR #352, opened 2019-10-09
Set namespace in e2e tests

piyush-garg:
As we have added a check for namespace exist
in commands, we should hit commands from a
defined namespace in e2e tests

This will set namespace in e2e tests

Fix typo mistakes

### DIFF
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -65,7 +65,8 @@ must_fail  "logs of taskrun before installing crd" tkn taskrun logs foo
 install_pipeline_crd
 
 # listing objects after CRD is created should not fail
-run_test  "list pipeelinerun for pipeline foo" tkn taskrun list foo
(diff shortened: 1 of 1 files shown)


test/e2e-tests.sh
```diff
@@ -65,7 +65,8 @@ must_fail  "logs of taskrun before installing crd" tkn taskrun logs foo
 install_pipeline_crd
 
 # listing objects after CRD is created should not fail
-run_test  "list pipeelinerun for pipeline foo" tkn taskrun list foo
+kubectl config set-context $(kubectl config current-context) --namespace=default
```
chmouel: can you test this on openshift CI ?

piyush-garg: yes, tested https://github.com/openshift/tektoncd-cli/pull/35

chmouel: thanks

PR #418, opened 2019-10-28
Check if TaskRef Exists for TaskRun Describe

### LINKED ISSUE #380: Describe task does not work for inline task spec
# Expected Behavior

```
tkn taskrun describe taskrun_name
```

This should show the description of taskrun

# Actual Behavior

```
tkn taskrun describe task-59p7t          
Failed to execute templateError: template: Describe taskrun:3:21: executing "Describe taskrun" at <.TaskRun.Spec.TaskRef.Name>: nil pointer evaluating *v1alpha1.TaskRef.Name
```

1. Create a taskrun with inline spec like

```
kind: TaskRun
metadata:
  generateName: task-
spec:
  taskSpec:
    steps:
      - name: test
        image: ubuntu
        command: ["/bin/bash"]
        args:
          - -c
          - |
            set -e
            set -x
            echo hello world
```
2. Get the taskrun name
```
tkn taskrun list           
NAME         STARTED         DURATION     STATUS      
task-59p7t   2 minutes ago   13 seconds   Succeeded
```
3. Hit describe command
```
tkn taskrun describe task-59p7t          
Failed to execute templateError: template: Describe taskrun:3:21: executing "Describe taskrun" at <.TaskRun.Spec.TaskRef.Name>: nil pointer evaluating *v1alpha1.TaskRef.Name
```

# Additional Info

CLI version 0.4.0

danielhelfand:
Closes #380 

This is a proposal to address using `tkn taskrun describe` when there is no `TaskRef` associated with the task run. This pr adds a method to check whether the field is present on the `TaskSpec`. 

# Submitter Checklist

review them:

for more details._

# Release Notes

```
Check whether a TaskRef is present for a task run as part of tkn tr desc
```


piyush-garg:
I think this is the case with Taskref in taskrun, Pipelineref in pipelinerun, resourcespec in both taskrun and pipelinerun.

We need to handle all and also in some way like in case of resourceref, if we can show the details that will be good

danielhelfand:
> I think this is the case with Taskref in taskrun, Pipelineref in pipelinerun, resourcespec in both taskrun and pipelinerun.
> 
> We need to handle all and also in some way like in case of resourceref, if we can show the details that will be good

I can certainly cover the other cases with `nil` checks in `fomatted/tekton.go` and also add resource details. Can we perhaps split these into different issues though? One for adding `nil` checks and one for showing details?

pkg/cmd/taskrun/describe.go
```diff
@@ -32,7 +32,7 @@ import (
 
 const templ = `Name:	{{ .TaskRun.Name }}
 Namespace:	{{ .TaskRun.Namespace }}
-Task Ref:	{{ .TaskRun.Spec.TaskRef.Name }}
+Task Ref:    {{ taskRefExists .TaskRun.Spec }}
```
chmouel: would it be better to just skip showing `Task Ref:` label if the task ref is not present  (instead of showing that it isnt
) ?

pkg/cmd/taskrun/describe_test.go
```diff
@@ -279,3 +279,66 @@ No steps
 
 	test.AssertOutput(t, expected, actual)
 }
+
+func TestTaskRunDescribe_no_taskref(t *testing.T) {
```
chmouel: Is the test with a taskref tested implicitely?

pkg/cmd/taskrun/describe.go
danielhelfand: Yeah, agreed. We can just test if the result of `taskRefExists` is an empty string and show nothing if it's blank.

pkg/cmd/taskrun/describe_test.go
danielhelfand: Every other test uses a `TaskRef` so that case is covered.

danielhelfand:
Following up here, I do not believe it is necessary to add nil checks for `PipelineRef` and `ResourceRef`. `TaskRef` appears to be a unique case that can be nil, and that appears to be causing the issue.

vdemeester:
> Following up here, I do not believe it is necessary to add nil checks for `PipelineRef` and `ResourceRef`. `TaskRef` appears to be a unique case that can be nil, and that appears to be causing the issue.

You are right but I think this is an oversight of `tektoncd/pipeline`, or at least it is not consistent between `TaskRun` and `PipelineRun`. In any case, we may also want to not display `PipelineRef` if it's empty (it's not gonna be `nil` but it might be empty)

danielhelfand:
> > Following up here, I do not believe it is necessary to add nil checks for `PipelineRef` and `ResourceRef`. `TaskRef` appears to be a unique case that can be nil, and that appears to be causing the issue.
> 
> You are right but I think this is an oversight of `tektoncd/pipeline`, or at least it is not consistent between `TaskRun` and `PipelineRun`. In any case, we may also want to not display `PipelineRef` if it's empty (it's not gonna be `nil` but it might be empty)

@vdemeester Yes, agreed on the oversight. It might be worthwhile to report this to pipelines. 

The current approach here does not display `PipelineRef` when it's empty though: 

```
{{- if ne .PipelineRun.Spec.PipelineRef.Name "" }}
Pipeline Ref:	{{ .PipelineRun.Spec.PipelineRef.Name }}
{{- end }}
```

And there's a test case that's been added to cover this (i.e. `TestPipelineRunDescribe_without_pipelineref`).

vdemeester:
@danielhelfand https://github.com/tektoncd/pipeline/pull/1497 and https://github.com/tektoncd/pipeline/pull/1495 :stuck_out_tongue:

danielhelfand:
@vdemeester Thanks! I'll create an issue for updating this when we update the version of pipelines.

pkg/helper/validate/validate.go
```diff
@@ -41,3 +46,13 @@ func NamespaceExists(p params) error {
 
 	return nil
 }
+
+// Check if TaskRef exists on a TaskRunSpec. Returns empty string if not present.
+func TaskRefExists(spec v1alpha1.TaskRunSpec) string {
+
+	if spec.TaskRef == nil {
+		return fieldNotPresent
```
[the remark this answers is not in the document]
danielhelfand: I would personally prefer to keep it to communicate what an empty string represents.

chmouel:
/lgtm
/approve
/meow

Thanks 🙇🏽‍♂️😻

PR #450, opened 2019-11-11
Show Status of Steps with TaskRun Describe

### LINKED ISSUE #440: Show Step Status for TaskRun Describe
`tkn tr describe tr_name` will show the names of steps as part of a task run, but it does not show the status of the steps. This was originally excluded from #303 because of not fully knowing how to check whether the step status was available yet.

# Expected Behavior

Show step status of steps for task run describe command:

```
Steps
NAME    STATUS
step1   Succeeded
step2   Running
```

# Actual Behavior

```
Steps
NAME
step1
step2
```

Run `tkn tr describe` with any taskrun

danielhelfand:
Closes #440 

This pull request adds the ability to see the status of a step with `tkn tr describe`. 

Possible values are `Completed` and `Error` for a step from `Terminated`. `Completed` does not always imply the step ran successfully. It represents that no error occurred for the step, or the step never ran due to a previous step failure.

`PodInitializing` will appear when the step is in a `Waiting` state. 

`Running` will appear when the step is actually executing and also when the pod has been initialized for the step. I am open to suggestions on if there's a better message that could be displayed to end users.

`STATUS` is mapped to the property `Reason` on the `StepState` from `Terminated` or `Waiting`, but will be `Running` if the `Running` property on the `StepState` isn't `nil`.

Limitation of approach in this pr is that steps are not ordered by start time. It may take some planning around how to deal with `StartedAt` not always being present for a step.

# Submitter Checklist

review them:

# Release Notes

```
Show step status with taskrun describe
```

### DIFF
--- a/pkg/helper/validate/validate.go
+++ b/pkg/helper/validate/validate.go
@@ -24,6 +24,7 @@ import (
 
 const (
 	fieldNotPresent = ""
+	pendingState    = "---"
 )
 
 type params interface {
@@ -55,4 +56,23 @@ func TaskRefExists(spec v1alpha1.TaskRunSpec) string {
 	}
 
 	return spec.TaskRef.Name
+}
+
+// Check if step is in waiting, running, or terminated state by checking StepState of the step.
+func StepReasonExists(state v1alpha1.StepState) string {
+
+	if state.Waiting == nil {
+
+		if state.Running != nil {
+			return "Running"
+		}
+
+		if state.Terminated != nil {
+			return state.Terminated.Reason
+		}
+
+		return pendingState
+	}
+
+	return state.Waiting.Reason
 }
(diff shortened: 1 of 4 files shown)


danielhelfand:
Changing to `WIP` as I am now seeing `Waiting` and `Running` aspects of `StepState` that should be taken into account.

pkg/helper/validate/validate.go
```diff
@@ -56,3 +57,22 @@ func TaskRefExists(spec v1alpha1.TaskRunSpec) string {
...
 	return spec.TaskRef.Name
 }
+
+// Check if step is in waiting, running, or terminated state by checking StepState of the step.
+func StepReasonExists(state v1alpha1.StepState) string {
+
+	if state.Waiting == nil {
+
+		if state.Running != nil {
+			return "Running"
```
danielhelfand: Open to suggestions is there's a better message that can be returned here. The issue centers around that this will display when the container is not actually executing. `state.Running != nil` when the pod is initialized and container is ready.

danielhelfand:
I have updated the pr description and have updated the pr to take into account `Running` and `Waiting` for the `StepStatus`.

pkg/helper/validate/validate.go
vdemeester: @danielhelfand yeah for now this is the best we can do I think…

PR #469, opened 2019-11-18
Include TaskRun Failure Message in Logs

### LINKED ISSUE #467: Error Message for tkn task start Difficult to Find
When using `tkn task start`, logs instantly start as expected, but a failure results in the following message:

```
Error: task build-docker-image-from-git-source create failed or has not started yet or pod for task not yet available
```

This doesn't do the best job of communicating the issue if a failure has occurred. You need to dig deeper by looking at `tkn tr desc` to find what the error was:

```
Name:        build-docker-image-from-git-source-run-g6bfb
Namespace:   default
Task Ref:    build-docker-image-from-git-source

Status
STARTED          DURATION    STATUS
14 minutes ago   ---         Failed(TaskRunValidationFailed)

Message
invalid input resources: TaskRun's declared resources didn't match usage in Task: Didn't provide required values: [docker-source]

Input Resources
No resources

Output Resources
No resources

Params
No params

Steps
No steps
```

It would be better if this information was communicated via logs of the task run, but, at the very least, we could print some type of message of where to look for the error message (i.e. `tkn taskrun desc`).

# Expected Behavior

Find specific error message in task run logs

# Actual Behavior

Error in logs doesn't communicate issue for taskrun failure

Run task with inputs/outputs
1. Create task with input
2. `tkn task start <task name> #Don't provide inputs/outputs

danielhelfand:
Closes #467 

This pull request adds a check if the taskrun has failed while starting up and includes the failure message in the taskrun logs.

# Submitter Checklist

review them:

# Release Notes

```
Make error message for taskrun start up failure easier to find
```

### DIFF
--- a/pkg/cmd/taskrun/describe.go
+++ b/pkg/cmd/taskrun/describe.go
@@ -184,5 +184,6 @@ func hasFailed(tr *v1alpha1.TaskRun) string {
 	if tr.Status.Conditions[0].Status == corev1.ConditionFalse {
 		return tr.Status.Conditions[0].Message
 	}
+
 	return ""
 }
--- a/pkg/cmd/taskrun/log_reader.go
+++ b/pkg/cmd/taskrun/log_reader.go
@@ -16,6 +16,7 @@ package taskrun
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -27,6 +28,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
+	"knative.dev/pkg/apis/duck/v1beta1"
 )
 
 type step struct {
@@ -109,7 +111,7 @@ func (lr *LogReader) readLiveLogs() (<-chan Log, <-chan error, error) {
 	p := pods.New(podName, lr.Ns, kube, lr.Streamer)
 	pod, err := p.Wait()
 	if err != nil {
-		return nil, nil, errors.Wrap(err, fmt.Sprintf("task %s failed", lr.Task))
+		return nil, nil, errors.New(fmt.Sprintf("task %s failed: %s. Run tkn tr desc %s for more details.", lr.Task, strings.TrimSpace(err.Error()), tr.Name))
 	}
 
 	steps := filterSteps(pod, lr.AllSteps)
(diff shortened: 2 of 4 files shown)


piyush-garg:
So my only concern with this is, it will work only for `follow` mode. Is that the target?

danielhelfand:
> So my only concern with this is, it will work only for `follow` mode. Is that the target?

No, you're right. We should also do this for non follow mode. It would make sense to edit [here](https://github.com/tektoncd/cli/blob/d1c9dfbd4583c3f2d00c82c0e92c5cade04b93c4/pkg/cmd/taskrun/log_reader.go#L137) to display the error.

piyush-garg:
Better to do something for both, and the differentiation starts [here](https://github.com/tektoncd/cli/blob/d1c9dfbd4583c3f2d00c82c0e92c5cade04b93c4/pkg/cmd/taskrun/log_reader.go#L73)

pkg/cmd/taskrun/log_reader.go
```diff
@@ -273,11 +279,25 @@ func (lr *LogReader) waitUntilPodNameAvailable(timeout time.Duration) (*v1alpha1
 			}
 			if first {
 				first = false
-				fmt.Fprintln(lr.Stream.Out, "Task still running ...")
+				fmt.Fprintln(lr.Stream.Out, "Waiting for logs to be available...")
```
danielhelfand: Updating message here as it can be misleading. For instance, if you run `tkn taskrun logs -f` after a task has failed, it would print `Task still running` even though it has already failed.

pkg/cmd/taskrun/log_reader.go
```diff
@@ -134,7 +140,7 @@ func (lr *LogReader) readAvailableLogs(tr *v1alpha1.TaskRun) (<-chan Log, <-chan
 	p := pods.New(podName, lr.Ns, kube, lr.Streamer)
 	pod, err := p.Get()
 	if err != nil {
-		return nil, nil, errors.Wrap(err, fmt.Sprintf("task %s failed", lr.Task))
+		return nil, nil, errors.Wrap(err, fmt.Sprintf("task %s has failed", lr.Task))
```
danielhelfand: Question here is whether to include `err.Error()` or `tr.Condition[0].message`?

pkg/cmd/taskrun/describe.go
```diff
@@ -184,5 +184,6 @@ func hasFailed(tr *v1alpha1.TaskRun) string {
 	if tr.Status.Conditions[0].Status == corev1.ConditionFalse {
 		return tr.Status.Conditions[0].Message
 	}
+
```
vdemeester: :thinking:

Issue #477, opened 2019-11-25
Can't specify the k8s context?


tsuna:
**tkn Version:** v0.5.1

`tkn` has a `--kubeconfig` flag but it doesn't seem to have a flag to specify which k8s context to use.  I have one config file and several contexts in it, and `tkn` only seems to use the default. Am I missing something or there is no way to tell it which context to use?

chmouel:
there is way if you use the variable KUBECONFIG but we can add a specific `--kubeconfig` flag

tsuna:
How do you specify the context to use with the `KUBECONFIG` variable?

vdemeester:
@chmouel there is no `--context` flag though (`kubectl` has one)

```
       --context=""
           The name of the kubeconfig context to use
```

chmouel:
Ah I stand corrected then, I usually would just use different KUBECONFIG and use kubens to change my default context in there, but that flag is much easier,

waveywaves:
I would like to work on this if that is alright

waveywaves:
@hrishin says /ilovethisissue will make it assign it to me

PR #481, opened 2019-11-27
Set current k8s context via tkn 

waveywaves:
# Changes

Fixes https://github.com/tektoncd/cli/issues/477
Allows the user to specify the kubecontext they would like to use the tekton cli in, given as an argument to the `--context` flag or its alias `-c`.
eg:
```
tkn pipelinerun yolo start --context homekube/kube:admin
tkn pipelinerun yolo start -c homekube/kube:admin

```

# Submitter Checklist

review them:

for more details._

# Release Notes

```
release-note
```

### DIFF
--- a/pkg/test/params.go
+++ b/pkg/test/params.go
@@ -22,11 +22,11 @@ import (
 )
 
 type Params struct {
-	ns, kc string
-	Tekton versioned.Interface
-	Kube   k8s.Interface
-	Clock  clockwork.Clock
-	Cls    *cli.Clients
+	ns, kConfig, kContext string
+	Tekton                versioned.Interface
+	Kube                  k8s.Interface
+	Clock                 clockwork.Clock
+	Cls                   *cli.Clients
 }
 
 var _ cli.Params = &Params{}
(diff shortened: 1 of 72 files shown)


chmouel:
/ok-to-test

Thanks vibhav, I think you are missing a test in there

waveywaves:
@chmouel I have updated `/pkg/test/params.go` Apart from that, I wasn't able to write the test for kubecontext as the return from the `config()` function is necessary for the test and the function itself is private, same goes for why the test for kubeconfig doesn't exist I suppose as it follows the same requirement.

pkg/test/params.go
```diff
@@ -22,11 +22,11 @@ import (
 )
 
 type Params struct {
-	ns, kc string
-	Tekton versioned.Interface
-	Kube   k8s.Interface
-	Clock  clockwork.Clock
-	Cls    *cli.Clients
+	ns, kc, c string
```
sthaha: `c` doesn't make it obvious what it is imho. How about kContext?

sthaha: and make kc -> kConfig

waveywaves: I have updated `c` to `kContext` and `kc` to `kConfig`.

chmouel:
/lgtm

thanks!!

Issue #524, opened 2019-12-17
Feature: `tkn task run` vs `tkn task start`


ImJasonH:
# Expected Behavior

I can "run" Tasks, not just "Start" them.

"Running" a Task results in a new "TaskRun" instance.

# Actual Behavior

In order to "run" a Task, I have to "Start" it.

When I "Start" a Task, it results in new "Task**Run**" instance.

# Additional Info

More concretely, I'd like to change the CLI to be able to call `tkn task run <task-name> --params...`, instead of `tkn task start ...`

If we decide to do this, we could support both for some amount of time, with a deprecation warning for `tkn task start`.

bobcatfish:
![image](https://user-images.githubusercontent.com/432502/71027048-74a1b180-20d8-11ea-85a9-29279b141dd8.png)

vdemeester:
btw, we can have both (aka `run` and `start`, one being an alias of the other, same way `docker run` and `docker container run` are aliases)

/kind question
/kind feature

danielhelfand:
As of now, it looks like we support [trigger](https://github.com/tektoncd/cli/blob/master/pkg/cmd/task/start.go#L95) as an alias for pipeline start. It's obviously pretty well hidden as it's not promoted very well:

```
$ tkn pipeline start -h

Start pipelines

Usage:
tkn pipeline start pipeline [RESOURCES...] [PARAMS...] [SERVICEACCOUNT] [flags]

Aliases:
  start, trigger

Examples:
  
# start pipeline foo by creating a pipelinerun named "foo-run-xyz123" from the namespace "bar"
tkn pipeline start foo -s ServiceAccountName -n bar

For params value, if you want to provide multiple values, provide them comma separated
like cat,foo,bar


Flags:
  -h, --help                          help for start
  -l, --labels strings                pass labels as label=value.
  -L, --last                          re-run the pipeline using last pipelinerun values
  -p, --param stringArray             pass the param as key=value or key=value1,value2
  -r, --resource strings              pass the resource name and ref as name=ref
  -s, --serviceaccount string         pass the serviceaccount name
      --showlog                       show logs right after starting the pipeline (default true)
      --task-serviceaccount strings   pass the service account corresponding to the task

Global Flags:
  -c, --context string      name of the kubeconfig context to use (default: kubectl config current-context)
  -k, --kubeconfig string   kubectl config file (default: $HOME/.kube/config)
  -n, --namespace string    namespace to use (default: from $KUBECONFIG)
[...]

PR #526, opened 2019-12-17
Rename `tkn pipeline start` to `tkn pipeline run`

ImJasonH:
`start` remains an alias for `run` for users who had been using that.

Part of #524

# Submitter Checklist

review them:

for more details._

# Release Notes

```
Rename `tkn pipeline start` to `tkn pipeline run` -- with an alias to continue supporting `start`
```


sthaha:
@ImJasonH would it better to just add `run` alias to the `start` command instead? If we did that instead of renaming `start` to `run` those script that depend on the `start` command need not change and we don't have to introduce a breaking change.

chmouel:
I don't mind either (start or run) but as @sthaha suggested, i'd rather see an alias than a breaking change,

vdemeester:
> @ImJasonH would it better to just add `run` alias to the `start` command instead? If we did that instead of renaming `start` to `run` those script that depend on the `start` command need not change and we don't have to introduce a breaking change.

There is an alias to `start` in the PR ([here](https://github.com/tektoncd/cli/pull/526/files#diff-3c87311eda1686abaaeba01f04ce8cbdR112)). A simplest change would be to keep the current `start.go` and add the `run` alias, the only thing then to "validate" is what should we generate for the documentation.

From https://github.com/tektoncd/cli/issues/524#issuecomment-566926653

> The question is then, which one should be the "documented" one, `run` or `start` ?  and which one is the alias. If we just want to be able to do `tkn task run`, then, an alias is enough. If we want "to move to `run`" (aka `man tkn-task-run`, and highlighting `run` instead of `start` in documentation), then a change making `start` becamo `run` (with an alias like #526 would be better). We can also enhance our doc/manpage generator to generate aliases commands to (aka for `tkn foo` that have a bar alias, generating `tkn-foo` and `tkn-bar` as manpage and cli reference doc).

sthaha:
Sorry about the confusion .. totally missed that an Alias was added for `run` as `start`.

vdemeester:
Actually, the more I think about that, the more I don't think `start` and `run` should be aliases, see https://github.com/tektoncd/cli/issues/524#issuecomment-566943727

ImJasonH:
Lots of new design changes for this being discussed in https://github.com/tektoncd/cli/issues/524 -- closing until we reach agreement.

PR #528, opened 2019-12-18
Set default to False for Show Logs on TaskRun and PipelineRun

### LINKED ISSUE #455: Set Default to False for Show Logs for TaskRun and PipelineRun
It's a bit unexpected to immediately show logs after starting a task run or a pipeline run. I realize that this was an intentional feature added for pipeline runs and task runs, but I think it's not fair to assume that users will want to immediately see logs with the CLI. 

I think the `--show-logs` flag gives users the ability to control whether they immediately want to see logs, but we should not assume they will by default. 

I wanted to open this up to discussion and see if there were any strong opinions as for why logs should be shown by default. 

# Expected Behavior

Starting a pipeline run or task run will not immediately show logs.

# Actual Behavior

Logs are shown.

chmouel:
Closes #455

```
We do not show the logs anymore right after starting the pipeline, use
--showlog to see it
```

### DIFF
--- a/docs/cmd/tkn_pipeline_start.md
+++ b/docs/cmd/tkn_pipeline_start.md
@@ -42,7 +42,7 @@ two parameters (foo and bar)
   -p, --param stringArray             pass the param as key=value or key=value1,value2
   -r, --resource strings              pass the resource name and ref as name=ref
   -s, --serviceaccount string         pass the serviceaccount name
-      --showlog                       show logs right after starting the pipeline (default true)
+      --showlog                       show logs right after starting the pipeline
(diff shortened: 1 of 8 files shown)


piyush-garg:
Also, should we remove `--showlog=false` from the tests ?

danielhelfand:
> Also, should we remove `--showlog=false` from the tests ?

It would be nice. With this change, I would think using `--showlog=false` would never really be needed. No issues if we want to do in a follow up pr though.

chmouel:
Sounds good, i removed them from the tests and regenerated the docs, thanks,

Issue #209, opened 2019-08-08
Edit Command for Pipeline, Resource, and Task Resources


danielhelfand:
This issue is a feature request for the ability to edit already created custom resource definitions for `tekton.dev` resources on a Kubernetes cluster. These resources include:
* `Pipeline`
* `Resource`
* `Task`

Currently, a user must use a second tool to edit these custom resources, such as `kubectl` or `oc`. Being able to do this directly through `tkn` would help with having a user not have to go back and forth with another tool while using `tkn`.

# Expected Behavior

The edit command should alter an already existing resource from a Kubernetes namespace. The user should be able to use a resource definition from a file, but it would also be helpful to have an interactive mode that allows the user to directly edit the resource definition via the CLI.

This interactive mode could be as simple as opening a text editor and allowing the user to save the definition, which would alter the resource on the cluster.

Example with `Pipeline`:

`tkn pipeline edit pipeline_name`
* This should use the name of the already created `pipeline` to alter the `pipeline` from the cluster namespace by opening an interactive editor

`tkn pipeline edit pipeline_resource.yaml`
* This should use the name of the already created `pipeline` to alter the `pipeline` from the cluster namespace by using the resource definition in the file

# Additional Info

Since these features are already implemented in other Kubernetes tools, the question is how what should `tkn` overlap with for these other tools in terms of functionality versus leave to the other tools completely. 

In my opinion, I think it makes sense in this case to have the ability to also delete through `tkn` along with other Kubernetes tools.

danielhelfand:
Please see #816 on why this is being closed.

PR #559, opened 2019-12-30
tkn task apply Command

danielhelfand:
Part of #209 

This pull request adds a `tkn task apply` command that adds the ability for a user to update a task that has already been created in a namespace. If the task does not exist in the namespace when using apply, the command will create the task.

This command supports both local and remote file definitions. Usage shown below:

`tkn task apply -f /local/file/path/task.yaml`

`tkn task apply -f https://www.github.com/task.yaml`

# Submitter Checklist

review them:

# Release Notes

```
Add tkn task apply command
```


danielhelfand:
/close

We need a non resource specific approach when using apply to support multi resource file definitions

Issue #574, opened 2020-01-07
Have Create and Apply Not Be Tied to Specific Resources


danielhelfand:
As of v0.6.0, create commands are structured as follows:

`tkn <resource> <action>` (e.g. `tkn pipeline create`)

When it comes to creating multiple resources in the same file definition, our current approach will not support that. The question is whether to support several different Tekton resources in the same file (e.g. a `Pipeline` and a `Task`) and whether to support this as a non resource-specific command (e.g. `tkn create`/`tkn apply`).

chmouel:
I think this should be at the topevel the same way we create and apply are toplevels in kubectl, 

i.e: we don't say `kubectl pods create` but `kubectl create pods`

danielhelfand:
Agreed. This would also allow us to support different kinds defined within the same file.

danielhelfand:
The current thought is to follow a [pattern used by `ko`](https://github.com/google/ko/blob/1fcfe62a9606a2781559ed10897a4ae7141d7628/pkg/commands/delete.go) to support top level `create` and `apply` commands through `tkn` that actually invoke `kubectl` commands. 

I will put together a proposal pr around this, and we can further debate if the idea makes sense.

danielhelfand:
Closing this issue. tkn <resource> create -f commands will be removed by version 1.0.0. The recommendation is to use kubectl for creating and updating of resources.

PR #582, opened 2020-01-08
tkn create Command Proposal

danielhelfand:
Associated with #574 

This pull request is a proposal for a `tkn create` command that invokes `kubectl create`. It would work as follows:

```
tkn create -f foo.yaml -n bar
```

This closely mirrors the approach implemented in [`ko delete`](https://github.com/google/ko/blob/1fcfe62a9606a2781559ed10897a4ae7141d7628/pkg/commands/delete.go). 

Pros of this approach:
* Less tool switching for users (Users will have to use `kubectl` less frequently when working with Tekton)
* Ability to create non Tekton resources that are used with Tekton

Cons:
* Requires `kubectl` to be installed/dependence on `kubectl` (could possibly swap out kubectl commands with this)
* We will most likely need to rely on `e2e` testing to support this without unit tests

# Submitter Checklist

# Release Notes

```
Add tkn create command that invokes kubectl create
```


go.sum
```diff
@@ -414,6 +414,7 @@ k8s.io/apimachinery v0.0.0-20191004074956-01f8b7d1121a h1:lDydUqHrbL/1l5ZQrqD1RI
 k8s.io/apimachinery v0.0.0-20191004074956-01f8b7d1121a/go.mod h1:ccL7Eh7zubPUSh9A3USN90/OzHNSVN6zxzde07TDCL0=
 k8s.io/cli-runtime v0.0.0-20191004110054-fe9b9282443f h1:vJOrMsZe+RD884n+WQ5So2oOp7SajI0Op3oOBg64ZsY=
 k8s.io/cli-runtime v0.0.0-20191004110054-fe9b9282443f/go.mod h1:qWnH3/b8sp/l7EvlDh7ulDU3UWA4P4N1NFbEEP791tM=
+k8s.io/cli-runtime v0.17.0 h1:XEuStbJBHCQlEKFyTQmceDKEWOSYHZkcYWKp3SsQ9Hk=
```
chmouel: I don't think you wanted this?

pkg/cmd/create/create.go
```diff
@@ -0,0 +1,71 @@
...
+			UnknownFlags: true,
+		},
+	}
+
+	return c
+}
+
+// returns a runCmd that simply passes CLI arguments
+// through to a binary named command.
+func invokeKubectl(command string) runCmd {
```
chmouel: What do you think about adding a check to look out for the binary and let the user that she would need to install `kubectl` before executing this command?

go.sum
vdemeester: @chmouel it's on the `sum` file, doesn't mean we get the dependency :stuck_out_tongue:

pkg/cmd/create/create.go
```diff
@@ -0,0 +1,71 @@
...
+
+	c := &cobra.Command{
+		Use:          "create",
+		Short:        "Uses kubectl create to create Kubernetes resources",
+		Example:      eg,
+		SilenceUsage: true,
+		Annotations: map[string]string{
+			"commandType": "main",
+		},
+		Run: invokeKubectl("kubectl"),
```
vdemeester: shouldn't it be `invokeCommand` then ? :stuck_out_tongue_closed_eyes: (as with the function itself doesn't have anything to do with `kubectl`)

danielhelfand: Yeah, that's a good call. I'll touch it up.

pkg/cmd/create/create.go
danielhelfand: Agreed. Let me add a check for `kubectl` binary.

danielhelfand: Check added via `isKubectlAvailable()`

chmouel:
/lgtm
/approve

Thanks 👍🏼🔥

PR #577, opened 2020-01-07
Update all delete commands to support multiple arguments

sbwsg:
# Changes

Fixes https://github.com/tektoncd/cli/issues/572

This commit allows the user to delete multiple resources of the same kind at once. For example, to delete multiple tasks the user can issue the command `tkn task delete foo bar baz -n mynamespace` and all three tasks, foo, bar, and baz, will be deleted.

The user is only asked to confirm the multiple deletions a single time. Each deletion is handled individually, meaning that it's possible for some deletions to fail while the rest go on to succeed. However, if any deletions do fail then the command will end as an error.

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [x] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Regenerate the manpages and docs with `make docs` and `make man` if needed.
- [x] Run the code checkers with `make check`
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

# Release Notes

```
Users can now delete multiple resources at once by providing more than one name to delete commands
```

PR #585, opened 2020-01-09
Refactor delete command code to be a bit more DRY

sbwsg:
# Changes

While implementing [multiple delete](https://github.com/tektoncd/cli/pull/577) I noticed that each delete command was implementing its own error handling and messaging even though each was doing exactly the same thing. This made working on the code more error-prone - I had to double check that I'd added each of my changes correctly across every repetition of the delete behaviour.

So this commit introduces a helper type, `Deleter`, that does most of the repetitive work and is now reused across all the delete commands.

# Submitter Checklist

### DIFF
--- a/pkg/cmd/pipelinerun/delete.go
+++ b/pkg/cmd/pipelinerun/delete.go
@@ -19,10 +19,9 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/tektoncd/cli/pkg/cli"
-	"github.com/tektoncd/cli/pkg/helper/names"
+	"github.com/tektoncd/cli/pkg/helper/deleter"
 	"github.com/tektoncd/cli/pkg/helper/options"
 	validate "github.com/tektoncd/cli/pkg/helper/validate"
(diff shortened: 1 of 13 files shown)


sbwsg:
> why the struct ?

Mostly because the function signature would be a bit gnarly when deleting both a resource and its relations I think.

pkg/cmd/pipelinerun/delete.go
```diff
@@ -78,22 +77,8 @@ func deletePipelineRuns(s *cli.Stream, p cli.Params, prNames []string) error {
...
-		}
-		success = append(success, prName)
-	}
-
-	if len(success) > 0 {
-		fmt.Fprintf(s.Out, "PipelineRuns deleted: %s\n", names.QuotedList(success))
-	}
-
-	return multierr.Combine(errs...)
+	d := deleter.New("PipelineRun", func(pipelineRunName string) error {
```
danielhelfand: Not that this should happen in this pr, but it may be a good idea to have TaskRuns be a related resource for pipelineruns.

PR #590, opened 2020-01-10
Add coloured status condition

chmouel:
Simple colours of status.Condition, if there is a unknown status it would fall
back to normal color.

It would respect `-C/--no-colour` if we don't want color

Fixed at the same time a small bug when status == Reason it would show
status(status) so twice, now it would only show it when status != Reason

How it would look like : 

image

(may add more colours or emojis in the future. 😛 )

# Changes

# Submitter Checklist

review them:

- [👍] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)

# Release Notes

```
conditions (ie: running/stopped or other) are now coloured..
```

### DIFF
--- a/pkg/formatted/k8s_test.go
+++ b/pkg/formatted/k8s_test.go
@@ -1,0 +1,15 @@
+package formatted
+
+import (
+	"testing"
+
+	"github.com/fatih/color"
+)
+
+func TestColor(t *testing.T) {
+	greenSuccess := color.New(color.FgHiGreen).Sprint("Succeeded")
+	cs := colorStatus("Succeeded")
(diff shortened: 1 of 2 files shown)


pkg/formatted/k8s_test.go
```diff
@@ -0,0 +1,16 @@
+package formatted
+
+import (
+	"testing"
+
+	"github.com/fatih/color"
+	_ "github.com/tektoncd/plumbing/scripts"
```
vdemeester: This import shouldn't be there :stuck_out_tongue_closed_eyes:

PR #591, opened 2020-01-10
Sort Steps for tkn taskrun desc by StartedAt

### LINKED ISSUE #460: Sort Steps by Order of Execution for tkn tr desc
Following up on #450.

The steps shown by `tkn taskrun describe` don't appear in the order that they are to be executed. Specifically, steps not explicitly defined as part of the task definition that show up as part of a task run make it confusing as to the order of what is to be executed. 

An example of a step not explicitly defined as part of a task but appears as part of a task run is [`s2i-nodejs`](https://github.com/openshift/pipelines-catalog/blob/0157bb7c7161c55fecaa6835137d01c61a360ec8/s2i-nodejs/s2i-nodejs-task.yaml) where an `image-digest-exporter` step appears that is not part of the task definition.

# Expected Behavior

Steps appear in order they are to be executed with `tkn tr desc`.

# Actual Behavior

Steps explicitly defined appear in correct order, but steps not part of task definition appear out of order.

1. Use the `s2i-nodejs` task to create a task run
2. Use what's currently in master for tkn and run `./tkn tr desc <tr_name>`
3. View the `Steps` portion of the output from `tkn tr desc`

danielhelfand:
Closes #460 

This pull request sorts steps that are part of a TaskRun by start time. The start time is from a list of StepStates part of the TaskRun. Each StepState has a Terminated/Running container state from which the `StartedAt` field can be pulled. If the container state is in a waiting state (i.e. PodInitializing), the order will be determined by how steps are ordered for the Task.

# Submitter Checklist

# Release Notes

```
Sort steps by start time for tkn tr desc
```

### DIFF
--- a/pkg/cmd/taskrun/describe.go
+++ b/pkg/cmd/taskrun/describe.go
@@ -16,6 +16,7 @@ package taskrun
 
 import (
 	"fmt"
+	"sort"
 	"text/tabwriter"
 	"text/template"
 
@@ -95,13 +96,13 @@ No params
 {{- end }}
 {{- end }}
 
-{{decorate "steps" ""}}{{decorate "underline bold" "Steps\n"}}
-
-{{- $l := len .TaskRun.Status.Steps }}{{ if eq $l 0 }}
+{{decorate "steps" ""}}{{decorate "underline bold" "Steps"}}
+{{$sortedSteps := sortStepStates .TaskRun.Status.Steps }}
+{{- $l := len $sortedSteps }}{{ if eq $l 0 }}
 No steps
 {{- else }}
  NAME	STATUS
-{{- range $step := .TaskRun.Status.Steps }}
+{{- range $step := $sortedSteps }}
 {{- $reason := stepReasonExists $step }}
  {{decorate "bullet" $step.Name }}	{{ $reason }}
 {{- end }}
@@ -177,6 +178,7 @@ func printTaskRunDescription(s *cli.Stream, trName string, p cli.Params) error {
 		"taskResourceRefExists": validate.TaskResourceRefExists,
 		"stepReasonExists":      validate.StepReasonExists,
 		"decorate":              formatted.DecorateAttr,
+		"sortStepStates":        sortStepStatesByStartTime,
 	}
 
 	w := tabwriter.NewWriter(s.Out, 0, 5, 3, ' ', tabwriter.TabIndent)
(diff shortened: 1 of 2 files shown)


pkg/cmd/taskrun/describe.go
```diff
@@ -194,3 +198,18 @@ func hasFailed(tr *v1alpha1.TaskRun) string {
 
 	return ""
 }
+
+func sortStepStatesByStartTime(steps []v1alpha1.StepState) []v1alpha1.StepState {
+	sort.Slice(steps, func(i, j int) bool {
+		if steps[j].Terminated == nil {
```
danielhelfand: /hold

I'm going to look into using `Running`'s `StartedAt` so `Running` steps can also be sorted

danielhelfand: Resolved

/hold cancel

danielhelfand:
/hold

I would like to do some research on what exactly leads to the ordering taking place and why the task ordering of the steps is not enough to ensure proper ordering.

go.sum
```diff
@@ -426,6 +426,7 @@ knative.dev/caching v0.0.0-20190719140829-2032732871ff/go.mod h1:dHXFU6CGlLlbzaW
 knative.dev/pkg v0.0.0-20190909195211-528ad1c1dd62 h1:0neKnC9hstF5Lewj6OaAJdXqk22IZ7pnOPQ8Zl8cfq4=
 knative.dev/pkg v0.0.0-20190909195211-528ad1c1dd62/go.mod h1:pgODObA1dTyhNoFxPZTTjNWfx6F0aKsKzn+vaT9XO/Q=
 knative.dev/pkg v0.0.0-20200108203553-8c6241207427 h1:swUPz/GhGE0bdXX8Ys6AXhjshbCcYXgW9weH7LvK28U=
+knative.dev/pkg v0.0.0-20200109235555-79d67498c2c4 h1:J8TW2j/+ESdQkL3aH2vF+FYbfLrEes58a0MeMxW7E+E=
```
chmouel: Is that what you wanted to have?

pkg/cmd/taskrun/describe.go
```diff
@@ -194,3 +198,45 @@ func hasFailed(tr *v1alpha1.TaskRun) string {
 
 	return ""
 }
+
+func sortStepStatesByStartTime(steps []v1alpha1.StepState) []v1alpha1.StepState {
```
chmouel: I think this function is a good candidate for a small and contained unit test 👼

chmouel:
> I would like to do some research on what exactly leads to the ordering taking place and why the task ordering of the steps is not enough to ensure proper ordering.

I think the ordering is hard on tekton because all steps are started waiting with entrypoint for others, and we don't know which one wait for which

danielhelfand:
/hold cancel

Think this is ready for another review. I have done done some research with the original task that raised this issue: [`s2i-nodejs`](https://github.com/openshift/pipelines-catalog/tree/master/s2i-nodejs). Sorting the steps by start time does put them in the correct order. You can use the task and reproduce the issue in #460. The changes here resolve it.

PR #607, opened 2020-01-15
Fix release tarball version to remove the v at first

### LINKED ISSUE #604: Tekton CLI version prints differently via brew vs downloaded
Tekton CLI version prints differently when installed via brew vs downloaded from GitHub:
```
$ /tmp/tkn_0.6.0_Darwin_x86_64/tkn version
Client version: 0.6.0
$ tkn version
Client version: v0.6.0
```

This causes issues for the Tekton VS Code extension

chmouel:
Fix so we don't have a v at the begining of the version number, so we have the
same version as the other released version.

Should make it for the next release.

Closes #604

### DIFF
--- a/tekton/release.sh
+++ b/tekton/release.sh
@@ -101,7 +101,7 @@ for c in ${COMMITS};do
 done
 
 # Add our VERSION so Makefile will pick it up when compiling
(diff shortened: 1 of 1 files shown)


chmouel:
/retest

😿

PR #608, opened 2020-01-15
Add flag to task and pipeline to only delete related resources

### LINKED ISSUE #589: Feature Request: Only Delete PipelineRuns or TaskRuns of a Pipeline or Task
As of `v0.6.0`, the CLI supports the ability to delete Pipelines/Tasks along with associated PipelineRuns/TaskRuns through the `--all` flag:

```
-a, --all                           Whether to delete related resources (pipelineruns) (default: false)

-a, --all                           Whether to delete related resources (taskruns) (default: false)
```

There should be an option to delete both PipelineRuns/TaskRuns associated with a Pipeline or Task but not the Pipeline/Task.

Results in the Pipeline named `pipelinename` still existing and all pipelineruns associated with that pipeline being deleted. 
```
tkn pipeline delete pipelinename --delete-pipelineruns
```

sbwsg:
# Changes

Fixes #589 

This commit adds a flag to delete resources related to a parent - a
taskrun's parent is a task, a pipelinerun's parent is a pipeline -
without deleting the task or pipeline.
    
The flags are:
`tkn taskrun delete --task task-name`
`tkn pipelinerun delete --pipeline pipeline-name`

TODO:

- [ ] Add unit tests for the rewritten flags --task and --pipeline

# Submitter Checklist

review them:

# Release Notes

```
tkn delete pipeline --delete-pipelineruns flag has been added to allow deletion of only the pipelineruns related to a pipeline (not the pipeline itself)
tkn delete task --delete-taskruns flag has been added to allow deletion of only the taskruns related to a task (not the task itself)
```

### DIFF
--- a/docs/cmd/tkn_pipeline_delete.md
+++ b/docs/cmd/tkn_pipeline_delete.md
@@ -28,7 +28,7 @@ or
 ### Options
 
 ```
-  -a, --all                           Whether to delete related resources (pipelineruns) (default: false)
+  -a, --all                           Whether to also delete related resources (pipelineruns) (default: false)
       --allow-missing-template-keys   If true, ignore any errors in templates when a field or map key is missing in the template. Only applies to golang and jsonpath output formats. (default true)
   -f, --force                         Whether to force deletion (default: false)
   -h, --help                          help for delete
--- a/docs/cmd/tkn_pipelinerun_delete.md
+++ b/docs/cmd/tkn_pipelinerun_delete.md
@@ -32,6 +32,7 @@ or
   -f, --force                         Whether to force deletion (default: false)
   -h, --help                          help for delete
   -o, --output string                 Output format. One of: json|yaml|name|go-template|go-template-file|template|templatefile|jsonpath|jsonpath-file.
+  -p, --pipeline string               The name of a pipeline whose pipelineruns should be deleted (does not delete the pipeline)
       --template string               Template string or path to template file to use when -o=go-template, -o=go-template-file. The template format is golang templates [http://golang.org/pkg/text/template/#pkg-overview].
 ```
 
--- a/docs/cmd/tkn_task_delete.md
+++ b/docs/cmd/tkn_task_delete.md
@@ -28,7 +28,7 @@ or
 ### Options
 
 ```
-  -a, --all                           Whether to delete related resources (taskruns) (default: false)
+  -a, --all                           Whether to also delete related resources (taskruns) (default: false)
       --allow-missing-template-keys   If true, ignore any errors in templates when a field or map key is missing in the template. Only applies to golang and jsonpath output formats. (default true)
   -f, --force                         Whether to force deletion (default: false)
   -h, --help                          help for delete
--- a/docs/cmd/tkn_taskrun_delete.md
+++ b/docs/cmd/tkn_taskrun_delete.md
@@ -32,6 +32,7 @@ or
   -f, --force                         Whether to force deletion (default: false)
   -h, --help                          help for delete
   -o, --output string                 Output format. One of: json|yaml|name|go-template|go-template-file|template|templatefile|jsonpath|jsonpath-file.
+  -t, --task string                   The name of a task whose taskruns should be deleted (does not delete the task)
       --template string               Template string or path to template file to use when -o=go-template, -o=go-template-file. The template format is golang templates [http://golang.org/pkg/text/template/#pkg-overview].
 ```
 
--- a/docs/man/man1/tkn-pipeline-delete.1
+++ b/docs/man/man1/tkn-pipeline-delete.1
@@ -21,7 +21,7 @@ Delete pipelines in a namespace
 .SH OPTIONS
 .PP
 \fB\-a\fP, \fB\-\-all\fP[=false]
-    Whether to delete related resources (pipelineruns) (default: false)
+    Whether to also delete related resources (pipelineruns) (default: false)
 
 .PP
 \fB\-\-allow\-missing\-template\-keys\fP[=true]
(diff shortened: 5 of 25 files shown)


sbwsg:
I wonder if the flags should be `--only-pipelineruns` and `--only-taskruns` instead of `--delete-pipelineruns` and `--delete-taskruns`.  Anyone got an opinion on that?

chmouel:
`tkn pipeline delete --only-runs` ? but that's may be too cryptic, 

I rather have the --only-* prefix which makes more sense imho

danielhelfand:
> `tkn pipeline delete --only-runs` ? but that may be too cryptic,
> 
> I rather have the --only-* prefix which makes more sense imho

`--only` is more clear to me that you aren't deleting anything else. I do prefer `--only-pipelineruns`/`--only-taskruns` to be more explicit.

sbwsg:
ok great, I've gone with `--only-taskruns` and `--only-pipelineruns`, thanks folks

vdemeester:
So… I think we need to think about this a little bit more :thinking: We have something a little bit like that when listing `TaskRun` and `PipelineRun`.

```
# List all taskruns of task 'foo' in namespace 'bar'
tkn taskrun list foo -n bar
```

This is *in* `taskrun` subcommand. You can filter the `TaskRun` of a given `Task`. 
- To be consistent, I would think we could have this on `delete` too, as `tkn taskrun delete --task=foo` or something ? 
- To be even more consitent, `tkn taskrun list foo` should be deprecated (but still supported for a while), and use a flag instead (the same as for *delete*, so `tkn taskrun list --task=foo`) ?

This would also allow to delete all `TaskRun` for multiple `Task`s, with `tkn taskrun delete --task=foo --task=bar`.

@chmouel @piyush-garg @danielhelfand @sbwsg wdyt ?

chmouel:
> To be consistent, I would think we could have this on delete too, as tkn taskrun delete --task=foo or something ?

IMHO I think that's less than intuitive... I like how @sbwsg  implemented it, where I can tell to myself : 

"I want to delete all  the taskruns that has been auto created by a task"

and map it into a command that looks as what i am thinking  : 

`tkn delete task --delete-taskruns`

your suggested command : 

`tkn taskrun delete --task=foo`

looks backward thinking to me.. (and we loose completion)

> To be even more consitent, tkn taskrun list foo should be deprecated (but still supported for a while), and use a flag instead (the same as for delete, so tkn taskrun list --task=foo) ?

👎  on this, even tho this may look more logical, this is not going to be intuitive for our end users

vdemeester:
:thinking: so `tkn taskrun list foo` and `tkn task delete --delete-taskruns` are intuitive but `tkn taskrun delete --task=foo` is not ? 

It might be just me, but as a user, if I want to delete all `TaskRun`s from a `Task` I would naively go to `tkn taskrun` subcommands than `tkn task` subcommands (aka I would look for a *filtering* option on the `TaskRun` delete command). This is what would feel intuitive for me :stuck_out_tongue:

chmouel:
Because you are deep into the project and you know what a run is 🤣 

We don't necessary expose the *Runs with CLI, we just say here is a task/pipeline and start it(tkn task start)....

 we give a taskrun objects so we can do stuff on it... but it really start from the `task` for the user...

vdemeester:
> Because you are deep neck into the project and you know what a run is rofl
> 
> We don't necessary expose the *Runs with CLI, we just say here is a task/pipeline and start it(tkn task start)....
> 
> we give a taskrun objects so we can do stuff on it... but it really start from the `task` for the user...

right, that might make sense :angel: but then should the flag be just about `run`s or even a subtask :smiling_imp:  ? (`tkn task delete runs`)

chmouel:
> ? (tkn task delete runs)

I am okay with that! but i'll let others weight in!

sbwsg:
I gotta be honest I don't feel strongly either way - do we want both?

`tkn taskrun delete --task my-task`
`tkn task delete my-task --only-taskruns`

My eye likes the syntax of the first one better.  But I agree that users are task-centric, not so much taskrun-centric.

@danielhelfand do you have any thoughts here from when you initially created the issue?

sbwsg:
Maybe we get even more fancy:

`tkn taskrun delete --filter "task=my-task"`?

chmouel:
Good idea @sbwsg  !

I am +1 on the filter thing... and we can add it to `tkn * list` too... (ie: #617) 

I think people are already used to the label selection from kubectl so this may transpose well..

vdemeester:
As @chmouel, I like the fancy solution :stuck_out_tongue:

siamaksade:
Filtering with explicit task and pipelinerun flags IMO is more intuitive. The key=value syntax incorrectly implies that it's similar to k8s labels:
`tkn taskrun delete --task tname`
`tkn taskrun delete --pipelinerun prname`

Same would go for the list:
`tkn taskrun list --task tname`
`tkn taskrun list --pipelinerun prname`

Could make sense to support the same syntax for the pipelinerun too in addition to the existing syntax:
`tkn pipelinerun list pname`
`tkn pipelinerun list --pipeline pname`

I'm opposed to indirect `delete` commands like:
`tkn delete task tname --delete-taskruns`

The command reads _delete tasks_ except that it doesn't do that, it _deletes taskruns_.

sbwsg:
Great feedback, thanks @siamaksade, you make some excellent points.  I think my next step here is to bring this up at the next CLI WG.

sbwsg:
OK cool, after quick chat in the working group, we're going to go with `tkn taskrun delete --task tname` style syntax

danielhelfand:
Failures are `-- FATAL: The documentation or manpages didn't seem to be generated :`

Just need to update docs/manpages.

vdemeester:
It is green :tada: 
/meow

chmouel:
/lgtm

🔥

sbwsg:
:pray: woo, cheers!

PR #609, opened 2020-01-15
Start doing testing against pipeline release nightly

### LINKED ISSUE #496: Start doing testing against nightly
We are currently testing only against the stable version of pipeline.

When there is breaking change in pipeline, we rush it to get it fixed like
mad(wo)men to fix the CI.

Let's do two different jobs, one against nightly and one against stable version
of pipeline.

/area testing
/kind feature

/cc @vdemeester @hrishin @sthaha @danielhelfand

chmouel:
Let's use nightly pipeline release.yaml, if we could not find it (i.e: nightly
pipeline as failed) let's fallback to the released version

Closes #496

### DIFF
--- a/test/e2e-common.sh
+++ b/test/e2e-common.sh
@@ -132,7 +132,14 @@ function install_pipeline_crd() {
   if [[ -n ${RELEASE_YAML} ]];then
 	latestreleaseyaml=${RELEASE_YAML}
   else
-    latestreleaseyaml=$(curl -s https://api.github.com/repos/tektoncd/pipeline/releases|python -c "import sys, json;x=json.load(sys.stdin);ass=x[0]['assets'];print([ x['browser_download_url'] for x in ass if x['name'] == 'release.yaml'][0])")
+    # First try to install latestreleaseyaml from nightly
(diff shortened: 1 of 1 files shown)


chmouel:
agreeed, but i have no idea how to do that :-)

vdemeester:
@chmouel no worries, we could do it using `prow` but I feel we can wait for `tektoncd/cli` CI using tekton :angel:

chmouel:
that works is only for midstream testing, we haven't done anything on upstream tekton-on-tekton  🏎 🍝 🐏 🌮 🏈 ⚽️

vdemeester:
> that works is only for midstream testing, we haven't done anything on upstream tekton-on-tekton , so that going to take quite quite some time stuck_out_tongue racing_car spaghetti ram taco football soccer

@chmouel What do you mean, we are starting to work on that too upstream :angel: :wink:

chmouel:
awesome, so then let's wait for that, i understand better how to update tekton tasks than prow jobs 🤣

chmouel:
Thinking about it, we could on the meantime rerun twice the tests, with different release (if we uninstall pipeline between runs)...

chmouel:
I think this is still mergeable!

PR #611, opened 2020-01-15
Generate steps name when none has been provided

### LINKED ISSUE #593: `tkn` assumes steps that has name
Steps may not have any names and the code is assuming we do, for example (just an example for taskrun but that's all over the describe commands) :

https://github.com/tektoncd/cli/blob/cf013bf243b5348b2762ac34166fb9f02807c68b/pkg/cmd/taskrun/describe.go#L100

When you don't have a name in steps you end up with a generated one like this : 

```
      name: step-unnamed-0
```

/kind bug

chmouel:
Let's generate the step name if the user has not provided one, like what is done
pipeline (which would shows up in tr)

Before:

image

After:

image

Closes #[593]

# Changes

# Submitter Checklist

review them:

for more details._

# Release Notes

```
release-note
```

### DIFF
--- a/pkg/cmd/task/describe.go
+++ b/pkg/cmd/task/describe.go
@@ -93,7 +93,7 @@ No params
  No steps
 {{- else }}
 {{- range $step := .Task.Spec.Steps }}
- {{decorate "bullet" $step.Name }}
+ {{ autoStepName $step.Name | decorate "bullet" }} 
 {{- end }}
 {{- end }}
 
@@ -193,6 +193,7 @@ func printTaskDescription(s *cli.Stream, p cli.Params, tname string) error {
 		"formatDuration":  formatted.Duration,
 		"formatCondition": formatted.Condition,
 		"decorate":        formatted.DecorateAttr,
+		"autoStepName":    formatted.AutoStepName,
 	}
 
 	w := tabwriter.NewWriter(s.Out, 0, 5, 3, ' ', tabwriter.TabIndent)
--- a/pkg/formatted/k8s.go
+++ b/pkg/formatted/k8s.go
@@ -15,6 +15,9 @@
 package formatted
 
 import (
+	"fmt"
+	"sync/atomic"
+
 	"github.com/fatih/color"
 	corev1 "k8s.io/api/core/v1"
 	"knative.dev/pkg/apis/duck/v1beta1"
@@ -29,9 +32,22 @@ var ConditionColor = map[string]color.Attribute{
 	"Pending":   color.FgHiWhite,
 }
 
+var stepCounter uint64
+
 // ColorStatus Get a status coloured
 func ColorStatus(status string) string {
 	return color.New(ConditionColor[status]).Sprint(status)
+}
+
+// AutoStepName when our stepName is empty return a generated name as generated
+// on pipeLine
+func AutoStepName(stepName string) string {
+	unnamedStep := fmt.Sprintf("unnamed-%d", stepCounter)
+	atomic.AddUint64(&stepCounter, 1)
+	if stepName != "" {
+		return stepName
+	}
+	return unnamedStep
 }
 
 // Condition returns a human readable text based on the status of the Condition
--- a/pkg/formatted/k8s_test.go
+++ b/pkg/formatted/k8s_test.go
@@ -3,6 +3,8 @@ package formatted
 import (
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+
 	"github.com/fatih/color"
 )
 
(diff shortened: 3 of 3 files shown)


pkg/formatted/k8s.go
```diff
@@ -29,11 +31,25 @@ var ConditionColor = map[string]color.Attribute{
 	"Pending":   color.FgHiWhite,
 }
 
+var stepCounter = -1
```
sthaha: should we start at 0?

pkg/formatted/k8s.go
```diff
@@ -29,11 +31,25 @@ var ConditionColor = map[string]color.Attribute{
...
 
+// AutoStepName when our stepName is empty return a genrated name as generated
+// on pipeLine
+func AutoStepName(stepName string) string {
+	if stepName != "" {
+		return stepName
+	}
+	// I don't think we should care about rwlock, gothread and stuff like this
+	// since it's only use in describe
+	stepCounter++
```
sthaha: agree although `atomic.AddUint64(&stepCounter, 1)` does the same without the need for locks

pkg/cmd/task/describe.go
```diff
@@ -93,7 +93,8 @@ No params
  No steps
 {{- else }}
 {{- range $step := .Task.Spec.Steps }}
- {{decorate "bullet" $step.Name }}
+{{- $stepName := autoStepname $step.Name}}
+ {{decorate "bullet" $stepName}}
```
sthaha: nit: IIUC, this can be written as `{{ autoStepname $stepName | decorate "bullet" }} `

pkg/formatted/k8s.go
chmouel: stepsname (as defined in pipeline) start a 0 so it makes the counter increase easier?

pkg/cmd/task/describe.go
chmouel: good call thanks, 

(just read properly the [gotemplate](https://golang.org/pkg/text/template/) documentation this morning)

pkg/formatted/k8s.go
chmouel: sound easy enough...

pkg/formatted/k8s.go
```diff
@@ -29,11 +32,24 @@ var ConditionColor = map[string]color.Attribute{
...
+
 // ColorStatus Get a status coloured
 func ColorStatus(status string) string {
 	return color.New(ConditionColor[status]).Sprint(status)
 }
 
+// AutoStepName when our stepName is empty return a genrated name as generated
+// on pipeLine
+func AutoStepName(stepName string) string {
+	if stepName != "" {
```
chmouel: I do this because I could not find a way to do this nicely in builtin gotemplate....

chmouel:
Sure! done...

/hold cancel

pkg/cmd/task/describe_test.go
```diff
@@ -238,6 +238,82 @@ Taskruns
...
+				tb.TaskSpec(
+					tb.Step("", "mandrake"),
+				),
+				tb.TaskSpec(
+					tb.Step("lamaracarena", "yggdrasil"),
+				),
+			),
+		},
+		TaskRuns: []*v1alpha1.TaskRun{
+			tb.TaskRun("tr-1", "ns",
```
vdemeester: Why do we create a `TaskRun` here, it's not referering our `Task`, so it's just noise in the test :stuck_out_tongue_closed_eyes:

pkg/cmd/task/describe_test.go
```diff
@@ -238,6 +238,82 @@ Taskruns
...
+
+Params
+
+ No params
+
+Steps
+
+ unnamed-0
+ alatouki
+ unnamed-1
```
vdemeester: This would be `unnamed-2` on the *backend*, that's why I wanted to add this test :angel:

pkg/cmd/task/describe_test.go
chmouel: whateverrr.. 🤷‍♂

pkg/cmd/task/describe_test.go
chmouel: Does it matter tho? It's generated name anyway?

vdemeester: Well, it does if the name of the step we display here is no the same of the one the controller will generate when executing the `TaskRun`

chmouel: if only there was no auto generated step names.....

pkg/formatted/k8s.go
```diff
@@ -29,11 +32,25 @@ var ConditionColor = map[string]color.Attribute{
...
+		atomic.AddUint64(&stepCounter, 1)
+		return stepName
+	}
+	unnamedStep := fmt.Sprintf("unnamed-%d", stepCounter)
+	atomic.AddUint64(&stepCounter, 1)
```
vdemeester: nit:

```go
unnamedStep := fmt.Sprintf("unnamed-%d", stepCounter)
atomic.AddUint64(&stepCounter, 1)
if stepName != "" {
	return stepName
}
return unnamedStep
```

pkg/formatted/k8s.go
sthaha: Just making sure .. so we want to print something like .. 

```
- A named step
- unnamed-2
- another-named-step
- unnamed-4
```
- right?

sthaha: Just saw that the test confirms it and it makes sense!

PR #618, opened 2020-01-16
Add tkn apply Command

### LINKED ISSUE #574: Have Create and Apply Not Be Tied to Specific Resources

danielhelfand:
Similar to #582. Closes #574.

This pull request is a proposal for a `tkn apply` command that invokes `kubectl apply`. It would work as follows:

```
tkn apply -f foo.yaml -n bar
```

This closely mirrors the approach implemented in [`ko delete`](https://github.com/google/ko/blob/1fcfe62a9606a2781559ed10897a4ae7141d7628/pkg/commands/delete.go). 

Pros of this approach:
* Less tool switching for users (Users will have to use `kubectl` less frequently when working with Tekton)
* Ability to create non Tekton resources that are used with Tekton

Cons:
* Requires `kubectl` to be installed/dependence on `kubectl` (could possibly swap out kubectl commands with this)
* We will most likely need to rely on `e2e` testing to support this without unit tests

# Submitter Checklist

# Release Notes

```
Add tkn apply command that invokes kubectl apply
```


chmouel:
I am starting to have second thoughts about those kubectl wrapped command, 

Perhaps this is going to be confusing for users?

If we start having all those wrapper command to k8s `tkn create` `tkn apply`, how would the user understand that apply create and others are targeted for k8/kubectl operation and not necessary to do Tekton operations?

chmouel:
cc @siamaksade

siamaksade:
@chmouel not really. I to have mixed feelings about it. `tkn` cannot hide away `kubectl` completely unless it mirrors all commands and that is not really what `tkn` is trying to achieve. Then the question would be what's the purpose of mirroring some commands?

Is this
```
kubectl create namespace demo
kubectl apply -f mypipeline.yaml
tkn pipeline start mypipeline 
```
that different from this?
```
kubectl create namespace demo
tkn apply -f mypipeline.yaml
tkn pipeline start mypipeline 
```

danielhelfand:
> Is this
> 
> ```
> kubectl create namespace demo
> kubectl apply -f mypipeline.yaml
> tkn pipeline start mypipeline 
> ```
> 
> that different from this?
> 
> ```
> kubectl create namespace demo
> tkn apply -f mypipeline.yaml
> tkn pipeline start mypipeline 
> ```

I do see the point and understand that `tkn` will in many ways need to be supplemented by `kubectl`. 

 Is it fair to say then that `tkn` should not support any `apply` or `create` commands? We have had requests to add this to `tkn`, but I suppose the main benefit would be to not have a `kubectl` dependency. The idea of actually redeveloping create/apply to do what `kubectl` already does, does not seem like the best use of anyone's time though. 

It certainly is convenient for users to not have to switch back and forth between tools, which would be the main benefit of this and could eventually be changed to actually not rely on `kubectl`. 

I am open to either approach (i.e. not supporting `create`/`apply` or moving forward with this approach). I only am going with this approach since it seemed to have support of people in the last working group for the CLI. 

However, I welcome any further discussion on it. What matters most to me is that we have what users will actually find useful. That's always a tough question though 😄 .

siamaksade:
@danielhelfand I do see the temptation and specially during demos I have had the same wish to show only `tkn` :) But the end for that road would be to mirror the entire `kubectl`, otherwise one way or other user has to switch back to it. If that's what the collective agreement is here, so be it although I'm not a big fan of it.

danielhelfand:
> @danielhelfand I do see the temptation and specially during demos I have had the same wish to show only `tkn` :) But the end for that road would be to mirror the entire `kubectl`, otherwise one way or other user has to switch back to it. If that's what the collective agreement is here, so be it although I'm not a big fan of it.

The more I do think about it, I am actually leaning more towards just not supporting create/apply via `tkn`. I think you make a lot of good points. Perhaps there are little things we can do like a `krew` plugin to help with this experience, but `kubectl` will always be needed.

Here's what I'll do:

Let's put this pr on hold and keep it open until after next week's working group. Let's discuss it with everyone, and, if others agree to not go with this approach, I'll close this and remove #582. 

/hold

PR #643, opened 2020-01-27
--dry-run Option for tkn pipeline start

danielhelfand:
Closes #372 

# Changes

This pull request adds a `--dry-run` option for `tkn pipeline start`. The idea would be to print the pipelinerun to stdout so a user can look over the pipelinerun without actually running the pipeline. 

An `--output` option will also be made available to output the dry run in yaml or json:

**Output the pipeline named pipelinename in namespace ns in yaml:**

`tkn pipeline start pipelinename -n ns --dry-run --output yaml`

**Output the pipeline named pipelinename in namespace ns in json:**
`tkn pipeline start pipelinename -n ns --dry-run -o json`

**Not using output flag with --dry-run uses yaml output default:**

`tkn pipeline start pipelinename -n ns --dry-run`

Example out:

**output = json**

```json
{
	"kind": "PipelineRun",
	"apiVersion": "tekton.dev/v1alpha1",
	"metadata": {
		"generateName": "deploy-pipeline-run-",
		"namespace": "default",
		"creationTimestamp": null
	},
	"spec": {
		"pipelineRef": {
			"name": "deploy-pipeline"
		},
		"resources": [
			{
				"name": "app-git",
				"resourceRef": {
					"name": "nodejs-ex-git"
				}
			},
			{
				"name": "app-image",
				"resourceRef": {
					"name": "nodejs-ex-image"
				}
			}
		],
		"serviceAccountName": "default",
		"podTemplate": {}
	},
	"status": {}
}
```

**output = yaml**

```yaml
kind: PipelineRun
metadata:
  creationTimestamp: null
  generateName: deploy-pipeline-run-
  namespace: default
spec:
  pipelineRef:
    name: deploy-pipeline
  podTemplate: {}
  resources:
  - name: app-git
    resourceRef:
      name: nodejs-ex-git
  - name: app-image
    resourceRef:
      name: nodejs-ex-image
  serviceAccountName: default
status: {}

```

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [x] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Regenerate the manpages and docs with `make docs` and `make man` if needed.
- [x] Run the code checkers with `make check`
[...]

PR #662, opened 2020-01-30
Add --dry-run Option for tkn task start

### LINKED ISSUE #654: Add --dry-run Option for tkn task start
Similar to #372 but for TaskRuns

# Expected Behavior

#643 is adding the ability to view a PipelineRun without creating the PipelineRun via a `--dry-run` option with `tkn pipeline start`. This same behavior should exist for `tkn task start` with viewing a TaskRun. 

This should be able to follow the design in #643 where `tkn task start <taskName> --dry-run`  produces yaml stdout of the TaskRun without actually creating the TaskRun.

There should also be an `--output` option that can be used with `--dry-run` to specify `json` or `yaml` output formats to display the TaskRun.

danielhelfand:
Closes #654 

Similar to #643, this pull request adds the ability to do a dry run of a taskrun with `tkn task start`. A notable issue with this, is that steps won't be included with this approach since it mainly relies on a `TaskRef` to start tasks, but it does show information on inputs, outputs, and params as well as the task reference being used. Using the `-f` option does provide information about steps since it does not require a TaskRef.

This pull request also makes a small change for the `--timeout` option, which is currently not available with `-f` due to only defining it when a task name argument is provided. 

The last change in this pr is getting rid of the `-o` shorthand for output with `--dry-run` for `tkn pipeline start --dry-run`. `-o` is not available with `tkn task start` because of outputs so I think it makes sense to keep it consistent across all commands. Only specifying `--output` works now. 

# Submitter Checklist

# Release Notes

```
Add --dry-run option for tkn task start
```


pkg/cmd/task/start.go
```diff
@@ -186,6 +194,7 @@ func startTask(opt startOptions, args []string) error {
 			TaskSpec: &task.Spec,
 		}
 	}
+	tr.Spec.Timeout = &metav1.Duration{Duration: timeoutSeconds}
```
danielhelfand: Moving setting `Timeout` here so that `-f` can set timeout for TaskRun.

pkg/cmd/pipeline/start.go
```diff
@@ -153,7 +153,7 @@ like cat,foo,bar
 	c.Flags().BoolVarP(&opt.Last, "last", "L", false, "re-run the pipeline using last pipelinerun values")
 	c.Flags().StringSliceVarP(&opt.Labels, "labels", "l", []string{}, "pass labels as label=value.")
 	c.Flags().BoolVarP(&opt.DryRun, "dry-run", "", false, "preview pipelinerun without running it")
-	c.Flags().StringVarP(&opt.Output, "output", "o", "", "format of pipelinerun dry-run (yaml or json)")
+	c.Flags().StringVarP(&opt.Output, "output", "", "", "format of pipelinerun dry-run (yaml or json)")
```
danielhelfand: Removing to keep consistency across `--dry-run` commands.

pkg/cmd/task/start_test.go
```diff
@@ -898,3 +900,130 @@ func Test_parseRes(t *testing.T) {
 		})
 	}
 }
+
+func TestTaskStart_ExecuteCommand(t *testing.T) {
```
danielhelfand: We should eventually move as many tests in this block as can be moved.

chmouel:
/lgtm
/approve

Thanks 👍🏼🔥

Issue #666, opened 2020-01-31
Start Task from Last should use the Same generateName


danielhelfand:
Similar to #639 but for TaskRuns. We should follow the same approach in #644 for `tkn task start --last`. Currently we are just using the task name for `generateName`:

https://github.com/tektoncd/cli/blob/2e5e5416af131c500b1c6dbbc590f00c1090c4ba/pkg/cmd/task/start.go#L189

PR #670, opened 2020-02-03
Fix overriding GenerateName on taskrun

16yuki0702:
# Changes
This fixes https://github.com/tektoncd/cli/issues/666

when user specified last flag, script always override GenerateName.
So using GenerateName if last TaskRun has a GenerateName,
If not, setting a GenerateName based on last TaskRun Name.

# Submitter Checklist

review them:

for more details._

# Release Notes

```
release-note
```

### DIFF
--- a/pkg/cmd/task/start.go
+++ b/pkg/cmd/task/start.go
@@ -207,6 +207,11 @@ func startTask(opt startOptions, args []string) error {
 		if err != nil {
 			return err
 		}
+		if len(trLast.ObjectMeta.GenerateName) > 0 {
+			tr.ObjectMeta.GenerateName = trLast.ObjectMeta.GenerateName
+		} else {
+			tr.ObjectMeta.GenerateName = trLast.ObjectMeta.Name + "-"
+		}
 		tr.Spec.Inputs = trLast.Spec.Inputs
 		tr.Spec.Outputs = trLast.Spec.Outputs
 		tr.Spec.ServiceAccountName = trLast.Spec.ServiceAccountName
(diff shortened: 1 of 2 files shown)


chmouel:
@16yuki0702 would you like to add your login to this list? : https://github.com/tektoncd/community/blob/master/org/org.yaml i think that's what needed so you don't need a `ok-to-test` for all your PR all the time

16yuki0702:
@chmouel 
> would you like to add your login to this list? : https://github.com/tektoncd/community/blob/master/org/org.yaml i think that's what needed so you don't need a ok-to-test for all your PR all the time

Yes, I'd like to add to the lists thanks :smile:

pkg/cmd/task/start.go
```diff
@@ -207,6 +207,11 @@ func startTask(opt startOptions, args []string) error {
 		if err != nil {
 			return err
 		}
+		if len(trLast.ObjectMeta.GenerateName) > 0 {
+			tr.ObjectMeta.GenerateName = trLast.ObjectMeta.GenerateName
+		} else {
+			tr.ObjectMeta.GenerateName = trLast.ObjectMeta.Name + "-"
```
chmouel: does that mean we alway going to have a generate name with --last ?

16yuki0702: If user start a task without ```--last```, GenerateName will be set as based on user specifiled name, like this.
```
tkn t start tp
```
GenerateName wil be set as
```
tp-run-
```

This behavior is the same as before the change.
https://github.com/tektoncd/cli/blob/master/pkg/cmd/task/start.go#L198

chmouel: ah gotcha thanks for explaining it to me!

16yuki0702: Thank you for your review :smile:

Issue #634, opened 2020-01-22
Add --all to the delete commands


chmouel:
I think it could be nice to have a --all to the delete commands, the same way
kubectl has it. I can surely use kubectl for it but i guess we want it on tkn
commands..

/kind feature

danielhelfand:
Assuming something like below deletes all tasks in namespace `ns`:

`tkn task delete --all -n ns`

csantanapr:
I'm in desperate need 🙏 of this command `tkn taskrun delete --all -n ns` for taskruns

This will allow me to clean all the pods hundreds)   and avoid having to switch cli `kubectl delete taskrun --all -n ns`

danielhelfand:
@csantanapr Yeah, it’s a good use case for it. We’re releasing a new version this week that will unfortunately not have this, but we can include in the next release and shouldn’t be too difficult to put in place.

searsaw:
In the meantime, here's a small snippet I use to clean them all up.

`tkn pr ls | awk '{print $1}' | xargs -n 1 tkn pr delete -f`

chmouel:
@searsaw fyi i think you will get the headers if you do this (i.e: NAME)

you can use the kubernetes command too, here is somethng that works : 

```
tkn pr delete -f $(tkn pr ls -o jsonpath="{range .items[*]}{.metadata.name}{'\n'}{end}")
````

**Comment by bitsofinfo** [comment]
agreed, right now you have to each one by one

**Comment by danielhelfand** [comment]
@bitsofinfo You can delete more than one at a time, but it requires you to write out each resource you would like to delete as of `v0.7.1`. For example:

```
tkn pr delete pr1 pr2 pr3 -f
```

We recently added `--all` for TaskRun deletion, and I'll be opening a pr today for deleting PipelineRuns. This will be available in master soon.

PR #676, opened 2020-02-04
--all option for tkn taskrun delete

danielhelfand:
Part of #634 

This pull request adds an `--all` option to `tkn taskrun delete` that will delete all taskruns that exist in a namespace. It follows the pattern we use for all delete commands by prompting the user if he or she would like to delete the resource. There is also the `-f` option to force a delete without a prompt. 

**Delete all taskruns in namespace ns:**

```
tkn taskrun delete --all -n ns

Are you sure you want to delete all TaskRuns in namespace "ns" (y/n): 
```

There will be a slight issue implementing this `--all` approach for Pipeline/Task: the `--all` flag is already in use, but I think we should change those to `--related` as it focuses on deleting related pipelineruns/taskruns. 

# Changes

# Release Notes

```
--all option for tkn taskrun delete that deletes all taskruns in a namespace
```

### DIFF
--- a/pkg/cmd/taskrun/delete.go
+++ b/pkg/cmd/taskrun/delete.go
@@ -28,7 +28,7 @@ import (
 )
 
 func deleteCommand(p cli.Params) *cobra.Command {
-	opts := &options.DeleteOptions{Resource: "taskrun", ForceDelete: false, ParentResource: "task"}
+	opts := &options.DeleteOptions{Resource: "taskrun", ForceDelete: false, ParentResource: "task", DeleteAllNs: false}
 	f := cliopts.NewPrintFlags("delete")
 	eg := `Delete TaskRuns with names 'foo' and 'bar' in namespace 'quux':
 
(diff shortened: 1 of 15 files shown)


pkg/helper/options/delete.go
```diff
@@ -28,24 +28,31 @@ type DeleteOptions struct {
 	ParentResource     string
 	ParentResourceName string
 	ForceDelete        bool
-	DeleteAll          bool
+	DeleteRelated      bool
```
danielhelfand: I renamed this as it more focuses on deleting related resources.

pkg/helper/options/delete.go
```diff
@@ -28,24 +28,31 @@ type DeleteOptions struct {
...
+	DeleteAllNs        bool
```
danielhelfand: `DeleteAllNs` since there will be ClusterTask case where it's not focused on deleting all resources in a namespace.

pkg/cmd/taskrun/delete.go
```diff
@@ -60,41 +60,58 @@ or
...
 			return cs.Tekton.TektonV1alpha1().TaskRuns(p.Namespace()).Delete(taskRunName, &metav1.DeleteOptions{})
 		})
-		d.DeleteRelated(s, []string{parentTask})
+		d.DeleteRelated(s, []string{opts.ParentResourceName})
+	}
+	if !opts.DeleteAllNs {
+		d.PrintSuccesses(s)
+	} else {
+		if d.Errors() == nil {
+			fmt.Fprintf(s.Out, "All TaskRuns deleted in namespace %q\n", p.Namespace())
```
danielhelfand: Print message saying all taskruns were deleted instead of showing every taskrun name that was deleted. Only displays if no errors.

chmouel:
looks good to me 

/lgtm

PR #687, opened 2020-02-06
Makefile fixes

chmouel:
Some makefile fixes:

* Include `lint-yaml` target when doing `make lint`
* Add `generate` target that would includes all the generation targets (ie man
  docs/golden)
* Fix `make fmt`, since $() is evaluated in makefile, need to use backquote for subshell
  exansion

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ check: lint test
 test: test-unit ## run all tests
 
 .PHONY: lint
-lint: ## run linter(s)
+lint: lint-yaml ## run linter(s)
 	@echo "Linting..."
 	@golangci-lint run ./... --timeout 5m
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -80,13 +80,16 @@ man: bin/docs ## update manpages
 	@echo "Update generated manpages"
 	@./bin/docs --target=./docs/man/man1 --kind=man
 
+.PHONY: generated
+generated: test-unit-update-golden man docs fmt ## generate all files that needs to be generated
```
vdemeester: I am a little bit worried to put `test-unit-update-golden` in `generated` as this might update golden file when not intended by the user (and thus hide a unwanted change in the output)

chmouel: why would it be? this should not happen isnt it ? I mean the point is to have the same exact output unless a code change?

chmouel: @vdemeester ping!

vdemeester: lol, not sure :stuck_out_tongue:

PR #698, opened 2020-02-10
--all option for tkn condition delete

danielhelfand:
Part of #634 

`--all` option for `tkn condition delete` to delete all conditions in a namespace.

# Submitter Checklist

review them:

# Release Notes

```
--all option for tkn cond delete that deletes all conditions in a namespace
```

### DIFF
--- a/pkg/cmd/condition/delete.go
+++ b/pkg/cmd/condition/delete.go
@@ -27,7 +27,7 @@ import (
 )
 
 func deleteCommand(p cli.Params) *cobra.Command {
-	opts := &options.DeleteOptions{Resource: "condition", ForceDelete: false}
+	opts := &options.DeleteOptions{Resource: "condition", ForceDelete: false, DeleteAllNs: false}
 	f := cliopts.NewPrintFlags("delete")
 	eg := `Delete Conditions with names 'foo' and 'bar' in namespace 'quux':
 
@@ -43,7 +43,7 @@ or
 		Aliases:      []string{"rm"},
 		Short:        "Delete a condition in a namespace",
 		Example:      eg,
-		Args:         cobra.MinimumNArgs(1),
+		Args:         cobra.MinimumNArgs(0),
 		SilenceUsage: true,
 		Annotations: map[string]string{
 			"commandType": "main",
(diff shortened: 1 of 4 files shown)


pkg/cmd/condition/delete.go
```diff
@@ -63,24 +63,51 @@ or
...
+		}
+	}
 	d.Delete(s, condNames)
-	d.PrintSuccesses(s)
+
+	if !deleteAll {
+		d.PrintSuccesses(s)
+	} else if deleteAll {
+		if d.Errors() == nil {
+			fmt.Fprintf(s.Out, "All Conditions deleted in namespace %q\n", p.Namespace())
```
chmouel: `has been` ?

pkg/cmd/condition/delete.go
```diff
@@ -63,24 +63,51 @@ or
...
+		if err != nil {
+			return err
+		}
+	}
 	d.Delete(s, condNames)
-	d.PrintSuccesses(s)
+
+	if !deleteAll {
+		d.PrintSuccesses(s)
+	} else if deleteAll {
```
chmouel: i think the logic can be a bit more clearer, ie : 


if deleteall { 
....
} else {
}

danielhelfand: The linter complains if I do it with else. It says to be more explicit.

chmouel: that's so weird....... did you take a note of which checker from golangci-lint this is coming from?

pkg/cmd/condition/delete.go
danielhelfand: This follows the language used in every pr for `--all`. The language also is past tense and brief, so I would prefer to keep it unless you feel strongly.

pkg/cmd/condition/delete.go
danielhelfand: No, but we can try it and get the results in this pr if you like?

pkg/cmd/condition/delete.go
chmouel: np!

pkg/cmd/condition/delete.go
danielhelfand: Yeah, so it fails with the following:

```
I0210 18:05:31.332] ==== TESTING IF GOLINT HAS BEEN DONE ====
I0210 18:05:31.333] =========================================
I0210 18:07:53.240] pkg/cmd/condition/delete.go:94:9: elseif: can replace 'else {if cond {}}' with 'else if cond {}' (gocritic)
I0210 18:07:53.246] 	} else {
I0210 18:07:53.246] 	       ^
```

chmouel: fair enough, thanks for tracking this. I'll try to do a bit more diging when i have some time because this doesn't feel right.

chmouel:
/lgtm
/approve

Thanks 👍🏼🔥

PR #720, opened 2020-02-12
Add --keep to delete --all, to keep the last N pipelineruns

### LINKED ISSUE #718: Add --keep `delete --all` for pipelineruns and taskruns
Example :

```
% tkn pipelineruns delete --all --keep=5
```

Will delete all pipelinesruns except the last 5 ones.

I think we should keep this option for the auto generated resources, like
taskruns and pipelineruns.

This was a query comping from a message on slack

https://tektoncd.slack.com/archives/CJ62C1555/p1581478036338900

asking if there was a way to limit the number of pipelinerun. There is no
support for this in `pipeline` atm but we can easily do that in cli and could be
much more flexible than what we can implement in `pipeline`.

/kind feature

chmouel:
# Changes

Add a --keep to delete --all so we can keep N numbers. Users can set this up in
a cronjobs to clean old pipelineruns but keeping only the last ones.

Closes #718

# Submitter Checklist

review them:

for more details._

### DIFF
--- a/pkg/cmd/pipelinerun/delete.go
+++ b/pkg/cmd/pipelinerun/delete.go
@@ -22,6 +22,7 @@ import (
 	"github.com/tektoncd/cli/pkg/cli"
 	"github.com/tektoncd/cli/pkg/helper/deleter"
 	"github.com/tektoncd/cli/pkg/helper/options"
+	prhsort "github.com/tektoncd/cli/pkg/helper/pipelinerun/sort"
 	validate "github.com/tektoncd/cli/pkg/helper/validate"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	cliopts "k8s.io/cli-runtime/pkg/genericclioptions"
@@ -70,6 +71,7 @@ or
 	f.AddFlags(c)
 	c.Flags().BoolVarP(&opts.ForceDelete, "force", "f", false, "Whether to force deletion (default: false)")
 	c.Flags().StringVarP(&opts.ParentResourceName, "pipeline", "p", "", "The name of a pipeline whose pipelineruns should be deleted (does not delete the pipeline)")
+	c.Flags().IntVarP(&opts.Keep, "keep", "", 0, "Keep n least recent number of pipelineruns when deleting alls")
 	c.Flags().BoolVarP(&opts.DeleteAllNs, "all", "", false, "Delete all pipelineruns in a namespace (default: false)")
 	_ = c.MarkZshCompPositionalArgumentCustom(1, "__tkn_get_pipelinerun")
 	return c
@@ -86,7 +88,7 @@ func deletePipelineRuns(s *cli.Stream, p cli.Params, prNames []string, opts *opt
 		d = deleter.New("PipelineRun", func(pipelineRunName string) error {
 			return cs.Tekton.TektonV1alpha1().PipelineRuns(p.Namespace()).Delete(pipelineRunName, &metav1.DeleteOptions{})
 		})
-		prs, err := allPipelineRunNames(p, cs)
+		prs, err := allPipelineRunNames(p, cs, opts.Keep)
 		if err != nil {
 			return err
 		}
@@ -109,7 +111,11 @@ func deletePipelineRuns(s *cli.Stream, p cli.Params, prNames []string, opts *opt
 		d.PrintSuccesses(s)
 	} else if opts.DeleteAllNs {
 		if d.Errors() == nil {
-			fmt.Fprintf(s.Out, "All PipelineRuns deleted in namespace %q\n", p.Namespace())
+			if opts.Keep > 0 {
+				fmt.Fprintf(s.Out, "All but %d PipelineRuns deleted in namespace %q\n", opts.Keep, p.Namespace())
+			} else {
+				fmt.Fprintf(s.Out, "All PipelineRuns deleted in namespace %q\n", p.Namespace())
+			}
 		}
 	}
 	return d.Errors()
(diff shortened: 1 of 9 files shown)


pkg/cmd/pipelinerun/delete.go
```diff
@@ -70,6 +71,7 @@ or
 	f.AddFlags(c)
 	c.Flags().BoolVarP(&opts.ForceDelete, "force", "f", false, "Whether to force deletion (default: false)")
 	c.Flags().StringVarP(&opts.ParentResourceName, "pipeline", "p", "", "The name of a pipeline whose pipelineruns should be deleted (does not delete the pipeline)")
+	c.Flags().IntVarP(&opts.Keep, "keep", "", 0, "Keep N number of pipelineruns when deleting alls")
```
danielhelfand:
```suggestion
	c.Flags().IntVarP(&opts.Keep, "keep", "", 0, "keep n number of pipelineruns when using --all flag")
```

pkg/cmd/pipelinerun/delete.go
```diff
@@ -132,13 +138,18 @@ func pipelineRunLister(p cli.Params, cs *cli.Clients) func(string) ([]string, er
...
+func allPipelineRunNames(p cli.Params, cs *cli.Clients, keep int) ([]string, error) {
 	pipelineRuns, err := cs.Tekton.TektonV1alpha1().PipelineRuns(p.Namespace()).List(metav1.ListOptions{})
 	if err != nil {
 		return nil, err
 	}
 	var names []string
-	for _, pr := range pipelineRuns.Items {
+	var counter = 0
+	for _, pr := range prhsort.SortPipelineRunsByStartTime(pipelineRuns.Items) {
+		if keep > 0 && counter != keep {
```
danielhelfand: Should there be some kind of error message if keep < 0? Currently, it will delete all pipelineruns.

pkg/cmd/pipelinerun/delete.go
danielhelfand: Would be nice in some way to indicate that ordering here is done by most recent pipelinerun.

chmouel: humm usually i would ask to my english master  👨‍🏫  (ie you 😄 )

what about ?

Keep the `last` n number of pipelineruns?

pkg/cmd/pipelinerun/delete.go
chmouel: in case If the user choose ?: 

--last=-1 

I guess  that's quite a weird user case

danielhelfand: Right, I mean it's totally unlikely, but we should prevent it by checking if keep < 0 similar to what is done for limit. So fail fast in `RunE` in this case.

pkg/cmd/pipelinerun/delete.go
danielhelfand: `keep last n number of pipelineruns` sounds good to me.

danielhelfand:
/hold

There is some feedback that I think should be addressed in this pr still.

pkg/cmd/pipelinerun/delete.go
vdemeester: @danielhelfand isn't the code above already checking that `keep` can't be `< 0` ?

danielhelfand:
> A couple things:
> 
> 1. There needs to be a prompt specifically for `--keep` when not using `-f`. Currently it's what is shown below, but it should be a message about keeping a certain amount of pipelines:
> 
> `Are you sure you want to delete all pipelineruns in namespace "default" (y/n):`
> 
> 2. This is the error message for if `--keep` is used without `--all`:
> 
> `Error: must provide pipelineruns to delete or --pipeline flag`
> 
> There should be a clearer error message.

@vdemeester Less about the issue with `keep` < 0. More about the issues listed above.

vdemeester:
@danielhelfand ah indeed

vdemeester:
@danielhelfand taken into account :angel: 
/hold cancel

pkg/cmd/pipelinerun/delete.go
```diff
@@ -80,13 +82,16 @@ func deletePipelineRuns(s *cli.Stream, p cli.Params, prNames []string, opts *opt
 	if err != nil {
 		return fmt.Errorf("failed to create tekton client")
 	}
+	if opts.Keep > 0 && !opts.DeleteAllNs {
```
danielhelfand: This error message will only be reached if using `-f`. My recommendation would be to check earlier in `RunE` or `Args` to account for prompt situation:

If I run `tkn pr delete --keep 7`, I get the following:

```
Error: must provide pipelinerun name(s) or use --pipeline flag or --all flag to use delete
```

pkg/helper/options/delete.go
```diff
@@ -31,9 +31,13 @@ type DeleteOptions struct {
 	DeleteRelated      bool
 	DeleteAllNs        bool
 	DeleteAll          bool
+	Keep               int
 }
 
 func (o *DeleteOptions) CheckOptions(s *cli.Stream, resourceNames []string, ns string) error {
+	if o.Keep > 0 && !(o.DeleteAllNs || o.DeleteAll) {
+		return fmt.Errorf("must provide pipelineruns to delete or --pipeline flag")
```
danielhelfand: Promise this is the last thing. Could the error message be:

```suggestion
		return fmt.Errorf("must use --all flag with --keep")
```

danielhelfand: Promise this is the last thing. Could the error message be:

```suggestion
		return fmt.Errorf("must use --all flag with --keep")
```

danielhelfand: Promise this is the last thing. Could the error message be:

```suggestion
		return fmt.Errorf("must use --all flag with --keep")
```

pkg/helper/options/delete.go
```diff
@@ -31,9 +31,13 @@ type DeleteOptions struct {
 	DeleteRelated      bool
 	DeleteAllNs        bool
 	DeleteAll          bool
+	Keep               int
 }
 
 func (o *DeleteOptions) CheckOptions(s *cli.Stream, resourceNames []string, ns string) error {
+	if o.Keep > 0 && !(o.DeleteAllNs || o.DeleteAll) {
```
danielhelfand: Promise this is the last thing. Could the error message be:

```suggestion
		return fmt.Errorf("must use --all flag with --keep")
```

PR #691, opened 2020-02-07
Fix bug about Complete logs not shown for failed TaskRun

16yuki0702:
# Changes
Related issue is https://github.com/tektoncd/cli/issues/587

This bug seems to be happen from When pod status is failed ```hasTaskRunFailed``` method returns error immediately,
so pod's logs can't be shown up.

This fix stop returning error immediately, if pod status is failed just logging error on standard error.

Test yaml files is
```
apiVersion: tekton.dev/v1alpha1
kind: Task
metadata:
  name: tt
spec:
  steps:
    - name: echo
      image: ubuntu
      command:
        - echol
      args:
        - "hello world"
---
apiVersion: tekton.dev/v1alpha1
kind: TaskRun
metadata:
  name: diff-tt
spec:
  taskRef:
    name: tt
---
```

from this fix, logs can be shown like below.
```
$ tkn tr ls
NAME       STARTED          DURATION     STATUS               
diff-tt2   45 minutes ago   10 seconds   Succeeded   
diff-tt    53 minutes ago   9 seconds    Failed 
```

```
$ tkn tr logs diff-tt
task tt has failed: "step-echo" exited with code 1 (image: "docker-pullable://ubuntu@sha256:8d31dad0c58f552e890d68bbfb735588b6b820a46e459672d96e585871acc110"); for logs run: kubectl -n default logs diff-tt-pod-h2gwv -c step-echo
[echo] {"level":"info","ts":1581057041.4130957,"logger":"fallback-logger","caller":"logging/config.go:69","msg":"Fetch GitHub commit ID from kodata failed: \"KO_DATA_PATH\" does not exist or is empty"}
[echo] 2020/02/07 06:30:42 Error executing command: exec: "echol": executable file not found in $PATH

container step-echo has failed  : [{"name":"","digest":"","key":"StartedAt","value":"2020-02-07T06:30:42Z","resourceRef":{}}]
```

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [ ] Regenerate the manpages and docs with `make docs` and `make man` if needed.
- [ ] Run the code checkers with `make check`
- [ ] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

[...]

PR #721, opened 2020-02-12
nil Check for lr.Stream for TaskRun Logs

danielhelfand:
#691 introduced a way to get the logs of a failed TaskRun without using the `-f` flag, but it will panic if the pod fails while initializing. This pull request adds a nil check for `lr.Stream` to prevent panics in this case. 

# Submitter Checklist

# Release Notes

```
Add nil check to prevent panics with pipelinerun logs
```


chmouel:
/lgtm

Looking Good! 🤙🏽

Issue #748, opened 2020-02-21
Refactoring log reader


16yuki0702:
Related issue is https://github.com/tektoncd/cli/issues/708

```pkg/cmd/taskrun/log_reader.go``` and ```pkg/cmd/pipelinerun/log_reader.go``` also has a few same logic.
I think It would be useful and readable moving these code into same package.

16yuki0702:
Addressed by https://github.com/tektoncd/cli/pull/749 https://github.com/tektoncd/cli/pull/754

/close

PR #749, opened 2020-02-21
Refactoring log reader.

16yuki0702:
Related issue is https://github.com/tektoncd/cli/issues/748

taskrun's logreader and pipelinerun's logreader has a few same logic.
moving these code into same package.

and remove reference from ```pkg/helper/taskrun/taskrun.go``` to ```cmd``` package.

# Changes

# Submitter Checklist

review them:

for more details._

# Release Notes

```
release-note
```

### DIFF
--- a/pkg/helper/log/pipeline_reader.go
+++ b/pkg/helper/log/pipeline_reader.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package pipelinerun
+package log
 
 import (
 	"fmt"
@@ -20,11 +20,7 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/tektoncd/cli/pkg/cli"
-	"github.com/tektoncd/cli/pkg/cmd/taskrun"
-	"github.com/tektoncd/cli/pkg/helper/log"
 	"github.com/tektoncd/cli/pkg/helper/pipelinerun"
-	"github.com/tektoncd/cli/pkg/helper/pods/stream"
 	trh "github.com/tektoncd/cli/pkg/helper/taskrun"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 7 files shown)


pkg/helper/log/pipeline_reader.go
```diff
@@ -79,9 +63,10 @@ func (lr *LogReader) readLiveLogs(pr *v1alpha1.PipelineRun) (<-chan log.Log, <-c
 				go func(tr trh.Run, taskNum int32) {
 					defer wg.Done()
 
-					tlr := tr.NewLogReader(lr.Ns, lr.Clients, lr.Streamer,
-						int(taskNum), lr.Follow, lr.AllSteps)
-					pipeLogs(logC, errC, tlr)
+					// clone the object to keep task number and name separately
+					c := r.clone()
```
16yuki0702: By this the reference from ```pkg/helper/taskrun``` to ```pkg/cmd/taskrun``` package can be deleted.

pkg/helper/log/task_reader.go
```diff
@@ -41,87 +42,70 @@ func (s *step) hasStarted() bool {
...
-	//Check if taskrun failed on start up
-	if err := hasTaskRunFailed(tr.Status.Conditions, lr.Task); err != nil {
-		if lr.Stream != nil {
-			fmt.Fprintf(lr.Stream.Err, "%s\n", err.Error())
+	// Check if taskrun failed on start up
+	if err := hasTaskRunFailed(tr.Status.Conditions, r.task); err != nil {
+		if r.stream != nil {
+			fmt.Fprintf(r.stream.Err, "%s\n", err.Error())
 		}
 		return nil, nil, err
```
16yuki0702: Isn't ```failed``` pod log shown again without ```-f``` flag 🤔 ?
(because return at this point before readStepsLogs)

following code is correct 🤔 ?
```
if r.stream != nil {
	fmt.Fprintf(r.stream.Err, "%s\n", err.Error())
} else {
	return nil, nil, err
}
```

danielhelfand: Yes, that was something that was overlooked on my part in #721. I think adding the else block makes sense. 

Overall, we need to clean up error messaging in logs/have better testing around this. Maybe even an e2e scenario.

16yuki0702: > Yes, that was something that was overlooked on my part in #721. I think adding the else block makes sense.

Oh, Thanks !!
I'll add else block in this PR for now :thumbsup:

> Overall, we need to clean up error messaging in logs/have better testing around this. Maybe even an e2e scenario.

+1

pkg/helper/log/pipeline_reader.go
```diff
@@ -96,34 +81,33 @@ func (lr *LogReader) readLiveLogs(pr *v1alpha1.PipelineRun) (<-chan log.Log, <-c
...
-	pl, err := tkn.TektonV1alpha1().Pipelines(lr.Ns).Get(pr.Spec.PipelineRef.Name, metav1.GetOptions{})
+	tkn := r.clients.Tekton
+	pl, err := tkn.TektonV1alpha1().Pipelines(r.ns).Get(pr.Spec.PipelineRef.Name, metav1.GetOptions{})
 	if err != nil {
-		return nil, nil, fmt.Errorf(err.Error())
+		return nil, nil, err
 	}
 
-	//Sort taskruns, to display the taskrun logs as per pipeline tasks order
+	// Sort taskruns, to display the taskrun logs as per pipeline tasks order
```
danielhelfand: We need to do sorting for readLivePipelineLogs too.

16yuki0702: Because ```readLivePipelineLogs``` get logs asynchronous in concurrent goroutine.
Logs are displayed in order from the started pod and this timing is uncontrollable.
if we sort order in ```readLivePipelineLogs``` we need to wait until all task pods will start.
(But it maybe lack of real time... tradeoff...🤔)

16yuki0702: Ahh... sorry, plase let me considering it.
there is possibility I misunderstanding it, sorry :sweat_smile:

danielhelfand: I follow what you're saying, and that was my thought here too as far as having to wait for all TaskRuns. I figured I would ask to make sure but no worries if this is something out of the scope of this pr. We can always iterate on it if there isn't a clear solution here.

16yuki0702: > I follow what you're saying, and that was my thought here too as far as having to wait for all TaskRuns. I figured I would ask to make sure but no worries if this is something out of the scope of this pr. We can always iterate on it if there isn't a clear solution here.

Thank you very much !!
yes, it seems be better addressing this on another PR :pray:

Issue #680, opened 2020-02-05
TektonCD CLI official container image


chmouel:
We don't have our own as from the cli project own, plumbing is building this for us:

 https://github.com/tektoncd/plumbing/blob/master/tekton/images/tkn/Dockerfile

We currently use this image on the `tkn` catalog task :

https://gcr.io/tekton-releases/dogfooding/tkn

should we start owning our images to be released at release time? with some proper tags and such that we can integrate into our release_process[1]

cc @vdemeester @afrittoli 

[1] Although part of the release process document is to update the Dockerfile to latest released version.

vdemeester:
+:100:, we should have the Dockerfile in there, and have nightly and releases task building and publishing this image from here :angel:

piyush-garg:
@vdemeester I think we have done this, can i close this

PR #770, opened 2020-03-03
Add tkn official image on contrib 🖌

vdemeester:
# Changes

This adds a multi-stage `Dockerfile` to build official tkn image.

Part of #680  (needs a plumbing update to get closed though)

/cc @chmouel 

# Submitter Checklist

review them:

for more details._

# Release Notes

```
Add official images recipe
```

### DIFF
--- a/contrib/tkn-image/Dockerfile
+++ b/contrib/tkn-image/Dockerfile
@@ -1,0 +1,10 @@
+ARG GOLANG_VERSION=1.13.8
+ARG DEBIAN_VERSION=10
+
+FROM golang:${GOLANG_VERSION} as builder
+COPY . /go/src/github.com/tektoncd/cli
+WORKDIR /go/src/github.com/tektoncd/cli
+RUN make bin/tkn
+
+FROM debian:${DEBIAN_VERSION} as tkn
+COPY --from=builder /go/src/github.com/tektoncd/cli/bin/tkn /usr/bin
--- a/contrib/tkn-image/README.md
+++ b/contrib/tkn-image/README.md
@@ -1,0 +1,9 @@
+# tkn oci image
+
+This directory contains the `Dockerfile` used to build the official
+`tkn` image. It is a multi-stage `Dockerfile` that defines 1 targets:
+
+```
+docker build --file=contrib/tkn-image/Dockerfile \
+    --target=tkn --tag tkn:latest .
+```


chmouel:
Should we update the RELEASE_PROCESS documentation on how to build this on release?

contrib/tkn-image/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+FROM golang:${GOLANG_VERSION} as builder
+COPY . /go/src/github.com/tektoncd/cli
+WORKDIR /go/src/github.com/tektoncd/cli
+RUN make bin/tkn
+
+FROM alpine:${ALPINE_VERSION} as tkn-alpine
+COPY --from=builder /go/src/github.com/tektoncd/cli/bin/tkn /usr/bin
+
+FROM ubuntu:${UBUNTU_VERSION} as tkn-ubuntu
```
chmouel: do we really need multi-os images? why not just the alpine one?

vdemeester: Because some people may want to use something else than `alpine` (something with a `glibc` instead of `musl`, …). It doesn't cost too much to build and publish those too :angel: (and reduce the need for the user to do it's own.

Question then is : what would be the default os for `latest` and `0.8.0` for ex ?

chmouel: so just use `ubuntu` or `fedora` and that would be the default? It may not cost much but it adds confusion....

vdemeester: `tkn:latest` would be 0.8.0 alpine based
`tkn:0.8.0` would be 0.8.0 alpine based
`tkn:ubuntu` would be 0.8.0 ubuntu based (as of today)
`tkn:0.8.0-ubuntu` would be 0.8.0 ubuntu based
`tkn:fedora` would be would be 0.8.0 fedora base (as of today)
etc…

A bit like what `golang` does :wink:

vdemeester: (the target name is not the final tag name :wink:)

contrib/tkn-image/README.md
```diff
@@ -0,0 +1,20 @@
+# tkn oci image
+
+This directory contains the `Dockerfile` used to build the official
+`tkn` image. It is a multi-stage `Dockerfile` that defines 3 targets:
+
+- `tkn-alpine` for a alpine-based image
+  ```bash
+  docker build --file=contrib/tkn-image/Dockerfile \
+      --target=tkn-apline tkn:alpine .
```
searsaw: `alpine` is spelled wrong in the target. Also, I believe the there is a `--tag` flag missing before `tkn:alpine`. This tag is missing in the other two commands in this README as well.

vdemeester: omfg :man_facepalming: I think I was tired :joy_cat:

vdemeester: fixed :angel:

vdemeester:
Updated to only create a `debian` based image :wink:

chmouel:
/lgtm
/approve

Thanks, less images, less confusion, less support 👍🏼🔥

PR #773, opened 2020-03-03
Convert --all to Delete all Pipelines/Tasks in Namespace

### LINKED ISSUE #634: Add --all to the delete commands

danielhelfand:
Closes #634 

This pull request changes the behavior of `--all` to delete all pipelines/tasks in a namespace. It converts the name of the original `--all` flags to `--prs` and `--trs` (but I can easily change this to something else). The thought behind having the flag be named `--prs`/`--trs` is because it also deletes pipelineruns/taskruns associated with a pipeline/task that is also going to be deleted.

This pull request also removes the warning flag about the `--all` flag behavior/name changing in v0.9.0. 

# Submitter Checklist

# Release Notes

```
Convert --all to --prs and --trs for tkn pipeline/task delete so --all can be used to delete all pipelines/tasks in a namespace
```

### DIFF
--- a/pkg/cmd/pipeline/delete.go
+++ b/pkg/cmd/pipeline/delete.go
@@ -16,7 +16,6 @@ package pipeline
 
 import (
 	"fmt"
-	"log"
 
 	"github.com/spf13/cobra"
 	"github.com/tektoncd/cli/pkg/cli"
@@ -28,7 +27,7 @@ import (
 )
 
 func deleteCommand(p cli.Params) *cobra.Command {
-	opts := &options.DeleteOptions{Resource: "pipeline", ForceDelete: false, DeleteRelated: false}
+	opts := &options.DeleteOptions{Resource: "pipeline", ForceDelete: false, DeleteRelated: false, DeleteAllNs: false}
 	f := cliopts.NewPrintFlags("delete")
 	eg := `Delete Pipelines with names 'foo' and 'bar' in namespace 'quux'
 
@@ -44,7 +43,7 @@ or
 		Aliases:      []string{"rm"},
 		Short:        "Delete pipelines in a namespace",
 		Example:      eg,
-		Args:         cobra.MinimumNArgs(1),
+		Args:         cobra.MinimumNArgs(0),
 		SilenceUsage: true,
 		Annotations: map[string]string{
 			"commandType": "main",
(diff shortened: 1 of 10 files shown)


pkg/cmd/pipeline/delete.go
```diff
@@ -73,7 +68,8 @@ or
 	}
 	f.AddFlags(c)
 	c.Flags().BoolVarP(&opts.ForceDelete, "force", "f", false, "Whether to force deletion (default: false)")
-	c.Flags().BoolVarP(&opts.DeleteRelated, "all", "a", false, "Whether to also delete related resources (pipelineruns) (default: false)")
+	c.Flags().BoolVarP(&opts.DeleteRelated, "prs", "", false, "Whether to delete pipeline(s) and related resources (pipelineruns) (default: false)")
```
danielhelfand: If there is a more preferable name than this, please let me know your thoughts.

danielhelfand:
/hold 

Also going to add tasks as part of this

pkg/cmd/pipeline/delete.go
```diff
@@ -28,7 +27,7 @@ import (
 )
 
 func deleteCommand(p cli.Params) *cobra.Command {
-	opts := &options.DeleteOptions{Resource: "pipeline", ForceDelete: false, DeleteRelated: false}
+	opts := &options.DeleteOptions{Resource: "pipeline", ForceDelete: false, DeleteRelated: false, DeleteAllNs: false}
```
pradeepitm12: Nit: `DeleteAllNs` looks like delete all namespaces

danielhelfand: It’s currently used in every delete command across the CLI, so I would recommend opening a separate issue on this that would be handled separately than in this pr. 

I do see your point though on it being confusing.

danielhelfand: It’s currently used in every delete command across the CLI, so I would recommend opening a separate issue on this that would be handled separately than in this pr. 

I do see your point though on it being confusing.

csantanapr:
Woot !!!

Issue #784, opened 2020-03-06
Change ClusterTask Timeout from Int to String and Remove -t Shorthand


danielhelfand:
Similar to #779, `tkn clustertask start --timeout` should be changed to a string duration format (i.e. `1h30m2s`). Currently it is an [integer](https://github.com/tektoncd/cli/blob/b5d1f0c796344d92b6a63f206ca7aa93536d4f7a/pkg/cmd/clustertask/start.go#L54) that is used to [specify a timeout in seconds](https://github.com/tektoncd/cli/blob/b5d1f0c796344d92b6a63f206ca7aa93536d4f7a/pkg/cmd/clustertask/start.go#L133).

Similarly, the `-t` shorthand should be removed to keep consistency across all timeout commands.

Unfortunately, this was missed in our previous round of deprecation/behavior change notices around timeout, so this change cannot take place until `v1.0.0`. We should log a warning message to users about this change in version `v0.9.0`. It can be the exact same format as what was used in #753.

### v0.9.0 Changes

Log warning message to user whenever `--timeout` is used with `tkn ct start`:

```
WARNING: The --timeout flag will no longer be specified in seconds in v1.0.0. Learn more here: https://github.com/tektoncd/cli/issues/784
WARNING: The -t shortand for --timeout will no longer be available in v1.0.0.
```

### v1.0.0 Changes

The overall goal here will be to specify a Timeout via `tkn` as follows:

```
tkn clustertask start ctName --timeout 30m2s
```

danielhelfand:
@kadern0, I think you have to assign yourself to the issue.

kadern0:
Done!

O mér., 11 de mar. de 2020 ás 13:42, Daniel Helfand (<
notifications@github.com>) escribiu:

vdemeester:
/assign @kadern0 

(this works as soon as you commented once, I think)

PR #793, opened 2020-03-11
Change clusterTask timeout from int to string and remove -t shorthand

kadern0:
Signed-off-by: kadern0 <kaderno@gmail.com>

# Changes

Changed clusterTask timeout from int64 to string and removed the shorthand `-t` option.
# Submitter Checklist

review them:

for more details._

# Release Notes

### DIFF
--- a/pkg/cmd/clustertask/start.go
+++ b/pkg/cmd/clustertask/start.go
@@ -56,7 +56,7 @@ type startOptions struct {
 	Last               bool
 	Labels             []string
 	ShowLog            bool
-	TimeOut            int64
+	TimeOut            string
 	DryRun             bool
 	Output             string
 }
@@ -144,7 +144,7 @@ like cat,foo,bar
 	c.Flags().BoolVarP(&opt.Last, "last", "L", false, "re-run the clustertask using last taskrun values")
 	c.Flags().StringSliceVarP(&opt.Labels, "labels", "l", []string{}, "pass labels as label=value.")
 	c.Flags().BoolVarP(&opt.ShowLog, "showlog", "", false, "show logs right after starting the clustertask")
-	c.Flags().Int64VarP(&opt.TimeOut, "timeout", "t", 3600, "timeout for taskrun in seconds")
+	c.Flags().StringVar(&opt.TimeOut, "timeout", "1h", "timeout for taskrun")
 	c.Flags().BoolVarP(&opt.DryRun, "dry-run", "", false, "preview taskrun without running it")
 	c.Flags().StringVarP(&opt.Output, "output", "", "", "format of taskrun dry-run (yaml or json)")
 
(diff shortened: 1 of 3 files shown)


danielhelfand:
/hold

Thanks for this pr @kadern0. Unfortunately, this can't be added until version `v0.10.0` as documented in #784. We would be happy to accept this, but it will have to wait until after the next release of `tkn`. 

If you are interested in opening a separate pr, we need to add a warning message about this change to users in `v0.9.0` that can be added into this upcoming release. The format of the message is detailed in the issue.

kadern0:
@danielhelfand what do you suggest? Can this wait until version `v01.0.0` ?

danielhelfand:
> @danielhelfand what do you suggest? Can this wait until version `v01.0.0` ?

I will assign you to the issue if that works for you so that it's clear to others it's being worked on. We can keep this open/give any feedback until `v1.0.0` and get this in to that release. Sound good?

kadern0:
Sounds good to me. Thanks!

pkg/cmd/clustertask/start.go
```diff
@@ -149,7 +149,7 @@ func startClusterTask(opt startOptions, args []string) error {
 	}
 
 	var ctname string
-	timeoutSeconds := time.Duration(opt.TimeOut) * time.Second
+	timeoutSeconds, _ := time.ParseDuration(opt.TimeOut)
```
danielhelfand: You'll want to check if an error happens during duration here. I would also recommend renaming the variable since the timeout is no longer specified in seconds:

```go
timeout, err := time.ParseDuration(opt.TimeOut)
if err != nil {
    return err
}
```

This would also require you to change the variable name where it is input to the `TaskRun`:

https://github.com/tektoncd/cli/blob/216f375412796f6c1086a31cb74e85ba212ff430/pkg/cmd/clustertask/start.go#L160

pkg/cmd/clustertask/start.go
```diff
@@ -130,7 +130,7 @@ like cat,foo,bar
 	c.Flags().BoolVarP(&opt.Last, "last", "L", false, "re-run the clustertask using last taskrun values")
 	c.Flags().StringSliceVarP(&opt.Labels, "labels", "l", []string{}, "pass labels as label=value.")
 	c.Flags().BoolVarP(&opt.ShowLog, "showlog", "", false, "show logs right after starting the clustertask")
-	c.Flags().Int64VarP(&opt.TimeOut, "timeout", "t", 3600, "timeout for taskrun in seconds")
+	c.Flags().StringVar(&opt.TimeOut, "timeout", "1h", "timeout for taskrun (default \"1h\")")
```
danielhelfand: My apologies for not explaining this better:

You actually don't need to include `(default \"1h\")` in the description. Cobra will add that piece to the description since you have specified the `1h` default.

kadern0: Thanks @danielhelfand now I got it. I've just updated the code with your suggestion.

danielhelfand:
This looks great! Thank you. Will keep you posted on the status of this, but I would say to expect that this will be introduced some time early next month.

danielhelfand:
/hold cancel 

Thanks for your patience @kadern0. Could you please sign the new CLA the Tekton project has adopted in the last month? Apologies, but I think this should be good to go.

kadern0:
Thanks @danielhelfand, I'm happy to sign the new CLA. Could you trigger the bot to verificate so I'm asked for it and I can sign it? (unless there is an easier way to do it)

danielhelfand:
Could you try doing a force push? That is unfortunately the only workaround that I have seen.

kadern0:
@danielhelfand, the force push did the trick (although I had another email set by default and it took me a few pushes to find out). 

Now that this is solved, in order to make my code pass the tests, the following warning has to be removed because `opt.Timeout` is no longer an int:
```
			if opt.TimeOut != 3600 {
				log.Println("WARNING: The --timeout flag will no longer be specified in seconds in v1.0.0. Learn more here: https://github.com/tektoncd/cli/issues/784")
				log.Println("WARNING: The -t shortand for --timeout will no longer be available in v1.0.0.")
			}
```

PR #813, opened 2020-03-17
Add an image command to push OCI images

sthaha:
# Changes

This command is able to create/push OCI compliant images that the Tekton
controller can resolve as a remote task reference. It technically can
upload more than just Task definitions.

# Submitter Checklist

review them:

for more details._

# Release Notes

```
A new image command that packages tekton resources into an OCI compliant image
```


vdemeester:
We discussed it a bit during the CLI working group. 
We felt it would be a safer bet to have this as a sub-command to `task` for now. This would allow us to have a `pull` command (or `fetch`) later on, and the same on `pipeline`. And it give us time to think about a top-level sub-command.

One additional thing is that the fact that it is publishing a image (pushing, pulling, …) is an "implementation" detail, hence having an `image` top-level sub-command felt weird.

/cc @sbwsg @danielhelfand 
/hold

sthaha:
> We felt it would be a safer bet to have this as a sub-command to task for now. This would allow us to have a pull command (or fetch) later on, and the same on pipeline. And it give us time to think about a top-level sub-command.

🤔 The intent of this command is to be able to package mulitple Tekton resource in one image. E.g. some pipelines, all associated tasks, resources, triggers, bindings etc into one image which is why I think this should become a top-level command instead of a subcommand of a particular resource.

vdemeester:
I realize I didn't proper convey the feedback from the WG and my discussions later with Pierre.

The general idea is to not rush having an `image` sub-command or any "creation of an oci image" in `tkn`. The **`oci` project it [`tektoncd/experimental`](https://github.com/tektoncd/experimental) is there to help us design properly that part**, gather user feedback. We definitely shouldn't rush adding a sub-command in the `tkn` cli — as we say, "in OSS, no is temporary, yes is forever" — deprecating or removing a sub-command is between hard and painful to plain not possible (`docker` is an example on this).

The `fetch`/`pull` part of this work is easier, and it feels safer to assume we could have it under the `task` or `pipeline` sub-commands.
- `tkn task pull` could pull the image and ask for which tasks to *get* from it (if more than one). Then should it apply to the cluster by default or output to `stdout` (so you could do `tkn task pull … | sed -e … | kubectl apply -f -`, …)
- `tkn pipeline pull` could pull the image and ask which pipeline to *get* **and** also get all related tasks, …

chmouel:
what about adding having this in mainline but as conditional with ldflag? so sunil or others can iterate in mainline without having to worry about support until stabilized,

vdemeester:
> what about adding having this in mainline but compiled in as conditional with ldflag? so sunil or others can iterate in mainline without having to worry about support until stabilized,

That could be a possibility, but:
- It adds a little bit of complexity (to build an experimental binary or not)
- It will bring some dependencies unused in the cli, only for this part (this is fine, but :upside_down_face:)
- This is **why** `tektoncd/experimental` exists, to experiment on stuff ; and a `cli` is really a good candidate to this (as it's not coupled to the controller behaviour, … it's easy to be managed on its own)

… and it is the same as building an image : I can use a tool to create an image that I run with something else (`buildah` for build, `docker` or `cri-o` for the run). I am not saying `tkn` shouldn't be able to create oci image/artifact with tasks and pipelines in it, I just wants us to take time to properly design this, experimenting and gathering feedback — and `tektoncd/experimental` exists for that reason

sthaha:
@vdemeester @chmouel does it then make sense to add an extn mechanism like the kubectl one? so that `experimental/oci` can just make a `tkn-image` binary and won't affect `tkn` code? I think I am close to be able to add that to `tkn` if it makes sense

PR #833, opened 2020-03-25
Fix Test Failure from go v1.14.0 Update

danielhelfand:
Updating to go v1.14.0 breaks `TestGetError` in `file/file_test.go`. This pull request adds quotes around the invalid url for the error message.

# Submitter Checklist

# Release Notes

```
N/A
```

### DIFF
--- a/pkg/file/file_test.go
+++ b/pkg/file/file_test.go
@@ -17,6 +17,7 @@ package file
 import (
 	"fmt"
 	"io/ioutil"
+	"strings"
 	"testing"
 
 	"github.com/ghodss/yaml"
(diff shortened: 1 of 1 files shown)


pkg/file/file_test.go
```diff
@@ -61,5 +61,5 @@ func TestGetError(t *testing.T) {
 	target := "httpz://foo.com/task.yaml"
 
 	_, err := LoadFileContent(p, target, IsYamlFile(), fmt.Errorf("invalid file format for %s: .yaml or .yml file extension and format required", target))
-	assert.Error(t, err, `Get httpz://foo.com/task.yaml: unsupported protocol scheme "httpz"`)
+	assert.Error(t, err, "Get \"httpz://foo.com/task.yaml\": unsupported protocol scheme \"httpz\"")
```
vdemeester: nit: using ` allows us to not have to *escape* `"`

danielhelfand:
> This will fail on go 1.13 though right ? 🤔
> (I wonder if we could do a fix that work on both 🙃)

We could parse the runtime go version and support both error output formats, but it's not pretty. I pushed up something for this to try.

vdemeester:
> We could parse the runtime go version and support both error output formats, but it's not pretty. I pushed up something for this to try.

Yeah this is ugly :stuck_out_tongue: I was more thinking of trying to use `string.Contains` with `unsupported protocol scheme` and maybe the url :stuck_out_tongue_closed_eyes:

danielhelfand:
> > We could parse the runtime go version and support both error output formats, but it's not pretty. I pushed up something for this to try.
> 
> Yeah this is ugly 😛 I was more thinking of trying to use `string.Contains` with `unsupported protocol scheme` and maybe the url 😝

Yeah, I over thought it. Changed to checking with `strings.Contains`.

PR #863, opened 2020-04-03
Use Timeout of --last or --use-pipelinerun/--use-taskrun for Start Commands

### LINKED ISSUE #488: Way of Testing tkn task start -t
/area testing

Currently, `tkn task start` features a `--timeout` or `-t` flag that allows a user to [set the timeout for a task run](https://github.com/tektoncd/cli/blob/ab673fa5ab2b48e5ce585961647d106cfc7ae1ab/pkg/cmd/task/start.go#L175). While the unit test case of actually setting the property is covered, it would be nice to have some sort of test that covers the case of setting the timeout to a value besides the default and testing if a task run times out appropriately. Testing this as an e2e case I think would be the way to go.

### LINKED ISSUE #862: timeout is not respected when using use-pipelinerun or --last
when using `tkn pipeline start pipl1 --use-pipelinerun piplr1` the timeout that
was set previously is not set on the newly created pipeline.

/kinds bugs

danielhelfand:
Closes #862
Closes #488  

# Changes

When using `tkn p start` with the `--last` flag or `--use-pipelinerun` flag, the timeout of the previous value isn't applied as described in #862. This pull request updates both `tkn p start` and `tkn t start` to use the timeout values of the previous PipelineRun or TaskRun.

# Submitter Checklist

review them:

# Release Notes

```
Use timeout value with --last and --use-pipelinerun/--use-taskrun with start commands
```

### DIFF
--- a/pkg/cmd/pipeline/start_test.go
+++ b/pkg/cmd/pipeline/start_test.go
@@ -492,6 +492,22 @@ func TestPipelineStart_ExecuteCommand(t *testing.T) {
 			wantError: true,
 			want:      "cannot use --last option with --filename option",
 		},
+		{
+			name: "Dry Run with --timeout specified",
+			command: []string{"start", "test-pipeline",
+				"-s=svc1",
+				"-r=source=scaffold-git",
+				"-p=pipeline-param=value1",
+				"-l=jemange=desfrites",
+				"-n", "ns",
+				"--dry-run",
+				"--timeout", "1s",
+			},
+			namespace:  "",
+			input:      cs2,
+			wantError:  false,
+			goldenFile: true,
+		},
 	}
 
 	for _, tp := range testParams {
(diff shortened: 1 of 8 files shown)


chmouel:
/lgtm
/approve

Thanks 👍🏼🔥

chmouel:
I was going to ask about unit test but unit test ask for themselves 😅

danielhelfand:
> I was going to ask about unit test but unit test asked for themselves 😅

Yeah, so I guess `tkn task start` with `--timeout` wasn't working before this change. So now it's showing up in unit tests. Let me update the golden files.

danielhelfand:
/hold

If we use similar testing approaches as with `--dry-run`, we can get some better unit testing here.

pkg/cmd/pipeline/start_test.go
```diff
@@ -1869,12 +1888,8 @@ func Test_start_pipeline_last(t *testing.T) {
 	pipeline := Command(p)
 	got, _ := test.ExecuteCommand(pipeline, "start", pipelineName,
 		"--last",
-		"-r=git-repo=scaffold-git",
```
danielhelfand, line 1872: The unit test for `tkn p start --last` wasn't really testing the feature since it overwrites many runtime values. The point of the test is to see whether the last pipelinerun's values are used, so I removed everything except `--last`.

PR #867, opened 2020-04-05
Read defaultTimeout for TaskRun and PipelineRun set in config-defaults

### LINKED ISSUE #865: tkn pipeline start does not use default timeout
# Version and Operating System

**tkn Version:** 0.8.0

**Operating System:**  macOS

# Expected Behavior

Pipelines started with `tkn pipeline start` should use the [configured default timeout](https://github.com/tektoncd/pipeline/blob/00194731669ca5a279ca92b2d94cfb85f6ae7f6d/docs/taskruns.md#configuring-default-timeout) from tekton pipelines.

# Actual Behavior

PipelineRuns created with `tkn pipeline start` get a timeout of `1h`.

1. Create these resources:
```yaml
---
kind: Task
metadata:
  name: timeout
spec:
  steps:
    - name: test
      image: busybox:latest
---
kind: Pipeline
metadata:
  name: timeout
spec:
  tasks:
    - name: test
      taskRef:
        name: timeout
```

2. Set a default timeout in tekton pipelines https://github.com/tektoncd/pipeline/blob/00194731669ca5a279ca92b2d94cfb85f6ae7f6d/docs/taskruns.md#configuring-default-timeout
```yaml
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: config-defaults
  namespace: tekton-pipelines
data:
  default-timeout-minutes: "180"
```

3. Start the pipeline run
```shell
$ tkn pipeline start timeout
Pipelinerun started: timeout-run-6gls4
...
```

4. See that the specified default timeout is not used:
```shell
$ kubectl get pr -o json timeout-run-6gls4 | jq '.spec.timeout'
"1h0m0s"
```

# Additional Info

waveywaves:
Read defaultTimeout for TaskRun and PipelineRun set in config-defaults

When `tkn task/pipeline start` is used, the default timeout value
configured in the config-defaults template is not consumed, and
the default timeout is taken as "1h".
To avoid this, we are reading the config-defaults ConfigMap in the
tekton-pipelines namespace and reading from key "default-timeout-minutes"
for the value of the default timeout set byt the user for their
instance of the pipeline controller.

# Changes
- Added `/pkg/config/config-default.go` which helps get the configMap `config-defaults` to check the TaskRun. 
- Used the above in `/pkg/cmd/{task,clustertask,pipeline}/start.go
- Removed support for `int64` typed timeout values and added support for `string` typed timeout values as it is done in task and clustertask

fixes #865 

# Submitter Checklist

review them:

for more details._

# Release Notes

```
release-note
```

### DIFF
--- a/docs/cmd/tkn_clustertask_start.md
+++ b/docs/cmd/tkn_clustertask_start.md
@@ -39,7 +39,7 @@ like cat,foo,bar
   -p, --param stringArray        pass the param as key=value or key=value1,value2
   -s, --serviceaccount string    pass the serviceaccount name
       --showlog                  show logs right after starting the clustertask
-  -t, --timeout int              timeout for taskrun in seconds (default 3600)
+  -t, --timeout string           timeout for taskrun (default "1h")
 ```
 
 ### Options inherited from parent commands
--- a/docs/man/man1/tkn-clustertask-start.1
+++ b/docs/man/man1/tkn-clustertask-start.1
@@ -60,8 +60,8 @@ Start clustertasks
     show logs right after starting the clustertask
 
 .PP
-\fB\-t\fP, \fB\-\-timeout\fP=3600
-    timeout for taskrun in seconds
+\fB\-t\fP, \fB\-\-timeout\fP="1h"
+    timeout for taskrun
 
 
 .SH OPTIONS INHERITED FROM PARENT COMMANDS
--- a/pkg/cmd/clustertask/start.go
+++ b/pkg/cmd/clustertask/start.go
@@ -18,7 +18,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"log"
 	"sort"
 	"strings"
 	"time"
@@ -27,6 +26,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/tektoncd/cli/pkg/cli"
 	"github.com/tektoncd/cli/pkg/cmd/taskrun"
+	pipelineconfig "github.com/tektoncd/cli/pkg/config"
 	"github.com/tektoncd/cli/pkg/flags"
 	"github.com/tektoncd/cli/pkg/labels"
 	"github.com/tektoncd/cli/pkg/options"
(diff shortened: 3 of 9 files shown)


waveywaves:
@danielhelfand Okay, so basically, this means that we should stop setting the default timeout from cli side also. But it is also important to know what the default timeout is on the controller (via config-defaults) to set the default timeout here. Otherwise setting 1h for every pipeline that doesn't have an explicit timeout given, where it should be setting defaultTimeout from the pipeline is still buggy behaviour.

>I don't think this approach is necessary as far as reading values from the ConfigMap to get the values. I was thinking what probably makes more sense is to have this handled by pipelines, which could apply the value of default-timeout-minutes if nothing is specified for timeout.

This would mean that, when a TaskRun is created with a timeout of value "1h" (default on cli) it would have to be intercepted and checked by the controller against the config-defaults timeout value (default on pipeline) and then the TImeout on the TaskRun needs to be mutated accordingly.

Obviously, this is a question of 1 mutation (intercept Taskrun, check and change timeout) vs 2 API calls (1. Get configmap, 2. create taskrun).

If we go with the former as you proposed, we will have to keep the default timeout on cli as static without the user able to set the default on the cli side, which also means that any feature request to store default cli config would have to be declined. Don't know if we will have any request of the sort, just thinking out loud.

danielhelfand:
What I was suggesting is the cli should not set anything for timeout unless specified by the user via the `--timeout` flag (i.e. no default for `--timeout` from `tkn`). This would allow pipelines to do all the work as far as setting the traditional default of `1h` as well as checking the ConfigMap for the custom default value for timeout.  

That being said, there are other issues that need to be resolved before to make sure we are approaching this correctly. Such as what is being worked on in #793 (i.e. the cli can't just introduce a breaking change like what is being done in this pr without giving warning) and the issue with `default-timeout-minutes` not being properly set with pipelines. The issue with pipelines should also be triaged before being picked up.

vdemeester:
> What I was suggesting is the cli should not set anything for timeout unless specified by the user via the `--timeout` flag (i.e. no default for `--timeout` from `tkn`). This would allow pipelines to do all the work as far as setting the traditional default of `1h` as well as checking the ConfigMap for the custom default value for timeout.

Yes, that should be the case.

Issue #820, opened 2020-03-19
How to start a pipeline with default params in a non-interactive shell


ckadner:
# Version and Operating System

**`tkn` version:**
```
Client version: 0.8.0
Pipeline version: v0.11.0-rc2
```

**Operating System:** MacOS 10.15 (Catalina)

# Expected Behavior

I was expecting there to be a flag that allows to run pipelines/tasks with parameter default values (if so defined) without prompting for user input, i.e.:

`tkn pipeline start <pipeline-name> --params-use-defaults`

This would be usefull when running pipelines from within a shell script

# Actual Behavior

When starting a pipeline, user get's prompted to confirm or override the default values for each pipeline parameter.

# Additional Info

```YAML
apiVersion: tekton.dev/v1alpha1
kind: Task
metadata:
  name: gcs-download
spec:
  inputs:
    params:
    - name: url
  steps:
  - args:
    - gsutil cat $0 | tee $1
    - $(inputs.params.url)
    - /tmp/results.txt
    command:
    - sh
    - -c
    image: google/cloud-sdk:216.0.0
    name: gcs-download
---
apiVersion: tekton.dev/v1alpha1
kind: Task
metadata:
  name: echo
spec:
  inputs:
    params:
    - name: path
  steps:
  - args:
    - echo "$0"
    - $(inputs.params.path)
    command:
    - sh
    - -c
    image: library/bash:4.4.23
    name: echo
---
apiVersion: tekton.dev/v1alpha1
kind: Pipeline
metadata:
  annotations:
    pipelines.kubeflow.org/pipeline_spec: '{"description": "A pipeline with two sequential
      steps.", "inputs": [{"default": "gs://ml-pipeline-playground/shakespeare1.txt",
      "name": "url", "optional": true}, {"default": "/tmp/results.txt", "name": "path",
      "optional": true}], "name": "Sequential pipeline"}'
  name: sequential-pipeline
spec:
  params:
  - default: gs://ml-pipeline-playground/shakespeare1.txt
    name: url
  - default: /tmp/results.txt
    name: path
  tasks:
  - name: gcs-download
    params:
    - name: url
      value: $(params.url)
    taskRef:
      name: gcs-download
  - name: echo
    params:
    - name: path
      value: $(params.path)
    runAfter:
    - gcs-download
    taskRef:
      name: echo
```

ppitonak:
Possible duplicate of https://github.com/tektoncd/cli/issues/736

Issue #736, opened 2020-02-17
UX: don't ask the user to input value for params with a default value


ppitonak:
# Version and Operating System

**tkn Version:**

0.7.1

**Operating System:** 

All

# Expected Behavior

When I start a pipeline and I don't specify all params, use default values.

# Actual Behavior

* It asks me for parameter value and shows me the default value.
* It is not possible to create a pipeline in non-interactive way although `tkn` has all data it needs.

# Steps to Reproduce the Problem

1. create everything but pipeline run from https://github.com/tektoncd/pipeline/blob/4b6a3901ecf753ee56ab9b67fe8296134bdbd426/examples/pipelineruns/conditional-pipelinerun.yaml
2. `tkn p start conditional-pipeline --showlog -s default -r source-repo=pipeline-git`

result: 
```
? Value for param `path` of type `string`? (Default is `README.md`) (README.md)
```

# Additional Info

piyush-garg:
You can just press enter and it will take the default value.

But it is nice use-case to run without interactiveness and we already have an issue for that in #571

ckadner:
Not prompting for parameters that have default values would also help when trying to start a pipeline in a script (#820)

ppitonak:
/remove-lifecycle stale

`tkn task start` and `tkn pipeline start` have parameter `--use-param-defaults` but it's missing in `tkn clustertask start`

Issue #571, opened 2020-01-07
`tkn` should not be interactive if we are not running with a tty


chmouel:
#Expected Behavior

Don't try interactiveness if we don't run interactively (ie in scripts or others), for example on `tkn pipelinerun logs` just fail if no argument is specified

# Actual Behavior

Do some interactiveness stuff with survey

# Steps to Reproduce the Problem

1. Write a shell script that do a `tkn pipelinerun logs` 
2. Should not be interactive

# Additional Info

That should detect a terminal or not https://github.com/mattn/go-isatty/

/kind bug

ckadner:
Also applies to starting a task/pipeline that has default values for all input parameters (#820)

PR #868, opened 2020-04-06
Flag to start a Pipeline without prompting for default value

anshulvermapatel:
When we start a Pipeline, it prompts for asking the parameter values even if the default values are already set in the pipeline configuration and if we just press enter it uses the default values and creates a PipelineRun.

There are many situations where we would need to run the `$ tkn pipeline start` command in a script and there getting a prompt would not be desirable because at that time it won't be running interactively or on an interactive shell.

Hence, I introduce a flag `--params-use-defaults`(the name can be changed if required) which when used will not prompt if the params already have default values configured and will use the default values.

Resolves [820](https://github.com/tektoncd/cli/issues/820), [736](https://github.com/tektoncd/cli/issues/736) and [571](https://github.com/tektoncd/cli/issues/571)
Do let me if there are any questions or any changes to make.

# Changes

# Submitter Checklist

review them:

for more details._

# Release Notes

```
release-note
```


anshulvermapatel:
@googlebot I signed it!

pkg/cmd/pipeline/start.go
```diff
@@ -152,6 +153,7 @@ like cat,foo,bar
 	c.Flags().StringVarP(&opt.PrefixName, "prefix-name", "", "", "specify a prefix for the pipelinerun name (must be lowercase alphanumeric characters)")
 	c.Flags().StringVarP(&opt.TimeOut, "timeout", "", "1h", "timeout for pipelinerun")
 	c.Flags().StringVarP(&opt.Filename, "filename", "f", "", "local or remote file name containing a pipeline definition to start a pipelinerun")
+	c.Flags().BoolVarP(&opt.ParamUseDefaults, "params-use-defaults", "", false, "allow using deafult parameter values without prompting for input")
```
danielhelfand, line 156: I wonder if a flag-based approach is the right way to go about this. Maybe what the better approach would be is to not prompt the user at all if a default is specified for a param and the user specifies nothing, and prompt if there is no default value and nothing is specified via -p. 

cc @piyush-garg

anshulvermapatel: @danielhelfand , Yes, I agree with you, I too thought so, but I thought that would be something which will be removed(removal of prompting), and I think the members should decide that. If you want I can change the PR to do that.

vdemeester: > I wonder if a flag-based approach is the right way to go about this. Maybe what the better approach would be is to not prompt the user at all if a default is specified for a param and the user specifies nothing, and prompt if there is no default value and nothing is specified via `-p`.
> 
> cc @piyush-garg

Even if we would like that approach, this cannot change in one go (we need to warn of the default behavior change, wait for a release or too and then switch). 
To be honest, I do like the fact that, by default, it asks me for values even if they have a default. It feels more user friendly for me, as it lets me override them if I need.

/cc @chmouel @pradeepitm12

chmouel: I think that was a @siamaksade request to have default...

anshulvermapatel:
@googlebot I fixed it

anshulvermapatel:
@googlebot I signed it!

pkg/cmd/pipeline/start.go
danielhelfand: Yes, I do appreciate interactive mode here and allowing the user flexibility as far as picking a default or overriding it as well. I think the main thing would be with whatever approach is recommended is that a user doesn't have to only use defaults in a non-interactive scenario. 

Maybe a list of params with defaults could be passed to pick which param defaults to use via a flag? We could also support an all param defaults flag as well? What I want to make sure of is that a user can pick certain param defaults to use while also overriding params that have defaults.

anshulvermapatel: @danielhelfand , One approach n mind is -
- If a param having default configured and not been specified with -p should take the default value.
- If a param having default configured and its value is specified with -p should override the default value.
- if any param does not have a default configured and its value is not specified with -p should give an interactive prompt to give the value.

What do you guys say? @vdemeester @chmouel

siamaksade: I find it confusing not to prompt user for parameters that have a default value, unless user specifically asks for this e.g. via a flag. "-p" should be the only indicator for not promoting otherwise IMO user should still get prompted but the default value should be displayed (current behaviour) so they can press enter and continue.

danielhelfand: I think the idea of the flag in this pr would be to not have to use `-p` to use param defaults while also allowing params with defaults to still be overrode using `-p`. Basically the idea would be to have a shortcut to specifying to use default values as opposed to having to use `-p` when the default is to be used. I think the idea makes sense from that perspective. And then prompting would remain exactly as is.

anshulvermapatel: @danielhelfand , yes, that is the idea .

danielhelfand, line 156: For the flag name, maybe `--use-param-default` instead?

Also slight rewording for the description:

```suggestion
	c.Flags().BoolVarP(&opt.ParamUseDefaults, "use-param-defaults", "", false, "use default parameter values without prompting for input")
```

danielhelfand: Yup, sorry for the confusion in the beginning. Glad we talked it out though and have clarity that this is the right approach.

anshulvermapatel: Then I think we should also change the `ParamUseDefaults` field `startOptions` struct to `UseParamDefaults`. Let me update the PR with these changes.

anshulvermapatel:
@danielhelfand 
1. Could you add unit tests for this feature under pipeline/start_test.go?
I am updating the PR which will include the tests.
2. We'll also want this same functionality for tkn task start/tkn clustertask start. Any thoughts on whether you would prefer to add them to this pr or separate ones?
Let's keep this one for `pipeline` only. I will create a new PR to include this functionality for `task` and `clustertask` in that same PR.

anshulvermapatel:
@googlebot I fixed it

anshulvermapatel:
@googlebot I signed it!

pkg/cmd/pipeline/start_test.go
```diff
@@ -454,6 +454,22 @@ func TestPipelineStart_ExecuteCommand(t *testing.T) {
...
+		{
+			name: "Dry Run with use param defaults and output=json",
+			command: []string{"start", "test-pipeline",
+				"-s=svc1",
+				"-r=source=scaffold-git",
+				"--use-param-defaults",
+				"-l=jemange=desfrites",
+				"-n", "ns",
+				"--dry-run",
+				"--output", "json",
```
danielhelfand, line 466: You can remove `--output json` as this is sufficiently covered by other tests, but think the test idea is great.

pkg/cmd/pipeline/start.go
```diff
@@ -152,6 +153,7 @@ like cat,foo,bar
 	c.Flags().StringVarP(&opt.PrefixName, "prefix-name", "", "", "specify a prefix for the pipelinerun name (must be lowercase alphanumeric characters)")
 	c.Flags().StringVarP(&opt.TimeOut, "timeout", "", "1h", "timeout for pipelinerun")
 	c.Flags().StringVarP(&opt.Filename, "filename", "f", "", "local or remote file name containing a pipeline definition to start a pipelinerun")
+	c.Flags().BoolVarP(&opt.UseParamDefaults, "use-param-defaults", "", false, "use deafult parameter values without prompting for input")
```
danielhelfand, line 156: typo: `deafult` -> `default`

pkg/cmd/pipeline/start_test.go
```diff
@@ -454,6 +454,22 @@ func TestPipelineStart_ExecuteCommand(t *testing.T) {
...
+			},
+			namespace:  "",
+			input:      cs2,
+			wantError:  false,
+			goldenFile: true,
+		},
```
danielhelfand, line 472: Can you also add another test for the case where a value is overrode using `-p` that has a default, but `--use-param-defaults` is also used to still appropriately use the defaults of other values?

pkg/cmd/pipeline/start.go
anshulvermapatel: oh sorry, will make the changes.

pkg/cmd/pipeline/start_test.go
anshulvermapatel: Sure will add it.

pkg/cmd/pipeline/start_test.go
anshulvermapatel: I wanted to test that the default value are actually getting populated or not, and that is why json output is matched with the golden file I created and they actually get populated, tests get passed :).

danielhelfand: Right, but not using `output` `json` will just use the default format of yaml instead, so it should still produce a golden file with appropriate output.

danielhelfand:
@anshulvermapatel You'll also want to run `make generated` before pushing up to regenerate docs for the CLI. That's the reason for build-test failures.

pkg/cmd/pipeline/start_test.go
anshulvermapatel: Then I have to change the golden file as well .. sure will do that ..

anshulvermapatel:
> @anshulvermapatel You'll also want to run `make generated` before pushing up to regenerate docs for the CLI. That's the reason for build-test failures.

Yaa I came to know about it, I will take care of it in my next update.

anshulvermapatel:
@danielhelfand , Please read the commit message.
I have made some new changes, created some new functions. Please do have a look. But it works fine now. Thank God you asked me to write the test to check if defaults are included for other params if one or more params are given through `-p`

I came to know about another issue. (don't know exactly if it is an issue or expected behavior)
When no param is given with `-p`, an interactive prompt asks for the params value, but if a param is given through `-p` then I think it still should ask for the value of other params through the prompt if `--use-param-default` is not included in the command. 
What do you say, shall I open an issue for this?

danielhelfand:
> I came to know about another issue. (don't know exactly if it is an issue or expected behavior)
When no param is given with `-p`, an interactive prompt asks for the params value, but if a param is given through `-p` then I think it still should ask for the value of other params through the prompt if `--use-param-default` is not included in the command.

I understand your perspective on it, but I think the thought was interactive mode only if nothing is specified for resources or params. We can discuss that further whether the feature could be changed, but I believe it's intended.

You'll need to rebase your pr here, but then I can test it out.

anshulvermapatel:
@danielhelfand 
Regarding the 'pull-tekton-cli-unit-tests' test failing.
~~~
 TestPipelineStart_ExecuteCommand/Dry_Run_with_only_--dry-run_with_use_param_defaults_and_specified_params: start_test.go:632: assertion failed: 
        --- expected
        +++ actual
        @@ -9,8 +9,8 @@
         spec:
         ··params:
        +··-·name:·rev-param
        +····value:·revision
         ··-·name:·pipeline-param
         ····value:·value1
        -··-·name:·rev-param
        -····value:·revision
         ··pipelineRef:
         ····name:·test-pipeline
~~~
The saw that the sequence of `params` varries that is why, if you see the default values are getting populated correctly but since they are in a different sequence, this test is failing.

Regarding the `cla/google`, I have signed the agreement, still the googlebot say that there is an issue, I do not understand where is the issue, can you please help me on this.(I changed corrected my email id for the commit)

danielhelfand:
> @danielhelfand
> Regarding the 'pull-tekton-cli-unit-tests' test failing.
> 
> ```
>  TestPipelineStart_ExecuteCommand/Dry_Run_with_only_--dry-run_with_use_param_defaults_and_specified_params: start_test.go:632: assertion failed: 
>         --- expected
>         +++ actual
>         @@ -9,8 +9,8 @@
>          spec:
>          ··params:
>         +··-·name:·rev-param
>         +····value:·revision
>          ··-·name:·pipeline-param
>          ····value:·value1
>         -··-·name:·rev-param
>         -····value:·revision
>          ··pipelineRef:
>          ····name:·test-pipeline
> ```
> 
> The saw that the sequence of `params` varries that is why, if you see the default values are getting populated correctly but since they are in a different sequence, this test is failing.

We should add a sort of the param values by name to assure consistent order. See how [pipelineresources are sorted](https://github.com/tektoncd/cli/blob/54aece7b833d7507f4b52e43a814d6b732179d8f/pkg/cmd/pipeline/start.go#L589) for reference. 

> Regarding the `cla/google`, I have signed the agreement, still the googlebot say that there is an issue, I do not understand where is the issue, can you please help me on this.(I changed corrected my email id for the commit)

I am not sure why, but can reach out on slack to others who might be able to help.

anshulvermapatel:
@danielhelfand  everything looks fine except the 'cla/google' , I have my correct email address in my latest commit, I am not sure, why this bot still fails the test.
I am seeking help on slack for this.

anshulvermapatel:
@googlebot I signed it!

anshulvermapatel:
@danielhelfand , I have created a [new PR](https://github.com/tektoncd/cli/pull/889) for this one as I was not able to fix the CLA test issue as I messed up with my email id in my commit.
I request you to please close this on and continue on the other one.
I am quite sure, the new one is merge-ready as all the changes we discussed here are there in that one. Please do have a look.
Thank You .

Issue #872, opened 2020-04-06
Interactive Mode for Describe Commands


danielhelfand:
# Expected Behavior

Similar to what is offered for viewing logs of PipelineRuns/TaskRuns, it would be nice to select what object to describe via an interactive approach. The idea would be that if no name argument is passed to a describe command, then a list will appear with items for the user to select from. The user can select which option to describe and then the description template will be filled out accordingly based on the object selected by the user.

An example would be with PipelineRuns:

**No argument provided results in interactive list:**

```
$ tkn pr desc
? Select pipelinerun:  [Use arrows to move, type to filter]
> deploy-pipeline-hello2-run-8ckwc started 18 hours ago
  deploy-pipeline-hello2-run-x2df8 started 18 hours ago
  deploy-pipeline-hello2-run-65tjl started 18 hours ago
  deploy-pipeline-hello2-run-8mbh4 started 18 hours ago
  deploy-pipeline-hello2-run-n8tnp started 18 hours ago
```

While PipelineRuns/TaskRuns are really good use cases for this since there are many of them/they all have uniques names, I don't see why this approach can't be applied to all describe commands throughout `tkn`. 

# Additional Info

Related to #871

anshulvermapatel:
@danielhelfand , do you intend you have this simple ask to select the `pr` or you also intend to have `FuzzyAsk` (Fzf) also which we give with `-F` or `--fzf` option in case of `logs` command

danielhelfand:
I see no reason why supporting both aren't applicable here since both are available with `logs`.

anshulvermapatel:
@danielhelfand , Okay .. I am creating a PR with that. Please have a look.

vinamra28:
- [X] PipelineRun
- [X] Pipeline
- [X] Task
- [X] ClusterTask
- [X] TaskRun
- [ ] Condition
- [ ] EventListener (#1353)
- [ ] TriggerBinding
- [ ] ClusterTriggerBinding
- [ ] TriggerTemplate

PuneetPunamiya:
- [x] PipelineRun
- [x] Pipeline
- [x] Task
- [x] ClusterTask
- [x] TaskRun
- [x] Condition
- [x] EventListener (#1353)
- [x] TriggerBinding (https://github.com/tektoncd/cli/pull/1356)
- [x] ClusterTriggerBinding (https://github.com/tektoncd/cli/pull/1356)
[...]

PR #890, opened 2020-04-11
Provides an interactive Prompt to select a PipelineRun/TaskRun 

anshulvermapatel:
Provides an interactive Prompt to select a PipelineRun/TaskRun to describe just like what is there in `logs` subcommand.
~~~
$ ./tkn pr desc
? Select pipelinerun:  [Use arrows to move, type to filter]
> sequential-pipeline-run-t4jmp started 1 hour ago
   sequential-pipeline-run-xwcxr started 1 hour ago
   sequential-pipeline-run-kfqfh started 1 hour ago
   sequential-pipeline-run-87jh2 started 1 hour ago
   sequential-pipeline-run-lvcwh started 1 hour ago
    
$ ./tkn tr desc
? Select taskrun:  [Use arrows to move, type to filter]
> sequential-pipeline-run-xwcxr-echo-8vqjv started 1 hour ago
   sequential-pipeline-run-87jh2-echo-gwwsj started 1 hour ago
   sequential-pipeline-run-t4jmp-echo-cz7t6 started 1 hour ago
   sequential-pipeline-run-kfqfh-echo-pfstb started 1 hour ago
   sequential-pipeline-run-lvcwh-echo-dbzc6 started 1 hour ago
~~~
    
Resolves Issue - [#872](https://github.com/tektoncd/cli/issues/872)
cc @danielhelfand 

# Changes

# Submitter Checklist

review them:

for more details._

# Release Notes

```
release-note
```

### DIFF
--- a/pkg/cmd/clustertask/describe.go
+++ b/pkg/cmd/clustertask/describe.go
@@ -27,6 +27,7 @@ import (
 	"github.com/tektoncd/cli/pkg/cli"
 	"github.com/tektoncd/cli/pkg/clustertask"
 	"github.com/tektoncd/cli/pkg/formatted"
+	"github.com/tektoncd/cli/pkg/options"
 	"github.com/tektoncd/cli/pkg/task"
 	"github.com/tektoncd/cli/pkg/taskrun/list"
 	trsort "github.com/tektoncd/cli/pkg/taskrun/sort"
@@ -114,6 +115,7 @@ NAME	STARTED	DURATION	STATUS
 
 func describeCommand(p cli.Params) *cobra.Command {
 	f := cliopts.NewPrintFlags("describe")
+	opts := &options.DescribeOptions{Params: p}
 	eg := `Describe a ClusterTask of name 'foo':
 
     tkn clustertask describe foo
@@ -131,7 +133,6 @@ or
 		Annotations: map[string]string{
 			"commandType": "main",
 		},
-		Args:         cobra.MinimumNArgs(1),
 		SilenceUsage: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			s := &cli.Stream{
@@ -145,12 +146,21 @@ or
 				return err
 			}
 
-			if output != "" {
-				clustertaskGroupResource := schema.GroupVersionResource{Group: "tekton.dev", Resource: "clustertasks"}
-				return actions.PrintObject(clustertaskGroupResource, args[0], cmd.OutOrStdout(), p, f, "")
+			if len(args) == 0 {
+				err = askClusterTaskName(opts, p)
+				if err != nil {
+					return err
+				}
+			} else {
+				opts.ClusterTaskName = args[0]
 			}
 
-			return printClusterTaskDescription(s, p, args[0])
+			if output != "" {
+				clustertaskGroupResource := schema.GroupVersionResource{Group: "tekton.dev", Resource: "clustertasks"}
+				return actions.PrintObject(clustertaskGroupResource, opts.ClusterTaskName, cmd.OutOrStdout(), p, f, "")
+			}
+
+			return printClusterTaskDescription(s, p, opts.ClusterTaskName)
 		},
 	}
 
@@ -233,4 +243,25 @@ func sortResourcesByTypeAndName(tres []v1beta1.TaskResource) []v1beta1.TaskResou
 	})
 
 	return tres
+}
+
+func askClusterTaskName(opts *options.DescribeOptions, p cli.Params) error {
+	cs, err := p.Clients()
+	if err != nil {
+		return err
+	}
+	clusterTaskNames, err := allClusterTaskNames(cs)
+	if err != nil {
+		return err
+	}
+	if len(clusterTaskNames) == 0 {
+		return fmt.Errorf("no clustertasks found")
+	}
+
+	err = opts.Ask(options.ResourceNameClusterTask, clusterTaskNames)
+	if err != nil {
+		return err
+	}
+
+	return nil
 }
--- a/pkg/cmd/pipelineresource/describe.go
+++ b/pkg/cmd/pipelineresource/describe.go
@@ -22,6 +22,7 @@ import (
 	"text/template"
 
 	"github.com/tektoncd/cli/pkg/formatted"
+	"github.com/tektoncd/cli/pkg/options"
 	"github.com/tektoncd/cli/pkg/printer"
 	validateinput "github.com/tektoncd/cli/pkg/validate"
 
@@ -62,6 +63,7 @@ const templ = `{{decorate "bold" "Name"}}:	{{ .PipelineResource.Name }}
 `
 
 func describeCommand(p cli.Params) *cobra.Command {
+	opts := &options.DescribeOptions{Params: p}
 	f := cliopts.NewPrintFlags("describe")
 	eg := `Describe a PipelineResource of name 'foo' in namespace 'bar':
 
@@ -77,7 +79,6 @@ or
 		Aliases:      []string{"desc"},
 		Short:        "Describes a pipeline resource in a namespace",
 		Example:      eg,
-		Args:         cobra.MinimumNArgs(1),
 		SilenceUsage: true,
 		Annotations: map[string]string{
 			"commandType": "main",
@@ -98,11 +99,20 @@ or
 				return err
 			}
 
-			if output != "" {
-				return describePipelineResourceOutput(cmd.OutOrStdout(), p, f, args[0])
+			if len(args) == 0 {
+				err = askPipelineResourceName(opts, p)
+				if err != nil {
+					return err
+				}
+			} else {
+				opts.PipelineResourceName = args[0]
 			}
 
-			return printPipelineResourceDescription(s, p, args[0])
+			if output != "" {
+				return describePipelineResourceOutput(cmd.OutOrStdout(), p, f, opts.PipelineResourceName)
+			}
+
+			return printPipelineResourceDescription(s, p, opts.PipelineResourceName)
 		},
 	}
 
@@ -166,4 +176,27 @@ func printPipelineResourceDescription(s *cli.Stream, p cli.Params, preName strin
 		return err
 	}
 	return w.Flush()
+}
+
+func askPipelineResourceName(opts *options.DescribeOptions, p cli.Params) error {
+	cliClients, err := p.Clients()
+	if err != nil {
+		return err
+	}
+
+	// the `allPipelineResourceNames( )` func is in delete.go, in the pipelineresource package.
+	pipelineResourceNames, err := allPipelineResourceNames(p, cliClients)
+	if err != nil {
+		return err
+	}
+	if len(pipelineResourceNames) == 0 {
+		return fmt.Errorf("no pipelineresources found")
+	}
+
+	err = opts.Ask(options.ResourceNamePipelineResource, pipelineResourceNames)
+	if err != nil {
+		return err
+	}
+
+	return nil
 }
--- a/pkg/cmd/pipelinerun/describe.go
+++ b/pkg/cmd/pipelinerun/describe.go
@@ -21,14 +21,22 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/tektoncd/cli/pkg/actions"
 	"github.com/tektoncd/cli/pkg/cli"
+	"github.com/tektoncd/cli/pkg/options"
+	prhelper "github.com/tektoncd/cli/pkg/pipelinerun"
 	prdesc "github.com/tektoncd/cli/pkg/pipelinerun/description"
 	"github.com/tektoncd/cli/pkg/validate"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	cliopts "k8s.io/cli-runtime/pkg/genericclioptions"
 )
 
+const (
+	defaultDescribeLimit = 5
+)
+
 func describeCommand(p cli.Params) *cobra.Command {
 	f := cliopts.NewPrintFlags("describe")
+	opts := &options.DescribeOptions{Params: p}
 	eg := `Describe a PipelineRun of name 'foo' in namespace 'bar':
 
     tkn pipelinerun describe foo -n bar
@@ -43,7 +51,6 @@ or
 		Aliases:      []string{"desc"},
 		Short:        "Describe a pipelinerun in a namespace",
 		Example:      eg,
-		Args:         cobra.MinimumNArgs(1),
 		SilenceUsage: true,
 		Annotations: map[string]string{
 			"commandType": "main",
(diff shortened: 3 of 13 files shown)


danielhelfand:
Haven't reviewed this yet, but will make time tomorrow to look it over. Just wanted to note this does not fully resolve #872 as the issue was meant for all `describe` commands across the CLI, not just TaskRun/PipelineRun. I have updated the issue with a list to keep track of progress made on the issue.

anshulvermapatel:
@danielhelfand , implemented `describe` for other commands as well `pipeline`,`pipelinerun`,`task`,`taskrun`,`clustertask` and `pipelineresource` .

anshulvermapatel:
@danielhelfand , Can you please give this a look

pkg/cmd/clustertask/describe.go
```diff
@@ -112,8 +113,13 @@ NAME	STARTED	DURATION	STATUS
 {{- end }}
 `
 
+const (
+	defaultDescribeLimit = 5
```
danielhelfand, line 117: In the case of non TaskRun/PipelineRun resources, I wonder if a limit is necessary. For TaskRuns/PipelineRuns, it makes sense since they are ordered by start time. However, all other objects are just ordered by name. Perhaps a better approach would be to just list all available in the namespace.

pkg/cmd/pipelinerun/describe.go
```diff
@@ -73,8 +99,40 @@ or
...
+		return err
+	}
+
+	prs, err := prhelper.GetAllPipelineRuns(opts.Params, lOpts, opts.Limit)
+
+	if err != nil {
+		return err
+	}
+	if len(prs) == 0 {
+		return fmt.Errorf("No pipelineruns found")
```
danielhelfand, line 125: Always use [lowercase for beginning an error message](https://github.com/golang/go/wiki/CodeReviewComments#error-strings). We need to be better about enforcing this in the CLI moving forward, so apologies if you saw this elsewhere in the CLI.

```suggestion
		return fmt.Errorf("no pipelineruns found")
```

pkg/cmd/pipelinerun/describe.go
```diff
@@ -64,6 +71,25 @@ or
...
+				err = askPipelineRunName(opts, p)
+				if err != nil {
+					return nil
+				}
+				if output != "" {
+					pipelineRunGroupResource := schema.GroupVersionResource{Group: "tekton.dev", Resource: "pipelineruns"}
+					return actions.PrintObject(pipelineRunGroupResource, opts.PipelineRunName, cmd.OutOrStdout(), p, f, p.Namespace())
+				}
+
+				return prdesc.PrintPipelineRunDescription(s, opts.PipelineRunName, p)
```
danielhelfand, line 90: There is no need to repeat this. This is already happening outside the `if len(args) == 0 { .. }`. I would suggest finding a way to capture the PipelineRun name selected and use it in the logic below:

```go
if output != "" {
    pipelineRunGroupResource := schema.GroupVersionResource{Group: "tekton.dev", Resource: "pipelineruns"}
    return actions.PrintObject(pipelineRunGroupResource, args[0], cmd.OutOrStdout(), p, f, p.Namespace())
}

return prdesc.PrintPipelineRunDescription(s, args[0], p)
```

pkg/cmd/task/describe.go
```diff
@@ -133,7 +139,7 @@ or
 		Annotations: map[string]string{
 			"commandType": "main",
 		},
-		Args:         cobra.MinimumNArgs(1),
+		//Args:         cobra.MinimumNArgs(1),
```
danielhelfand, line 142: Remove commented out code.

pkg/options/describe_test.go
```diff
@@ -0,0 +1,286 @@
+// Copyright © 2019 The Tekton Authors.
```
danielhelfand, line 1:
```suggestion
// Copyright © 2020 The Tekton Authors.
```

pkg/options/describe_test.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	testParams := []struct {
+		name     string
+		resource string
+		prompt   prompt.Prompt
+		options  []string
+		want     DescribeOptions
+	}{
+		{
```
danielhelfand, line 111: All of the tests here seem to only be testing selecting the first object. Please add more tests to select a different object than just the first in the list.

pkg/options/describe_test.go
```diff
@@ -0,0 +1,286 @@
...
+			want: DescribeOptions{
+				PipelineName:         "",
+				PipelineRunName:      "",
+				TaskName:             "",
+				TaskrunName:          "sample-task-run1",
+				PipelineResourceName: "",
+				ClusterTaskName:      "",
+			},
+		},
+		// Mine
```
danielhelfand, line 161: Remove this

pkg/options/describe.go
```diff
@@ -0,0 +1,163 @@
...
+				Options: options,
+			},
+		},
+	}
+
+	if err := survey.Ask(qs, &ans, opts.AskOpts); err != nil {
+		return err
+	}
+
+	// The Resource* constants are defined in options.go
```
danielhelfand, line 90:
```suggestion
	// The Resource constants are defined in options.go
```

pkg/options/describe.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/AlecAivazis/survey/v2/terminal"
+	"github.com/fatih/color"
+	"github.com/ktr0731/go-fuzzyfinder"
+	"github.com/tektoncd/cli/pkg/cli"
+	prdesc "github.com/tektoncd/cli/pkg/pipelinerun/description"
+	trdesc "github.com/tektoncd/cli/pkg/taskrun/description"
+)
+
+/*
```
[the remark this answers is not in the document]
anshulvermapatel: Done

pkg/options/describe.go
```diff
@@ -0,0 +1,157 @@
+// Copyright © 2019 The Tekton Authors.
```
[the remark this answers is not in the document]
anshulvermapatel: done

pkg/cmd/taskrun/describe.go
```diff
@@ -73,6 +119,9 @@ or
 		},
 	}
 
+	c.Flags().IntVarP(&opts.Limit, "limit", "", defaultTaskRunLimit, "lists number of taskruns")
```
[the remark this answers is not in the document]
anshulvermapatel: Done

pkg/cmd/pipelinerun/describe.go
```diff
@@ -74,6 +119,9 @@ or
 		},
 	}
 
+	c.Flags().IntVarP(&opts.Limit, "limit", "", defaultDescribeLimit, "lists number of pipelineruns")
```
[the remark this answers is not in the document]
anshulvermapatel: done

pkg/cmd/pipelinerun/describe.go
```diff
@@ -74,6 +119,9 @@ or
...
+	c.Flags().BoolVarP(&opts.Fzf, "fzf", "F", false, "use fzf to select a pipelinerun")
```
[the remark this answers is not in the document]
anshulvermapatel: done

pkg/cmd/taskrun/describe.go
```diff
@@ -73,6 +119,9 @@ or
...
+	c.Flags().BoolVarP(&opts.Fzf, "fzf", "F", false, "use fzf to select a taskrun")
```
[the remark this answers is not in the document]
anshulvermapatel: done

pkg/options/describe.go
```diff
@@ -0,0 +1,157 @@
...
+
+	return nil
+}
+
+func (opts *DescribeOptions) FuzzyAsk(resource string, options []string) error {
+	chosencolouring := color.NoColor
+	defer func() {
+		color.NoColor = chosencolouring
+	}()
+	// Remove colors as fuzzyfinder doesn't support it!
```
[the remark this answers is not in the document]
anshulvermapatel: Sure, made the changes.

pkg/options/describe_test.go
```diff
@@ -0,0 +1,156 @@
...
+				test.AssertOutput(t, tp.want, err.Error())
+			} else if err != nil {
+				t.Errorf("unexpected Error")
+			}
+		})
+	}
+}
+
+func TestDescribeOptions_Ask(t *testing.T) {
+
```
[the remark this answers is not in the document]
anshulvermapatel: done

pkg/cmd/clustertask/describe.go
anshulvermapatel: @danielhelfand , this limit is there for `logs` also. There can be lots of `pr(s)` and `tr(s)` , I think we should keep the limit and anyways we are giving an option(`--limit`) to increase it if the user wants. If you think that the default limit should not be there, then I think we should do the same for `logs` also.

pkg/cmd/pipelinerun/describe.go
anshulvermapatel: Sure, will keep that in mind. made the change.

pkg/cmd/task/describe.go
anshulvermapatel: Done, sorry about that.

pkg/options/describe_test.go
anshulvermapatel: will make the changes.

pkg/options/describe_test.go
anshulvermapatel: Sorry :( , removed

pkg/options/describe.go
anshulvermapatel: Done

pkg/cmd/clustertask/describe.go
danielhelfand: Yes, my mistake. I think the user experience around this is a bit odd, but yes you are correct. This can be kept.

pkg/cmd/pipelinerun/describe.go
danielhelfand: This should still be removed. The point of the `if len(args) == 0 { ... }` block is to select the appropriate PipelineRun name. All the other logic within the `if` block is already happening. This should be fixed for all commands.

pkg/options/describe.go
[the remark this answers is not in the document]
danielhelfand: It still says 2019?

pkg/options/describe.go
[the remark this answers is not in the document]
danielhelfand: The comments are still here?

pkg/options/describe.go
danielhelfand: The `*` is still there.

pkg/cmd/clustertask/describe.go
danielhelfand: So yes, I think doing this in the case of PipelineRuns/TaskRuns makes sense for describe. Because, as you say, there are many. However, in the case of resources that are not organized in a meaningful way (i.e. ClusterTasks, Tasks, Pipelines, etc) and are bound to have fewer in a namespace, there should be no limit. With `logs`, it makes sense because those are PipelineRun/TaskRun related. There will be many so it makes sense to limit them. 

I could see later on how a filter option might help here, but I don't see limit as a good approach with non PipelineRun/TaskRun resources.

danielhelfand: So yes, I think doing this in the case of PipelineRuns/TaskRuns makes sense for describe. Because, as you say, there are many. However, in the case of resources that are not organized in a meaningful way (i.e. ClusterTasks, Tasks, Pipelines, etc) and are bound to have fewer in a namespace, there should be no limit. With logs, it makes sense because those are PipelineRun/TaskRun related. There will be many so it makes sense to limit them.

I could see later on how a filter option might help here, but I don't see limit as a good approach with non PipelineRun/TaskRun resources.

Logs have their own `--limit` option already set, so my opinion is it doesn't really fit here.

anshulvermapatel: Okay, I will remove the limit for non- TR/PR resources.

anshulvermapatel:
@danielhelfand , I just marked them done as I did the changes locally but did not push it. I have made all the changes asked and pushed them now. Please have a look and let me know if they are good.

pkg/options/describe_test.go
```diff
@@ -0,0 +1,433 @@
+// Copyright © 2020 The Tekton Authors.
```
danielhelfand, line 1: I am reaching out to other devs on `tkn` who are more familiar with `interactive` testing approaches to see if there would be better approaches than this for testing this feature. I am not fully convinced these are really covering this feature. 

I realize this approach is used with other features, but we might want to rethink it as a viable approach for these kinds of tests.

anshulvermapatel: `/pkg/options/describe_test.go` is based on `/pkg/options/logs_test.go` since Describe Options and Logs Options are similer, I kept the tests for describe.go similer to test present for logs.go go in `options` package.

pkg/cmd/pipelineresource/describe.go
```diff
@@ -167,3 +177,27 @@ func printPipelineResourceDescription(s *cli.Stream, p cli.Params, preName strin
...
 }
+
+func askPipelineResourceName(opts *options.DescribeOptions, p cli.Params) error {
+	//prs, err := prhelper.GetAllPipelineRuns(opts.Params, lOpts, opts.Limit)
+	cliClients, err := p.Clients()
+	if err != nil {
+		return err
+	}
+
+	// the `allPipelineResourceNames( )` func is in delete.go, same location.
```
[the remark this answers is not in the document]
anshulvermapatel: made changes ..

pkg/cmd/pipelineresource/describe.go
```diff
@@ -167,3 +177,27 @@ func printPipelineResourceDescription(s *cli.Stream, p cli.Params, preName strin
 	}
 	return w.Flush()
 }
+
+func askPipelineResourceName(opts *options.DescribeOptions, p cli.Params) error {
+	//prs, err := prhelper.GetAllPipelineRuns(opts.Params, lOpts, opts.Limit)
```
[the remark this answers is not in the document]
anshulvermapatel: I am sorry for and this, I have rebuilt the docs now.

pkg/options/describe_test.go
danielhelfand: Yes, it is my opinion that those should have never been accepted as tests and am not sure why they were. Basically, you can see the code coverage [here](https://github.com/tektoncd/cli/pull/890#issuecomment-618881461) and click on the link for a further break down. The tests aren't covering anything but rather trying to mock the behavior, which isn't really helpful when someone changes the code later on as there will be no precedent for what should still work after a change. So basically we need a testing pattern for interactive testing. I would suggest moving this tests into the unit tests for subcommands (i.e. `/pkg/cmd/<resoure>/describe_test.go` so the logic being added is also covered in describe command code as well. I'll also try and think of a testing approach here.

anshulvermapatel: I think the `/pkg/option/describe.go` is just for asking which resource to describe, so we should only test if the correct resources are been given in options and choosing one is setting the correct resource name in `options.DescribeOptions` . I think the tests should be here only in /pkg/options/ because the test for if a resource is being described properly or the description based tests are in the unit tests of subcommands. Here we are choosing a resource so the tests for the 'choosing' part should be here.
Yes, I agree with the coverage part as tests for FuzzyAsk are not there. We would need to include them.

danielhelfand: Ah, my apologies for the confusion in all this. I see what you are saying about testing `Ask` and think these are valid.

danielhelfand: > Yes, I agree with the coverage part as tests for FuzzyAsk are not there. We would need to include them.

If you do have thoughts around testing `FuzzyAsk`, I think that would be great though. That would be my only other ask for this pr.

danielhelfand: Something to potentially consider is how some of the funcs from `fuzzyfinder` are tested:

https://github.com/ktr0731/go-fuzzyfinder/blob/d856dcab9233fa197ce52a4daa33c4b6dca85248/fuzzyfinder_test.go#L258

anshulvermapatel: `FuzzyAsk` tests are not implemented for `logs` also :P , I don't know why I think it would be challenging for me but I will give it a try.

anshulvermapatel:
> I'm willing to let this go through without unit tests for `--fzf` as this is probably better handled in an `e2e` scenario, which can be followed up on after this.
> 
> /lgtm

Okay :D

chmouel:
I think it's kinda hard to e2e/unittest `--fzf` reliably

/approve

Issue #802, opened 2020-03-13
tkn taskrun delete --keep Doesn't Sort TaskRuns By Start Time Before Deleting


danielhelfand:
# Version and Operating System

**tkn Version:** v0.8.0

# Expected Behavior

`tkn tr delete --all --keep` doesn't [sort TaskRuns by start time](https://github.com/tektoncd/cli/blob/0151eac9175363c77484903289333cb0aad50e07/pkg/cmd/taskrun/delete.go#L147) before deleting them. The description for the command says that it `Keep n least recent number of taskruns when deleting`. As of now, the flag will delete TaskRuns that are sorted by name by default. 

In general, a better deletion filtering method as described in #800 would probably make sense here, but, at the very least, this should be updated since this behavior is not expected by users. 

Sorting should only be applied when `--keep` is specified. The way it is [implemented for PipelineRuns](https://github.com/tektoncd/cli/blob/0151eac9175363c77484903289333cb0aad50e07/pkg/cmd/pipelinerun/delete.go#L148), sorting takes place when `--all` is specified, which is unnecessary. This should be addressed for PipelineRuns as well so it only sorts when `--keep` is specified.

PR #893, opened 2020-04-12
Sort TaskRuns based on StartTime before deletion with keep

waveywaves:
fixes https://github.com/tektoncd/cli/issues/802

When using --keep with `tkn tr delete` TaskRuns should be
sorted based on StartTime, which promotes deletion based
on the Age of the TaskRun. With only the older TaskRuns
deleted and the newer ones kept, as should be the case.

# Changes

# Submitter Checklist

review them:

for more details._

# Release Notes

```
release-note
```

### DIFF
--- a/pkg/cmd/taskrun/delete.go
+++ b/pkg/cmd/taskrun/delete.go
@@ -18,13 +18,15 @@ import (
 	"errors"
 	"fmt"
 
+	trsort "github.com/tektoncd/cli/pkg/taskrun/sort"
+
 	"github.com/spf13/cobra"
 	"github.com/tektoncd/cli/pkg/actions"
 	"github.com/tektoncd/cli/pkg/cli"
 	"github.com/tektoncd/cli/pkg/deleter"
 	"github.com/tektoncd/cli/pkg/options"
 	trlist "github.com/tektoncd/cli/pkg/taskrun/list"
-	validate "github.com/tektoncd/cli/pkg/validate"
+	"github.com/tektoncd/cli/pkg/validate"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	cliopts "k8s.io/cli-runtime/pkg/genericclioptions"
@@ -147,6 +149,7 @@ func allTaskRunNames(cs *cli.Clients, keep int, ns string) ([]string, error) {
 	if err != nil {
 		return nil, err
 	}
+	trsort.SortByStartTime(taskRuns.Items)
 	var names []string
 	var counter = 0
 	for _, tr := range taskRuns.Items {


pkg/cmd/taskrun/delete.go
```diff
@@ -147,6 +147,7 @@ func allTaskRunNames(cs *cli.Clients, keep int, ns string) ([]string, error) {
 	if err != nil {
 		return nil, err
 	}
+	reverse(taskRuns)
```
danielhelfand, line 150: The list of TaskRuns is not sorted by start time, so reversing the list will not work. You need to actually sort the TaskRuns using the [`sortByStartTime`](https://github.com/tektoncd/cli/blob/f2b65dd2be25975ec91c702811f1d101dc052d64/pkg/cmd/taskrun/list.go#L207) func. `reverse` should not be needed at all as you can just access the last element of the slice.

waveywaves: I had tested this incorrectly which made me opt for the reverse. I have reverted back to the initial sorting I did in https://github.com/tektoncd/cli/commit/2778b104684e3e2161cdd642b10020b431feb30a

pkg/cmd/taskrun/delete.go
```diff
@@ -17,14 +17,15 @@ package taskrun
 import (
 	"errors"
 	"fmt"
+	taskrun "github.com/tektoncd/cli/pkg/taskrun/sort"
```
danielhelfand, line 20: Can you rename this to `trsort` to keep it consistent with how it's referenced else where?

pkg/cmd/taskrun/delete_test.go
```diff
@@ -18,7 +18,9 @@ import (
 	"io"
```
danielhelfand, line 18: These edits to `delete_test.go` aren't really testing anything. We have sufficient testing to see if TaskRun sorting is working, so it's all right to not include explicit unit tests in this case. This would probably be a better candidate for an `e2e` test, but it's all right to hold off on that for now as this feature will likely be changed to address other issues.

pkg/cmd/taskrun/delete.go
waveywaves: Done

pkg/cmd/taskrun/delete_test.go
waveywaves: Done

waveywaves:
@danielhelfand Any final verdict on this ?

danielhelfand:
pull-tekton-cli-integration-tests-0_10 seem to be stuck here.

pkg/cmd/taskrun/delete.go
```diff
@@ -17,14 +17,15 @@ package taskrun
 import (
 	"errors"
 	"fmt"
+	trsort "github.com/tektoncd/cli/pkg/taskrun/sort"
```
danielhelfand, line 20: The linter is complaining about this import. Try running `gofmt -w` on this file to resolve the way your imports are are formatted.

waveywaves:
@danielhelfand Can you PTAL ?

chmouel:
/lgtm
/approve

Thanks 👍🏼🔥

PR #901, opened 2020-04-15
Remove Default Timeout of 1h for Pipeline and Task start

danielhelfand:
Part of #865 

This pull request partially addresses #865. It removes default values for `tkn pipeline start` and `tkn task start` for the `--timeout` option. This will need to eventually also be applied to `tkn clustertask start`, but there needs to be changes made via #793 before that can take place.

Removing the default should be backwards compatible as pipelines will still apply the `1h` default in the event nothing is specified via `tkn`. This will also respect the `default-timeout-minutes` property specified in config-defaults. 

# Submitter Checklist

review them:

# Release Notes

```
Remove 1h default for --timeout for pipeline and task start to allow default-timeout-minutes to be correctly applied
```


chmouel:
nice, thanks @danielhelfand 👍

Issue #800, opened 2020-03-12
Change Behavior of What --keep Saves from Deletion


danielhelfand:
# Version and Operating System

**tkn Version:** v0.8.0

# Expected Behavior

The `--keep` option for `tkn pr ls` and `tkn tr ls` should delete the least recent TaskRuns or PipelineRuns as opposed to the most recent. It is more common to want to keep more recent entries than later ones. 

An alternative suggestion would be for `--keep` to keep by something other than a number. An example being a date range or a variety of filter options. 

Whatever the choice, the deletion approach of `--keep` should be changed from its current deletion approach.

An example is as follows for keeping most recent TaskRuns:

`tkn tr ls`:

```
NAME                STARTED          DURATION     STATUS               
sleep1s-run-qkh5n   31 seconds ago   26 seconds   Succeeded   
sleep1s-run-t7zxw   32 seconds ago   27 seconds   Succeeded   
sleep1s-run-8svzx   33 seconds ago   32 seconds   Succeeded   
sleep1s-run-vr6v9   33 seconds ago   31 seconds   Succeeded   
sleep1s-run-87kwl   34 seconds ago   31 seconds   Succeeded   
sleep1s-run-zczw8   35 seconds ago   32 seconds   Succeeded   
sleep1s-run-dhqdl   36 seconds ago   34 seconds   Succeeded   
sleep1s-run-n725h   37 seconds ago   33 seconds   Succeeded   
sleep1s-run-98d2r   38 seconds ago   32 seconds   Succeeded   
sleep1s-run-5rxfg   39 seconds ago   33 seconds   Succeeded
```

`tkn tr delete --all --keep 3 & tkn tr ls` (Most recent are saved):

```
NAME                STARTED          DURATION     STATUS               
sleep1s-run-qkh5n   31 seconds ago   26 seconds   Succeeded   
sleep1s-run-t7zxw   32 seconds ago   27 seconds   Succeeded   
sleep1s-run-8svzx   33 seconds ago   32 seconds   Succeeded   
```

# Actual Behavior

An example is as follows:

`tkn tr ls`:

```
NAME                STARTED          DURATION     STATUS               
sleep1s-run-qkh5n   31 seconds ago   26 seconds   Succeeded   
sleep1s-run-t7zxw   32 seconds ago   27 seconds   Succeeded   
sleep1s-run-8svzx   33 seconds ago   32 seconds   Succeeded   
[...]

PR #921, opened 2020-04-21
Fix --keep Description and Remove Need to use with --all Flag

danielhelfand:
Slight cleanup of `--keep` flag for `tkn pr delete`/`tkn tr delete`. With this change, users will no longer need to use `--all` with `--keep` but still can. This pull request will set `--all` to true when `--keep` is greater than 0. This is needed since `--keep` relies on funcs used with `--all`. 

This pull request also corrects the description to clearly communicate what PipelineRuns/TaskRuns are kept when using `--keep`. #800 is incorrect in stating that most recent PipelineRuns/TaskRuns are not kept. I believe the original issue was opened because of #802, which caused confusion cause of what was kept. 

This pull request does not solve #800 as the behavior should probably be changed later on, but it does clean up certain user experience aspects of `--keep`.

# Submitter Checklist

review them:

# Release Notes

```
Do not require use of --all with --keep flag and correct description of keep option to say that most recent pipelinerun/taskruns are kept
```

### DIFF
--- a/pkg/cmd/taskrun/delete_test.go
+++ b/pkg/cmd/taskrun/delete_test.go
@@ -206,7 +206,7 @@ func TestTaskRunDelete(t *testing.T) {
 		},
 		{
 			name:        "Delete all keeping 2",
-			command:     []string{"delete", "--all", "-f", "--keep", "2", "-n", "ns"},
+			command:     []string{"delete", "-f", "--keep", "2", "-n", "ns"},
 			dynamic:     seeds[4].dynamicClient,
 			input:       seeds[4].pipelineClient,
 			inputStream: nil,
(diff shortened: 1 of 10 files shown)


danielhelfand:
/hold

Slight test issue I would like to correct

pkg/cmd/taskrun/delete_test.go
```diff
@@ -206,16 +206,25 @@ func TestTaskRunDelete(t *testing.T) {
...
+			command:     []string{"delete", "-f", "--keep", "2", "-n", "ns"},
 			dynamic:     seeds[4].dynamicClient,
 			input:       seeds[4].pipelineClient,
 			inputStream: nil,
 			wantError:   false,
 			want:        "All but 2 TaskRuns deleted in namespace \"ns\"\n",
 		},
+		{
+			name:        "Delete all keeping 1 with --all flag",
+			command:     []string{"delete", "-f", "--keep", "1", "-n", "ns"},
```
danielhelfand, line 218: Should have `--all` flag to verify this won't be a breaking change.

danielhelfand: Addressed

PR #926, opened 2020-04-22
Fix pipeline version not working

### LINKED ISSUE #925: tkn version shows the pipeline version as unknown when pipeline is installed in the namespace other than tekton-pipelines
# Version and Operating System

**tkn Version:**
Latest

**Operating System:** 
All

# Expected Behavior
`tkn version` should show pipeline version even if the pipeline is installed in any namespace

# Actual Behavior
`tkn version` shows pipeline version as unknown when pipeline is installed in namespace `tekton-pipelines`

1. Install pipeline in any namespace other than 'tekton-pipelines`
2. Run `tkn version` command and observe the `pipeline version`

# Additional Info
Command logs
```
$ tkn version
Client version: dev
Pipeline version: unknown
```

piyush-garg:
This will fix the issue of pipeline version showing
unknown when installed in other namespace
Now it will check tekton-pipelines and openshift-pipelines
namespace and if not found then will check all
namespaces for deployment

Fix #925

# Submitter Checklist

review them:

for more details._

# Release Notes

```
Fix pipeline version not showing when installed in other namespace
```

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -1,3 +1,17 @@
+// Copyright © 2020 The Tekton Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package version
 
 import (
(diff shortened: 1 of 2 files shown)


chmouel:
I think that's going to be crazy slow, and it will fail if you don't have right to list the avail NS on cluster,  when /most/ of the times  pipeline will be installed in the default namespace.

Can we start by trying in default namespace first and if we can't find it there list the namespaces....

Another way us just to provide a --pipeline-namespace switch and just let the user do this.

/hold

vdemeester:
I think, we could look into `tekton-pipelines` and `openshift-pipelines` namespace (if present) for 0.9.0.
For the next releases, we need to set thoses lables / notations on something easier to query, the CRD themselves :angel:

piyush-garg:
If a user doesn't have access for it, it will show as `Pipeline version: unknown`

vdemeester:
> If a user doesn't have access for it, it will show as `Pipeline version: unknown`

access for ?

chmouel:
@vdemeester that sounds better to set thos annotation on the CRD

@piyush-garg yes that would be a defect then... If you have access to the tekton-pipeline namespace only it will currently works because you can access it directly, but if you don't have namespace listing it will not even try to go to tekton-pipeline namespace.....

vdemeester:
I am fine with *hardcoding* tekton-pipelines and openshift-pipelines for 0.9.0 :stuck_out_tongue:

pkg/version/version.go
```diff
@@ -5,25 +5,26 @@ import (
...
 	listOptions := metav1.ListOptions{
 		LabelSelector: "app.kubernetes.io/component=controller,app.kubernetes.io/name=tekton-pipelines",
 	}
 
-	deployments, err := c.Kube.AppsV1().Deployments(pipelineNamespace).List(listOptions)
-	if err != nil {
-		return "", err
+	var deployments []v1.Deployment
+	for _, namespace := range namespaces {
+		d, _ := c.Kube.AppsV1().Deployments(namespace).List(listOptions)
```
vdemeester, line 23: We should at least log the error :wink:

vdemeester: it's a nit, not sure if it make sense really (to log the error)

chmouel: why would you list the deployments? i don't think that change much compared to the previous iteration of this patch, what about first trying the static ones and then listing deployment as last course ?

piyush-garg: PR updated as per your comment. Please let me know if any other change required.

pkg/version/version.go
```diff
@@ -1,29 +1,63 @@
...
+	}
+
+	if version == "" {
+		d, _ := c.Kube.AppsV1().Deployments("").List(listOptions)
+		deployments = append(deployments, d.Items...)
+		version = findVersion(deployments)
+	}
+
+	if version == "" {
+		return "", fmt.Errorf("Error getting the tekton pipelines deployment version. Version is unknown")
```
chmouel, line 52: Can we get just a single word, as "unknown", reasoning is that we can script it easily... (and it would be backward compatible as before

piyush-garg: yes this is a function called by versionCmd, there it prints the only version if it receives an error

piyush-garg: Basically this error is ignored there https://github.com/tektoncd/cli/blob/master/pkg/cmd/version/version.go#L56

chmouel: 👍

PR #944, opened 2020-04-28
Add flag to skip check flag in version cmd

piyush-garg:
This will add a ldflag to skip the check
flag in version cmd. By default it will
work as the current behaviour. It will
skip adding the flag if build as passing
the flsg skipCheckFlag var as true

The flag to skip check flag can be passed as
go build -ldflags '-X github.com/tektoncd/cli/pkg/cmd/version.skipCheckFlag=true' ./cmd/tkn

# Submitter Checklist

review them:

for more details._

# Release Notes

### DIFF
--- a/pkg/cmd/version/version.go
+++ b/pkg/cmd/version/version.go
@@ -33,6 +33,9 @@ import (
 // NOTE: use go build -ldflags "-X github.com/tektoncd/cli/pkg/cmd/version.clientVersion=$(git describe)"
 var clientVersion = devVersion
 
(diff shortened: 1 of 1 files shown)


danielhelfand:
Why is this needed? Is this causing an issue with the release?

chmouel:
@danielhelfand this is for distros, so when they compile their packages, they only let the packages being the sources of the update. I think it's a policy somewhere for Debian and Fedora packages (+derivatives)

PR #947, opened 2020-04-29
Improve get pipelines version func

eddycharly:
This PR improves the way the `GetPipelineVersion` func works.

It does so by adding support for standard/recommended [k8s labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/).
The func also looks in all namespaces by default.

This is part of the implementation for this proposal: https://github.com/tektoncd/pipeline/issues/2497

The related PR in pipelines is here: https://github.com/tektoncd/pipeline/pull/2501

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -23,36 +23,45 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	pipelinesControllerSelector    string = "app.kubernetes.io/part-of=tekton-pipelines,app.kubernetes.io/component=controller,app.kubernetes.io/name=controller"
+	oldpipelinesControllerSelector string = "app.kubernetes.io/component=controller,app.kubernetes.io/name=tekton-pipelines"
+)
+
 // GetPipelineVersion Get pipeline version, functions imported from Dashboard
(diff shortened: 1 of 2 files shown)


pkg/version/version.go
```diff
@@ -23,36 +23,45 @@ import (
...
+	if err != nil {
+		return nil, err
 	}
 
-	if version == "" {
-		return "", fmt.Errorf("Error getting the tekton pipelines deployment version. Version is unknown")
+	// NOTE: If the new labels selector returned nothing, try with old labels selector
+	// The old labels selectors are deprecated and should be removed at some point
+	if deployments == nil || len(deployments.Items) == 0 {
+		appSelector := "app.kubernetes.io/component=controller,app.kubernetes.io/name=tekton-pipelines"
```
chmouel, line 57: maybe you can put that as const like we do for `pipelinesControllerSelector` as `oldpipelinesControllerSelector`

vdemeester: Also, this will fail for previous version of tekton wouldn't it ? (`app.kubernetes.io` labels are in 0.12 not before, right ?)

eddycharly: @vdemeester if the new selector returns nothing, we try with the old selector as a fallback

eddycharly: @chmouel done

Issue #799, opened 2020-03-12
Add --no-headers flag for all list commands


ppitonak:
# Version and Operating System

**tkn Version:**
0.8.0

**Operating System:** 
all systems

# Expected Behavior
When using `tkn`, I should be able to list resources without header, i.e. `tkn ct list --no-headers` would produce something like

```
buildah                    23 hours ago
buildah-v0-10-0            23 hours ago
jib-maven                  23 hours ago
kn                         23 hours ago
```

instead of 

```
NAME                       AGE
buildah                    23 hours ago
buildah-v0-10-0            23 hours ago
jib-maven                  23 hours ago
kn                         23 hours ago
```

# Actual Behavior

Header is always printed

# Steps to Reproduce the Problem

1.
2.
3.

# Additional Info

Both `kubectl` and `oc` have this flag
```
--no-headers=false: When using the default or custom-column output format, don't print headers (default print headers).
```

waveywaves:
Tracker for --no-header in different `list` commands
- [ ] TaskRun
- [ ] PipelineRun
- [ ] Task
- [ ] Pipeline
- [ ] PipelineResource
- [ ] TriggerBinding
- [ ] TriggerTemplate
- [ ] EventListener
- [ ] Condition

piyush-garg:
We can add the same for clustertriggerbinding and clustertask also

waveywaves:
@piyush-garg Added ClusterTriggerBinding and ClusterTask to the tracker

ppitonak:
What's the plan for ClusterTriggerBinding, ClusterTask, TriggerBinding, TriggerTemplate, EventListener and Condition?

ppitonak:
@waveywaves do you have estimate when the rest will be implemented?

pratikjagrut:
Tracker for `--no-headers` in different `list` commands
* [x]  TaskRun #834
* [x]  PipelineRun #834
* [x]  Task #840
* [x]  Pipeline #840
* [x]  PipelineResource #971
* [x]  ClusterTriggerBinding #1243 
* [x]  ClusterTask #1244 
* [ ]  TriggerBinding
* [ ]  TriggerTemplate
* [ ]  EventListener
* [ ]  Condition

piyush-garg:
All subtasks done

/close

Issue #785, opened 2020-03-09
Support --all-namespaces flag for list commands


pradeepitm12:
Cli should support `--all-namespaces` for all the `list` subcommands.

Currently, cli `list` subcommands take flag `-n` namespace as an input, which lists the given resource of a given namespace
  `tkn foo list -n bar`

Support to `all-namespaces` flag will help the user to list a particular resource in all the namespaces.
  `tkn task list --all-namespaces`

Adding to this, if `all-namespaces` flag is enabled then cli should show an extra column which tells the namespace too.

waveywaves:
- [x] TaskRun https://github.com/tektoncd/cli/pull/791
- [x] PipelineRun https://github.com/tektoncd/cli/pull/791
- [ ] Task
- [ ] Pipeline
- [ ] PipelineResource
- [ ] TriggerBinding
- [ ] TriggerTemplate
- [ ] EventListener
- [ ] Condition

danielhelfand:
This was accidentally closed via #791 but should remain open until other resources also support this flag.

waveywaves:
https://github.com/tektoncd/cli/pull/840 opened to which covers the requirement for tasks and pipelines

pratikjagrut:
Opened PR #1234 to add --all--namespaces flag in triggerbinding list command.

pratikjagrut:
* [x]  TaskRun #791
* [x]  PipelineRun #791
* [x]  Task #840
* [x]  Pipeline #840
* [x]  PipelineResource #971
* [x]  TriggerBinding #1234 
* [x]  TriggerTemplate #1235 
* [x]  EventListener #1236 
* [ ]  Condition

danielhelfand:
Closing this cause of #1241

PR #971, opened 2020-05-07
Add no-headers and all-namespaces flags for `resource list`

waveywaves:
# Changes

--no-headers doesnt print the headers in this case making it
easier for users to parse through the output.
--all-namespace displays pipelines from all namespaces

wrt https://github.com/tektoncd/cli/issues/799 
https://github.com/tektoncd/cli/issues/785

# Submitter Checklist

review them:

for more details._

# Release Notes

```
release-note
```

### DIFF
--- a/pkg/cmd/pipelineresource/list.go
+++ b/pkg/cmd/pipelineresource/list.go
@@ -17,12 +17,12 @@ package pipelineresource
 import (
 	"fmt"
 	"os"
-	"sort"
 	"strings"
 	"text/tabwriter"
 
 	"github.com/spf13/cobra"
 	"github.com/tektoncd/cli/pkg/cli"
+	pressort "github.com/tektoncd/cli/pkg/pipelineresource/sort"
 	"github.com/tektoncd/cli/pkg/printer"
 	validinput "github.com/tektoncd/cli/pkg/validate"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
@@ -37,7 +37,9 @@ const (
 )
 
 type ListOptions struct {
-	Type string
+	Type          string
+	NoHeaders     bool
+	AllNamespaces bool
 }
 
 func listCommand(p cli.Params) *cobra.Command {
@@ -61,8 +63,10 @@ func listCommand(p cli.Params) *cobra.Command {
 		Args: cobra.MinimumNArgs(0),
 		RunE: func(cmd *cobra.Command, args []string) error {
 
-			if err := validinput.NamespaceExists(p); err != nil {
-				return err
+			if !opts.AllNamespaces {
+				if err := validinput.NamespaceExists(p); err != nil {
+					return err
+				}
 			}
 
 			cs, err := p.Clients()
@@ -81,13 +85,23 @@ func listCommand(p cli.Params) *cobra.Command {
 				return fmt.Errorf("failed to list pipelineresources. Invalid resource type %s", opts.Type)
 			}
 
-			pres, err := list(cs.Resource, p.Namespace(), opts.Type)
+			namespace := p.Namespace()
+			if opts.AllNamespaces {
+				namespace = ""
+			}
+
+			pres, err := list(cs.Resource, namespace, opts.Type)
 			stream := &cli.Stream{
 				Out: cmd.OutOrStdout(),
 				Err: cmd.OutOrStderr(),
 			}
+
 			if err != nil {
-				fmt.Fprintf(os.Stderr, "Failed to list pipelineresources from %s namespace \n", p.Namespace())
+				ns := namespace
+				if opts.AllNamespaces {
+					ns = "all"
+				}
+				fmt.Fprintf(os.Stderr, "Failed to list pipelineresources from %s namespace(s) \n", ns)
 				return err
 			}
 
@@ -101,7 +115,7 @@ func listCommand(p cli.Params) *cobra.Command {
 				return printer.PrintObject(stream.Out, pres, f)
 			}
 
-			err = printFormatted(stream, pres)
+			err = printFormatted(stream, pres, opts.NoHeaders, opts.AllNamespaces)
 			if err != nil {
 				fmt.Fprint(os.Stderr, "Failed to print Pipelineresources \n")
 				return err
@@ -112,6 +126,8 @@ func listCommand(p cli.Params) *cobra.Command {
 
 	f.AddFlags(cmd)
 	cmd.Flags().StringVarP(&opts.Type, "type", "t", "", "Pipeline resource type")
+	cmd.Flags().BoolVarP(&opts.NoHeaders, "no-headers", "", opts.NoHeaders, "do not print column headers with output (default print column headers with output)")
+	cmd.Flags().BoolVarP(&opts.AllNamespaces, "all-namespaces", "A", opts.AllNamespaces, "list pipeline resources from all namespaces")
 
 	return cmd
 }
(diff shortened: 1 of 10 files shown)


pkg/cmd/pipelineresource/list.go
```diff
@@ -81,13 +83,17 @@ func listCommand(p cli.Params) *cobra.Command {
...
+				namespace = ""
+			}
+			pres, err := list(cs.Resource, namespace, opts.Type)
 			stream := &cli.Stream{
 				Out: cmd.OutOrStdout(),
 				Err: cmd.OutOrStderr(),
 			}
 			if err != nil {
-				fmt.Fprintf(os.Stderr, "Failed to list pipelineresources from %s namespace \n", p.Namespace())
+				fmt.Fprintf(os.Stderr, "Failed to list pipelineresources from %s namespace \n", namespace)
```
piyush-garg, line 96: This will show up with double space in case of all-namespace

pkg/cmd/pipelineresource/list.go
```diff
@@ -61,7 +63,7 @@ func listCommand(p cli.Params) *cobra.Command {
 		Args: cobra.MinimumNArgs(0),
 		RunE: func(cmd *cobra.Command, args []string) error {
 
-			if err := validinput.NamespaceExists(p); err != nil {
+			if err := validinput.NamespaceExists(p); err != nil && !opts.AllNamespaces {
```
piyush-garg, line 66: We don't even need to check whether namespace exists if `all-namespace` is true

pkg/cmd/pipelineresource/list.go
waveywaves: agreed I will update this accordingly

pkg/cmd/pipelineresource/list.go
waveywaves: We need to, as the test for this needs `-n` 
This was confirmed in the previous PRs where we were adding the --all-namespaces flag.
And when we are getting the output for all namespaces we do not need to check if the particular namespace exists.

danielhelfand: You can do this to avoid checking for the namespace and unit tests should work: 

```go
if !opts.AllNamespaces {
    if err := validinput.NamespaceExists(p); err != nil {
        return err
    }
}

namespace := p.Namespace()
if opts.AllNamespaces {
    namespace = ""
}
```

pkg/cmd/pipelineresource/list.go
```diff
@@ -81,13 +83,21 @@ func listCommand(p cli.Params) *cobra.Command {
...
 				Out: cmd.OutOrStdout(),
 				Err: cmd.OutOrStderr(),
 			}
 			if err != nil {
-				fmt.Fprintf(os.Stderr, "Failed to list pipelineresources from %s namespace \n", p.Namespace())
+				ns := namespace
+				if opts.AllNamespaces {
+					ns = "all"
+				}
+				fmt.Fprintf(os.Stderr, "Failed to list pipelineresources from %s namespace \n", ns)
```
danielhelfand, line 100:
```suggestion
				fmt.Fprintf(os.Stderr, "Failed to list pipelineresources from %s namespaces \n", ns)
```

pkg/cmd/pipelineresource/list.go
piyush-garg: I have raised #974 for other commands and removed `-n` in tests also. And I have done the same as daniel mentioned above.

pkg/cmd/pipelineresource/list.go
waveywaves: I think `namespace(s)` would be better. Should I use this instead ?

danielhelfand: Yup, good catch. Makes sense to me.

waveywaves:
@danielhelfand can you PTAL ? I have moved sorting related functions to `pkg/pipelineresource/sort` just like we have it for the other resources

pkg/cmd/pipelineresource/list.go
```diff
@@ -17,12 +17,12 @@ package pipelineresource
 import (
 	"fmt"
 	"os"
-	"sort"
 	"strings"
 	"text/tabwriter"
 
 	"github.com/spf13/cobra"
 	"github.com/tektoncd/cli/pkg/cli"
+	pressort "github.com/tektoncd/cli/pkg/pipelineresource/sort"
```
piyush-garg, line 25: you can remove pressort here

pkg/cmd/pipelineresource/list_test.go
```diff
@@ -123,6 +123,18 @@ func TestPipelineResourceList(t *testing.T) {
 			args:      []string{"list", "-n", "test-ns-1", "-o", "jsonpath={range .items[*]}{.metadata.name}{\"\\n\"}{end}"},
 			wantError: false,
 		},
+		{
+			name:      "No Headers",
+			command:   command(t, pres, ns),
+			args:      []string{"list", "-n", "test-ns-1", "--no-headers"},
+			wantError: false,
+		},
```
piyush-garg, line 131: There should be two golden files, as you have added 2 tests, why 3 golden files?

pkg/cmd/pipelineresource/list.go
waveywaves: Why remove this ? It's being used to get the functions SortByNamespace and SortByTimeAndDate. :thinking: 
If you mean that sort should be used instead of pressort then I am just following convention from other resources where sorting libraries are named as trsort pr sort and such.

pkg/cmd/pipelineresource/list_test.go
waveywaves: Oh I missed that. Let me update it.

pkg/cmd/pipelineresource/list.go
piyush-garg: alias is to differentiate when two packages are imported of same name, so just asked to remove the alias `presort`

pkg/pipelineresource/sort/by_namespace.go
```diff
@@ -0,0 +1,48 @@
...
+func (pres byNamespace) Less(i, j int) bool {
+	var lt, eq bool
+	if lt, eq = pres.compareNamespace(pres[i].Namespace, pres[j].Namespace); eq {
+		if pres[j].Spec.Type < pres[i].Spec.Type {
+			return false
+		}
+		if pres[j].Spec.Type > pres[i].Spec.Type {
+			return true
+		}
+		return pres[j].Name > pres[i].Name
```
danielhelfand, line 45: Please add unit tests to cover these lines.

waveywaves: I have updated the unit tests.

waveywaves:
Can you PTAL @piyush-garg @pradeepitm12

pkg/cmd/pipelineresource/list.go
```diff
@@ -17,12 +17,12 @@ package pipelineresource
 import (
 	"fmt"
 	"os"
-	"sort"
 	"strings"
 	"text/tabwriter"
 
 	"github.com/spf13/cobra"
 	"github.com/tektoncd/cli/pkg/cli"
+	"github.com/tektoncd/cli/pkg/pipelineresource/sort"
```
danielhelfand, line 25: Run `gofmt -w` on this file and push up resulting change to resolve the build failure.

pkg/cmd/pipelineresource/list.go
danielhelfand, line 25: Why was `pressort` added back?

waveywaves: @danielhelfand It had to be, the goimports was making the alias default to `pipelineresource` and `sort` was not use as the function names in sort were in use in other places. If you check `describe` or `list` commands for `pipelinerun` or `taskrun` similar format for the alias is used for the respective sort pkgs being `prsort` and `trsort`. Following the convention, I have had to move it back to `pressort` (Pipeline resource sort).

PR #992, opened 2020-05-22
American spelling of `color`

jjasghar:
# Changes

- The Kubernetes ecosystem and community have standardized on the
American spelling of words. This commit converts to `colour` to `color`
to be consistent with the community at large.

Signed-off-by: JJ Asghar <jjasghar@gmail.com>

# Submitter Checklist

review them:

for more details._

# Release Notes

```
Changed the spelling of colour to color.
```

### DIFF
--- a/docs/cmd/tkn_clustertask.md
+++ b/docs/cmd/tkn_clustertask.md
@@ -14,7 +14,7 @@ Manage clustertasks
   -c, --context string      name of the kubeconfig context to use (default: kubectl config current-context)
   -h, --help                help for clustertask
   -k, --kubeconfig string   kubectl config file (default: $HOME/.kube/config)
-  -C, --nocolor            disable coloring (default: false)
+  -C, --nocolor             disable coloring (default: false)
 ```
 
 ### SEE ALSO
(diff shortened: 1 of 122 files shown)


danielhelfand:
/ok-to-test

Thanks @jjasghar! No issues with this on my end, but might be worth a discussion with some others before proceeding.

vdemeester:
@jjasghar thanks for the PR :wink: I have no issue with it either, we can talk about it during today's WG, but if kubernetes went on with it, I'm cool following :wink:

vdemeester:
Looking around (kubernetes toolings, `ls`, `rg`, `fd`, `grep`, …), `color` is pretty much the standard used, it would be nice if we follow the same. But we should also try not to break our users.

So, I think we could do the following : 
- keep `--colour` (without the short flag) around and mark it as deprecated (to be removed for 1.0)
- add `--color` (with `-c`) and handle when both `--color` and `--colour` are used (just in case). `-c`/`--color` would override `--colour` value, but `--colour` needs to be taken into account if it's the only one used
- add the deprecation notice to the release note
- […]
- Remove `--colour` for `tkn` 1.0

@chmouel @danielhelfand @piyush-garg wdyt ?

Issue #1029, opened 2020-06-19
--use-param-defaults Option for tkn task start


danielhelfand:
As originally introduced in #889 for `tkn pipeline start`, the `--use-param-defaults` option allows users to avoid specifying values via interactive prompts if they want to use param default values. 

This option should also be available for `tkn task start` now that it features interactive prompts for starting a Task. The design should follow the same pattern outlined in #889. 

This should also be available for `tkn clustertask start` eventually after #1013 is closed. 

/kind feature

PR #1036, opened 2020-06-29
Add --use-param-defaults option for tkn task start

Divyansh42:
# Changes

When we start a task using command `tkn task start`, it prompts for asking the parameter values even if the default values are already set in the task configuration and if we just press enter it uses the default values and creates a TaskRun.

To avoid the interactive prompt `--use-params-default` option can be used which will take the `params` value from the default value provided and created the TaskRun.

If default value is `not` provided for any `param` then an interactive prompt will start for the params in which there is no `default`
value.

Fixes part of #1029 

Signed-off-by: Divyansh42 <diagrawa@redhat.com>

# Submitter Checklist

review them:

for more details._

# Release Notes

```release-note
Added  `--use-param-defaults` option for `tkn task start` command.
```


pradeepitm12:
@Divyansh42 Thanks for adding all the tests
/lgtm

pkg/cmd/task/start_test.go
```diff
@@ -3962,6 +3962,86 @@ func TestTaskStart_ExecuteCommand(t *testing.T) {
 			wantError:  false,
 			goldenFile: true,
 		},
+		{
+			name: "Dry Run with --use-param-defaults and specified params",
```
danielhelfand, line 3966: So this unit test isn't actually testing `--use-param-defaults` because we need to define a param default for a `task-1` param. This test here is just passing in a param value and setting it for the TaskRun. What would be a nice way to test this is having a default value defined for a param (e.g. `myarg`) and another param that also has a default but is set to a value that overrides the default.

Setting params can be done [here](https://github.com/tektoncd/cli/blob/20570c11ae379bf75aaf06bfa604b81e0bbf0563/pkg/cmd/task/start_test.go#L3882). It should be similar to what was done for [`tkn pipeline start` tests](https://github.com/tektoncd/cli/blob/1db36d84bc1ce714bf8956d6450fdc2609a51173/pkg/cmd/pipeline/start_test.go#L698).

Let me know if I can further clarify anything here.

pkg/options/start.go
```diff
@@ -348,6 +348,32 @@ func (intOpts *InteractiveOpts) TaskWorkspaces(task *v1beta1.Task) error {
 	return nil
 }
 
+// Include the default param values in opts.Params when `--use-param-defaults` is used
```
danielhelfand, line 351: Would it make more sense to define this directly in `task/start.go` since it only is for tasks?

Divyansh42: Yup, makes sense, Thanks for the suggestion :)

danielhelfand:
Thanks for correcting the unit tests @Divyansh42. I think the last thing this needs is an e2e test, which I think could be done as follows:

1. Add a param default to the [Task test data](https://github.com/tektoncd/cli/blob/35aff032c72ab0d45fb3778689333553fbf2e87e/test/resources/read-file.yaml#L22), which should have a default value of `docs/README.md`. 
2. Add this same test defined [here](https://github.com/tektoncd/cli/blob/35aff032c72ab0d45fb3778689333553fbf2e87e/test/e2e/task/start_test.go#L66) but add `--use-param-defaults` after to the [command run](https://github.com/tektoncd/cli/blob/35aff032c72ab0d45fb3778689333553fbf2e87e/test/e2e/task/start_test.go#L67).
3. Change the name of the test to something like "Start TaskRun using tkn task start command with --use-param-defaults"

We can follow up with something like a describe command to verify the param was properly set, or it can be added as part of this pr. Whatever you prefer.

Divyansh42:
@danielhelfand 

> Thanks for correcting the unit tests @Divyansh42. I think the last thing this needs is an e2e test, which I think could be done as follows, but feel free to suggest a better approach:
> 
> 1. Add a param default to the [Task test data](https://github.com/tektoncd/cli/blob/35aff032c72ab0d45fb3778689333553fbf2e87e/test/resources/read-file.yaml#L22), which should have a default value of `docs/README.md`.
> 2. Add this same test defined [here](https://github.com/tektoncd/cli/blob/35aff032c72ab0d45fb3778689333553fbf2e87e/test/e2e/task/start_test.go#L66) but add `--use-param-defaults` after to the [command run](https://github.com/tektoncd/cli/blob/35aff032c72ab0d45fb3778689333553fbf2e87e/test/e2e/task/start_test.go#L67).
> 3. Change the name of the test to something like "Start TaskRun using tkn task start command with --use-param-defaults"
> 
> We can follow up with something like a describe command to verify the param was properly set, or it can be added as part of this pr. Whatever you prefer.

@danielhelfand added e2e test. Please have a look :)

test/e2e/task/start_test.go
```diff
@@ -83,12 +83,30 @@ Waiting for logs to be available...
...
 		res := tkn.Run("taskrun", "list")
 		expected := e2e.ListAllTaskRunsOutput(t, c, false, map[int]interface{}{
 			0: &e2e.TaskRunData{
 				Name:   "read-task-run-",
-				Status: "Succeeded",
+				Status: "",
+			},
+			1: &e2e.TaskRunData{
+				Name:   "read-task-run-",
+				Status: "",
```
danielhelfand, line 109: What is the reason for removing Status?

danielhelfand: There is a function used in e2e testing for pipelines that waits until the TaskRun finishes:

https://github.com/tektoncd/pipeline/blob/2e737622b576c373894aa4ac2d6f83dc46096cb3/test/v1alpha1/wait.go#L72

Example of using it:

https://github.com/tektoncd/pipeline/blob/89077a45f003f47287a5ea948cda66f506bf5932/test/taskrun_test.go#L75

Divyansh42: Thanks @danielhelfand for the suggestion. :)
Added the same.

piyush-garg:
Just a thought on this, like when we want to use param default in pipelinerun or taskrun, we just don't specify them and pipeline controller automatically picks the default value. Similarly here, I think we don't need to add params section in taskrun if the `--use-param-defaults` is used, and the controller will do the work.

My thought was if we use the `--use-param-defaults` flag we will just skip the params sections in creating taskruns :D

If there are any cons to this approach, then sorry I did not think of that. cc @Divyansh42 @vdemeester @danielhelfand

pkg/cmd/task/start.go
```diff
@@ -531,6 +537,39 @@ func (opt *startOptions) getInputs() error {
 		opt.Workspaces = append(opt.Workspaces, intOpts.Workspaces...)
 	}
 
+	if opt.UseParamDefaults && !opt.Last && opt.UseTaskRun == "" {
+		if err := opt.includeDefaultParams(opt.task); err != nil {
```
vdemeester, line 541: This shouldn't be required, the controller (server-side) will do the work for us if we don't send those parameters.

piyush-garg: Yes exactly, and in that case, we need to remove this from pipeline start also.

danielhelfand: Yes, good catch. This was missed the first time around adding this with pipeline start.

Divyansh42:
Cleaned up the code and added the functionality, if default param is `not` provided then an interactive prompt will start to take input for value of the params.

/cc @piyush-garg @danielhelfand @vdemeester

test/e2e/task/start_test.go
```diff
@@ -83,13 +84,83 @@ Waiting for logs to be available...
...
+
+		if err := e2e.WaitForTaskRunState(c, taskRun1GeneratedName, e2e.TaskRunSucceed(taskRun1GeneratedName), "TaskRunSucceed"); err != nil {
+			t.Errorf("Error waiting for TaskRun to Succeed: %s", err)
+		}
+
+		if err := e2e.WaitForTaskRunState(c, taskRun2GeneratedName, e2e.TaskRunSucceed(taskRun2GeneratedName), "TaskRunSucceed"); err != nil {
+			t.Errorf("Error waiting for TaskRun to Succeed: %s", err)
+		}
+
+		if err := e2e.WaitForTaskRunState(c, taskRun2GeneratedName, e2e.TaskRunSucceed(taskRun3GeneratedName), "TaskRunSucceed"); err != nil {
```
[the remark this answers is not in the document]
Divyansh42: @danielhelfand previously I tried to show  all the params in the describe command for taskrun (whether default value for param is provided or not),  but now I have modified the code to show only `params` which do not have `default` value defined aka params which we provide in the interactive prompt. 

Done this change based on the suggestion from 
@piyush-garg 
>  I think we don't need to add params section in taskrun if the --use-param-defaults is used, and the controller will do the work.

and from @vdemeester 
> So indeed, we don't really need to "add" the default values when building the Run object stuck_out_tongue_closed_eyes

@piyush-garg @vdemeester Please correct me if I understood anything different :sweat_smile:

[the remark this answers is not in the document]
danielhelfand: I understand the reasoning for not specifying default values on the TaskRun object, but I guess it has an unintended effect on `tkn taskrun describe` since the command is expecting the TaskRun to have the param values specified under the TaskRun spec. If param values, whether default or not, are not on TaskRun spec, then the issue is that `tkn tr desc` doesn't show param values specified at run time.

I think your implementation is fine, but we should not introduce something that will break other commands. We may need to look into refactoring taskrun/pipelinerun describe commands to expect that runtime values of params can appear elsewhere. Or pipelines should potentially also apply the param values server side on the spec of the runtime objects.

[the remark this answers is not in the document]
vdemeester: > I think your implementation is fine, but we should not introduce something that will break other commands. We may need to look into refactoring taskrun/pipelinerun describe commands to expect that runtime values of params can appear elsewhere. Or pipelines should potentially also apply the param values server side on the spec of the runtime objects.

So, I am fine *breaking* the `tkn taskrun desc` command over this, mainly because if the user creates the `TaskRun` himself (relying on default parameters too), it will show the same way as this PR makes it — so although it changes *how* `TaskRun` are sent, it's consistent with what happens when `tkn` did not start the `TaskRun`. 

We need an issue to track work on "refining" the `desc` command but I don't think we should block this PR for this.

vdemeester:
> When using `--use-param-defaults` with `./tkn p start`, params do show up with `./tkn pr desc`, which appears to be because pipeline start adds default values as params to the PipelineRun object.
> 
> This is perhaps related to how Pipelines applies the default value, which appears to be under `.Status.TaskRunStatusFields.TaskSpec.Params`. The current TaskRun/PipelineRun describe commands are expecting the value to be under `.Spec.Params` for the TaskRun or PipelineRun.
> 
> Perhaps there should be a redo as far as having describe commands collect this information from the status instead?

Right, we may want to update our `describe` UI to show more information about that (aka user provided params vs "default"/controller provided params), but this is orthogonal to this PR though. As commented about, this behaviour is, imho, consistent with what happens when the `TaskRun` didn't get created by `tkn`.

danielhelfand:
Sure, I am comfortable letting this through despite the issue with describe commands. I will raise a separate issue outside of this pr as a feature request to support controller added params.

Issue #1000, opened 2020-05-27
interactive pipeline/task run fails with (-f) option


praveen4g0:
# Expected Behavior
```
./tkn` task start -f https://raw.githubusercontent.com/tektoncd/catalog/v1beta1/git/git-clone.yaml
? Value for param `url` of type `string`? https://github.com/kelseyhightower/nocode
? Value for param `revision` of type `string`? (Default is `master`) 1.0.0
? Value for param `refspec` of type ``? (Default is ``) 
? Value for param `submodules` of type `string`? (Default is `true`) true
? Value for param `depth` of type `string`? (Default is `1`) 0
? Value for param `sslVerify` of type `string`? (Default is `true`) false
? Value for param `subdirectory` of type `string`? (Default is ``) hellomoto
? Value for param `deleteExisting` of type `string`? (Default is `false`) true
Please give specifications for the workspace: output 
? Name for the workspace : output
? Value of the Sub Path :  
?  Type of the Workspace : emptyDiry/
? Type of EmtpyDir : 
Taskrun started: git-clone-run-rbgwg
```

# Expected
Should throw validation errors if any while starting task/pipeline

# Actual
But I am seeing below error, when see logs
```./tkn taskrun logs git-clone-run-rbgwg -f -n default
Error: task Task 0 has failed: invalid input params: param types don't match the user-specified type: [refspec]
```

# Additional Info
 **workaround:** if we add type explicitly it works as expected, so I guess we need have validation check before starting pipeline or task run

anshulvermapatel:
I can work on this.
/assign

anshulvermapatel:
/reopen
keeping this one open till the PR gets merged.

PR #1069, opened 2020-07-13
Gives a validation error if any `param` does not have a type defined in the task/pipeline definition

anshulvermapatel:
Fixes [1000](https://github.com/tektoncd/cli/issues/1000)
Interactive pipeline/task start fail when `-f` option is used and a `param` in the definition does not have the `type` defined.
This fix gives a validation error if any `param` does not have a type defined.

# Changes

# Submitter Checklist

review them:

for more details._

# Release Notes


anshulvermapatel:
@danielhelfand , Yes, good point on clustertask that is also should be included.
Secondly talking about the type being a required property as anyways when a pipeline/task is run, it fails at it does not recognize the type and it fails and the message we get in the pipeline logs. This is to validate that before it actually creates a PR/TR out of it which will eventually fail if not validated.
I think we should compare the type with the type available and throw an error if an unknown type is mentioned in the definition. I wrongly just compared it with empty string.
What do you say?

piyush-garg:
I guess the of the param type is not defined, `string` is the default type.

anshulvermapatel:
@piyush-garg , As far as I have seen, when a parameter does not have a type defined, the TR/PR fails with error message like -
~~~
$ ./tkn-linux-amd64 pr logs tutorial-pipeline-run-q2tz5
PipelineRun tekton/tutorial-pipeline-run-q2tz5 parameters have mismatching types with Pipeline tekton/tutorial-pipeline-run-q2tz5's parameters: parameters have inconsistent types : [pathToDockerFile]
$ ./tkn-linux-amd64 tr logs git-clone-run-48ngw
task Task 0 has failed: invalid input params for task git-clone-run-48ngw: param types don't match the user-specified type: [refspec]
~~~
~~~
  params:
    - name: refspec
      description: (optional) git refspec to fetch before checking out revision
      default: ""
    - name: submodules
      description: defines if the resource should initialize and fetch the submodules
      type: string
      default: "true"
~~~

danielhelfand:
> @piyush-garg , As far as I have seen, when a parameter does not have a type defined, the TR/PR fails 

Right, I think my question is why it is required by `tkn`? As part of the solution to this issue, I think it would be worthwhile to not make type not required when using the `--filename` option with start.

pkg/cmd/pipeline/start.go
```diff
@@ -765,6 +765,15 @@ func parsePipeline(taskLocation string, p cli.Params) (*v1beta1.Pipeline, error)
...
 	}
+
+	for _, param := range pipeline.Spec.Params {
+		if param.Type == "" {
+			err := fmt.Errorf("The param '%s' does not have a type", param.Name)
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
```
danielhelfand, line 776: Wondering if all errors for not having a `type` specified should be displayed at once instead of returning the error on first failure. So if two params are specified without types, tkn should show both errors for both params.

pkg/cmd/pipeline/start.go
```diff
@@ -765,6 +765,15 @@ func parsePipeline(taskLocation string, p cli.Params) (*v1beta1.Pipeline, error)
 	if err := yaml.UnmarshalStrict(b, &pipeline); err != nil {
 		return nil, err
 	}
+
+	for _, param := range pipeline.Spec.Params {
+		if param.Type == "" {
+			err := fmt.Errorf("The param '%s' does not have a type", param.Name)
```
danielhelfand, line 771:
```suggestion
			err := fmt.Errorf("param %s does not have a type", param.Name)
```

pkg/cmd/task/start.go
```diff
@@ -214,6 +214,14 @@ func parseTask(taskLocation string, p cli.Params) (*v1beta1.Task, error) {
 	if err := yaml.UnmarshalStrict(b, &task); err != nil {
 		return nil, err
 	}
+	for _, param := range task.Spec.Params {
```
danielhelfand, line 217: Same comments as above for pipeline.

pkg/cmd/pipeline/start.go
anshulvermapatel: @danielhelfand , Yes, and what I am also thinking is that currently I am comparing this with an empty string, but it may be possible that the type is invalid(other than string/array), I think in that case also we should give an error for each param of course.

pkg/cmd/task/start.go
piyush-garg, line 217: Is not better to have this common function in params package, and use it. We can then use for pipeline, task and clustertask

pkg/cmd/pipeline/start.go
piyush-garg: https://github.com/tektoncd/pipeline/blob/master/pkg/apis/pipeline/v1beta1/param_types.go#L76

There are the Param types, you can check if the value is one of them or not. If not then show error and I think it is better to show error for all in once as daniel suggested.

pkg/cmd/task/start.go
anshulvermapatel: @piyush-garg , yes, correct, I will write a common function in params package. `-f` option for clustertask has not been implemented yet, so I am doing this currently for pipeline and task. Or may be I can open a new PR to implement `-f` flag for clustertask and one that gets merged we proceed with this one, what do you say?

piyush-garg:
/remove-lifecycle rotten

@anshulvermapatel Can you please fix the tests

anshulvermapatel:
> /remove-lifecycle rotten
> 
> @anshulvermapatel Can you please fix the tests

Hi @piyush-garg I am sorry for the long delay. I will check on the test today.

anshulvermapatel:
@piyush-garg , Hi, the tests are passed, can you please check and merge this one.

pradeepitm12 (approved):

PR #1092, opened 2020-07-31
Modify tkn version to accept namespace as ldflag and flag

savitaashture:
# Changes
**Issue:** Authenticated user(non-admin) cannot get the `tkn version` info because of permission issue as tkn version generally looks pipeline and triggers controller deployment across the cluster(all-namespaces) and non-admin user generally doesn't have permission to view on all namespaces.

So to satisfy admin and non-admin requirements to get the `tkn-version` come up with the below solution from WG meeting
1. Provide namespace as `ldflag` during `tkn` compilation
2. User can specify namespace as flag `tkn version -n {namespace}`
3. default namespace will be `tekton-pipelines` if no `ldflag` and `-n` flag specified

**Output:**
1. Specified namespace in `ldflag` is `tekton-pipelines1` and controllers are running in `tekton-pipelines`
```
$ GOOS=linux GOARCH=amd64 go build -mod=vendor -ldflags "-X github.com/tektoncd/cli/pkg/cmd/version.namespace=tekton-pipelines1"
$ tkn version
Client version: dev
Pipeline version: unknown, pipeline controller may be installed in another namespace please use tkn version -n {namespace}
Triggers version: unknown, triggers controller may be installed in another namespace please use tkn version -n {namespace}
Dashboard version: unknown, dashboard controller may be installed in another namespace please use tkn version -n {namespace}

```

2. Specified correct namespace where controllers are running using `flag`
```
$ tkn version -n tekton-pipelines
Client version: dev
Pipeline version: v0.14.2
Triggers version: devel
Dashboard version: v0.9.0

```
3. Built tkn binary from master code
By default `tkn version` looks Deployment in `tekton-pipeline` or `openshift-pipelines` namespace 
```
$ go build
$ tkn version
Client version: dev
Pipeline version: v0.14.2
Triggers version: devel
Dashboard version: v0.9.0
```

```
$ tkn version --help
Prints version information

Usage:
tkn version [flags]

Flags:
  -c, --check              check if a newer version is available
  -h, --help               help for version
  -n, --namespace string   namespace to check installed controller version

```

# Submitter Checklist

review them:

for more details._

# Release Notes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,9 @@
 YAML_FILES := $(shell find . -type f -regex ".*y[a]ml" -print)
 
+ifneq ($(NAMESPACE),)
+	NAMESPACELDFLAG := -X github.com/tektoncd/cli/pkg/cmd/version.namespace=$(NAMESPACE)
+endif
+
 ifneq ($(SKIP_CHECK_FLAG),)
 	SKIPLDFLAG := -X github.com/tektoncd/cli/pkg/cmd/version.skipCheckFlag=$(SKIP_CHECK_FLAG)
 endif
@@ -12,9 +16,13 @@ ifneq ($(RELEASE_VERSION),)
 	VERSIONLDFLAG := -X github.com/tektoncd/cli/pkg/cmd/version.clientVersion=$(RELEASE_VERSION)
 endif
 
-FLAGS := $(VERSIONLDFLAG)$(SKIPLDFLAG)
+FLAGS := $(VERSIONLDFLAG)
 ifneq ($(SKIPLDFLAG),)
-	FLAGS := $(VERSIONLDFLAG) $(SKIPLDFLAG)
+	FLAGS := $(FLAGS) $(SKIPLDFLAG)
+endif
+
+ifneq ($(NAMESPACELDFLAG),)
+	FLAGS := $(FLAGS) $(NAMESPACELDFLAG)
 endif
 
 ifneq ($(FLAGS),)
--- a/pkg/cmd/version/version.go
+++ b/pkg/cmd/version/version.go
@@ -30,15 +30,18 @@ import (
 	"github.com/tektoncd/cli/pkg/version"
 )
 
-// NOTE: use go build -ldflags "-X github.com/tektoncd/cli/pkg/cmd/version.clientVersion=$(git describe)"
-var clientVersion = devVersion
-
-// flag to skip check flag in version cmd
-var skipCheckFlag = "false"
-
 const (
 	devVersion       = "dev"
 	latestReleaseURL = "https://api.github.com/repos/tektoncd/cli/releases/latest"
+)
+
+var (
+	// NOTE: use go build -ldflags "-X github.com/tektoncd/cli/pkg/cmd/version.clientVersion=$(git describe)"
+	clientVersion = devVersion
+	// flag to skip check flag in version cmd
+	skipCheckFlag = "false"
+	// NOTE: use go build -ldflags "-X github.com/tektoncd/cli/pkg/cmd/version.namespace=tekton-pipelines"
+	namespace string
 )
 
 // Command returns version command
@@ -56,19 +59,23 @@ func Command(p cli.Params) *cobra.Command {
 
 			cs, err := p.Clients()
 			if err == nil {
-				pipelineVersion, _ := version.GetPipelineVersion(cs)
+				pipelineVersion, _ := version.GetPipelineVersion(cs, namespace)
 				if pipelineVersion == "" {
-					pipelineVersion = "unknown"
+					pipelineVersion = "unknown, " +
+						"pipeline controller may be installed in another namespace please use tkn version -n {namespace}"
 				}
+
 				fmt.Fprintf(cmd.OutOrStdout(), "Pipeline version: %s\n", pipelineVersion)
-				triggersVersion, _ := version.GetTriggerVersion(cs)
+				triggersVersion, _ := version.GetTriggerVersion(cs, namespace)
 				if triggersVersion == "" {
-					triggersVersion = "unknown"
+					triggersVersion = "unknown, " +
+						"triggers controller may be installed in another namespace please use tkn version -n {namespace}"
 				}
 				fmt.Fprintf(cmd.OutOrStdout(), "Triggers version: %s\n", triggersVersion)
-				dashboardVersion, _ := version.GetDashboardVersion(cs)
+				dashboardVersion, _ := version.GetDashboardVersion(cs, namespace)
 				if dashboardVersion == "" {
-					dashboardVersion = "unknown"
+					dashboardVersion = "unknown, " +
+						"dashboard controller may be installed in another namespace please use tkn version -n {namespace}"
 				}
 				fmt.Fprintf(cmd.OutOrStdout(), "Dashboard version: %s\n", dashboardVersion)
 			}
@@ -83,6 +90,9 @@ func Command(p cli.Params) *cobra.Command {
 			return err
 		},
 	}
+
+	cmd.Flags().StringVarP(&namespace, "namespace", "n", namespace,
+		"namespace to check installed controller version")
 
 	if skipCheckFlag != "true" {
 		cmd.Flags().BoolVarP(&check, "check", "c", false, "check if a newer version is available")
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -33,8 +33,8 @@ const (
 )
 
 // GetPipelineVersion Get pipeline version, functions imported from Dashboard
-func GetPipelineVersion(c *cli.Clients) (string, error) {
-	deploymentsList, err := getDeployments(c, pipelinesControllerSelector, oldPipelinesControllerSelector)
+func GetPipelineVersion(c *cli.Clients, ns string) (string, error) {
+	deploymentsList, err := getDeployments(c, pipelinesControllerSelector, oldPipelinesControllerSelector, ns)
 
 	if err != nil {
 		return "", err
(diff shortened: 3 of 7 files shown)


pkg/version/version.go
```diff
@@ -23,6 +23,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// NOTE: use go build -ldflags "-X github.com/tektoncd/cli/pkg/version.defaultNamespace=tekton-pipelines"
+var defaultNamespace = "tekton-pipelines"
```
piyush-garg, line 27: Is not better to put defaultNamespace value in flag itself

vdemeester: @piyush-garg wdym ? we need to have that variable "defined" so that we can overwrite it with the `ldflags`. Being here seems to make more sense to me than in the `cmd` part (with the flags).

chmouel:
Not sure if that was discussed but can't we have a list of most common namespaces in `defaultNamespace` by default?

```go
defaultNamespaces := []string{
   "tekton-pipeline",
   "openshift-pipeline",
   "otherinthefutureofpopularnamespacefortekton",
}
```

so an upstream tkn binary can work on openshift-pipelines by default (and vice versa) ?

we try the list in order to find the version in there?

savitaashture:
> ```go
> pipeline
> ```

Hi @chmouel I don't remember exactly like we discussed to keep list of namespace but we can do that instead of one namespace `tekton-pipeline` we can keep 2 right now 
```
defaultNamespaces := []string{
   "tekton-pipeline",
   "openshift-pipeline",
}
```
But i am not sure if we have someother namespace used by anyother vendor.

So shall i go ahead with those 2 namespace right now ?

Makefile
```diff
@@ -12,9 +16,9 @@ ifneq ($(RELEASE_VERSION),)
 	VERSIONLDFLAG := -X github.com/tektoncd/cli/pkg/cmd/version.clientVersion=$(RELEASE_VERSION)
 endif
 
-FLAGS := $(VERSIONLDFLAG)$(SKIPLDFLAG)
+FLAGS := $(VERSIONLDFLAG)$(SKIPLDFLAG)$(NAMESPACEFLAG)
 ifneq ($(SKIPLDFLAG),)
-	FLAGS := $(VERSIONLDFLAG) $(SKIPLDFLAG)
+	FLAGS := $(VERSIONLDFLAG) $(SKIPLDFLAG)$(NAMESPACEFLAG)
```
piyush-garg, line 21: I think we need an if case for this also, there may be scenarios of only namespace flag provided

Makefile
```diff
@@ -1,5 +1,9 @@
 YAML_FILES := $(shell find . -type f -regex ".*y[a]ml" -print)
 
+ifneq ($(NAMESPACE),)
+	NAMESPACEFLAG := -X github.com/tektoncd/cli/pkg/version.defaultNamespace=$(NAMESPACE)
```
piyush-garg, line 4: Should we name it `NAMESPACELDFLAG` to be consistent with all other flags

pkg/cmd/version/version.go
```diff
@@ -30,17 +30,19 @@ import (
...
 	devVersion       = "dev"
 	latestReleaseURL = "https://api.github.com/repos/tektoncd/cli/releases/latest"
 )
 
+var (
+	// NOTE: use go build -ldflags "-X github.com/tektoncd/cli/pkg/cmd/version.clientVersion=$(git describe)"
+	clientVersion = devVersion
+	// flag to skip check flag in version cmd
+	skipCheckFlag = "false"
+	namespace     string
```
piyush-garg, line 43: @vdemeester I mean we can put the default value here, and this can be overridden with LDFLAG and also namespace flag in cmd. And we can further use this in all functions. This way we can remove the defaultNamespace flag from pkg cc @savitaashture

pkg/version/version.go
piyush-garg: Replied here https://github.com/tektoncd/cli/pull/1092/files#r464241241

pkg/cmd/version/version.go
piyush-garg: This can be single source for namespace then rather than two different for cmd and LDFLAG

piyush-garg:
> > ```go
> > pipeline
> > ```
> 
> Hi @chmouel I don't remember exactly like we discussed to keep list of namespace but we can do that instead of one namespace `tekton-pipeline` we can keep 2 right now
> 
> ```
> defaultNamespaces := []string{
>    "tekton-pipeline",
>    "openshift-pipeline",
> }
> ```
> 
> But i am not sure if we have someother namespace used by anyother vendor.
> 
> So shall i go ahead with those 2 namespace right now ?

Nit: AFAIR, namespaces are 
```
defaultNamespaces := []string{
   "tekton-pipelines",
   "openshift-pipelines",
}
```

chmouel:
@savitaashture Sounds good to me ! (the list of @piyush-garg openshift-pipeline with s :))

vdemeester:
```go
defaultNamespaces := []string{
   "tekton-pipelines",
   "openshift-pipelines",
}
```

If we use that, how does `ldflags` work with an array of string ? :thinking:

chmouel:
@vdemeester If there is a ldflag that was specified it will : 

1. Try the one from ldlfags
2. Try the defaultNamespaces

Makefile
savitaashture: I am not sure i understand completely like in which scenario we need if condition 

and in the current code also i see with and without if condition `flags` are same i mean without any condition flags are [here](https://github.com/tektoncd/cli) and with condition flags are [here](https://github.com/tektoncd/cli/blob/master/Makefile#L17) and both are same correct me if i mistaken something

pkg/cmd/version/version.go
savitaashture: @piyush-garg updated code to have same var for both cmd and LDFLAG but not for default because now default is an []string{} and as per my understanding LDFLAG doesn't take the `[]` .
Still trying if something can be done to make it single var

savitaashture:
@chmouel @vdemeester @piyush-garg 

Addressed all the comments and rebased please do have a look 

Thank you

chmouel:
@savitaashture i think the unit test needs to be fixed for 0_10?

savitaashture:
> @savitaashture i think the unit test needs to be fixed for 0_10?

@chmouel Done

vdemeester (approved):
/meow

Makefile
```diff
@@ -12,9 +16,9 @@ ifneq ($(RELEASE_VERSION),)
 	VERSIONLDFLAG := -X github.com/tektoncd/cli/pkg/cmd/version.clientVersion=$(RELEASE_VERSION)
 endif
 
-FLAGS := $(VERSIONLDFLAG)$(SKIPLDFLAG)
+FLAGS := $(VERSIONLDFLAG)$(SKIPLDFLAG)$(NAMESPACELDFLAG)
 ifneq ($(SKIPLDFLAG),)
-	FLAGS := $(VERSIONLDFLAG) $(SKIPLDFLAG)
+	FLAGS := $(VERSIONLDFLAG) $(SKIPLDFLAG) $(NAMESPACELDFLAG)
```
piyush-garg, line 21: we need to handle the cases here carefully, I think it will cause an issue when you will have skip flag but not namespace flag

or 

we need to handle namespace flag case

pkg/cmd/version/version.go
```diff
@@ -56,19 +59,23 @@ func Command(p cli.Params) *cobra.Command {
 
 			cs, err := p.Clients()
 			if err == nil {
-				pipelineVersion, _ := version.GetPipelineVersion(cs)
+				pipelineVersion, _ := version.GetPipelineVersion(cs, namespace)
 				if pipelineVersion == "" {
-					pipelineVersion = "unknown"
+					pipelineVersion = "unknown, " +
+						"pipeline controller may be installed in another namespace please use tkn version -n {namespace}"
```
piyush-garg, line 65: Do we need this, as we are planning to no show version in case of unknown

savitaashture: I see still it exist in the current code so in order to keep in sync with master added this change but if there is a plan to not to show `unknown` then i think we can remove while doing the PR related to removing `unknown` changes

WDYT?

Makefile
savitaashture: I have tested without specifying `skip` flag like below

```
GOOS=linux GOARCH=amd64 go build -mod=vendor -ldflags "-X github.com/tektoncd/cli/pkg/cmd/version.namespace=tekton-pipelines"
```

and it works without any issue 

let me know if i miss understand the point which you have mentioned

piyush-garg: You are not using makefile here :D

savitaashture: Updated makefile

tested with following scenario

* `NAMESPACE=tekton-pipelines SKIP_CHECK_FLAG=true RELEASE_VERSION=0.13 make amd64`
* `NAMESPACE=tekton-pipelines make amd64`
* `NAMESPACE=tekton-pipelines RELEASE_VERSION=0.13 make amd64`
* `SKIP_CHECK_FLAG=true NAMESPACE=tekton-pipelines make amd64`

pkg/cmd/version/version.go
```diff
@@ -84,6 +91,9 @@ func Command(p cli.Params) *cobra.Command {
 		},
 	}
 
+	cmd.Flags().StringVarP(&namespace, "namespace", "n", namespace,
```
piyush-garg, line 94: Just thought instead of this, we should set the global flags here so that we will get all along with namespace

piyush-garg: I think this will not make much sense here as the need to accept param from ldflag

piyush-garg (approved):
/lgtm

PR #1149, opened 2020-09-02
Fixing clustertask e2e tests

VeereshAradhya:
The current clustertask e2e test assumes that the test will be run on a cluster which does not have any clustertasks

Fixing it to work with cluster which already has some clustertasks

# Changes

# Submitter Checklist

review them:

for more details._

# Release Notes

### DIFF
--- a/test/e2e/clustertask/start_test.go
+++ b/test/e2e/clustertask/start_test.go
@@ -16,6 +16,10 @@
 package clustertask
 
 import (
+	"fmt"
+	"os"
+	"regexp"
+	"strings"
 	"testing"
 	"time"
 
@@ -47,39 +51,47 @@ func TestClusterTaskInteractiveStartE2E(t *testing.T) {
 	kubectl := cli.NewKubectl(namespace)
 	tkn, err := cli.NewTknRunner(namespace)
 	assert.NilError(t, err)
+	// Set environment variable NO_CLUSTERTASK_FOUND_TEST to any value to skip "No ClusterTasks found" test
+	if os.Getenv("NO_CLUSTERTASK_FOUND_TEST") == "" {
+		t.Run("Get list of ClusterTasks when none present", func(t *testing.T) {
+			res := tkn.Run("clustertask", "list")
+			expected := "No ClusterTasks found\n"
+			res.Assert(t, icmd.Expected{
+				ExitCode: 0,
+				Err:      icmd.None,
+				Out:      expected,
+			})
 
-	t.Run("Get list of ClusterTasks when none present", func(t *testing.T) {
-		res := tkn.Run("clustertask", "list")
-		expected := "No ClusterTasks found\n"
-		res.Assert(t, icmd.Expected{
-			ExitCode: 0,
-			Err:      icmd.None,
-			Out:      expected,
 		})
-	})
-
+	}
 	t.Logf("Creating clustertask read-clustertask")
-	kubectl.MustSucceed(t, "create", "-f", helper.GetResourcePath("read-file-clustertask.yaml"))
+	res := kubectl.MustSucceed(t, "create", "-f", helper.GetResourcePath("read-file-clustertask.yaml"))
+	regex := regexp.MustCompile(`read-clustertask-[a-z0-9]+`)
+	clusterTaskName := regex.FindString(res.Stdout())
 
 	t.Logf("Creating git pipeline resource in namespace: %s", namespace)
 	kubectl.MustSucceed(t, "create", "-f", helper.GetResourcePath("git-resource.yaml"))
 
 	t.Run("Get list of ClusterTasks", func(t *testing.T) {
 		res := tkn.Run("clustertask", "list")
-		expected := builder.ListAllClusterTasksOutput(t, c, map[int]interface{}{
-			0: &builder.TaskData{
-				Name: "read-clustertask",
-			},
-		})
-		res.Assert(t, icmd.Expected{
-			ExitCode: 0,
-			Err:      icmd.None,
-			Out:      expected,
-		})
+		if os.Getenv("NO_CLUSTERTASK_FOUND_TEST") == "" {
+			expected := builder.ListAllClusterTasksOutput(t, c, map[int]interface{}{
+				0: &builder.TaskData{
+					Name: clusterTaskName,
+				},
+			})
+			res.Assert(t, icmd.Expected{
+				ExitCode: 0,
+				Err:      icmd.None,
+				Out:      expected,
+			})
+		} else {
+			assert.Assert(t, strings.Contains(res.Stdout(), clusterTaskName))
+		}
 	})
 
 	t.Run("Start ClusterTask with flags", func(t *testing.T) {
-		res := tkn.MustSucceed(t, "clustertask", "start", "read-clustertask",
+		res := tkn.MustSucceed(t, "clustertask", "start", clusterTaskName,
 			"-i=source="+tePipelineGitResourceName,
 			"-p=FILEPATH=docs",
 			"-p=FILENAME=README.md",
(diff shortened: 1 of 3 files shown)


test/e2e/clustertask/start_test.go
```diff
@@ -232,4 +214,21 @@ Waiting for logs to be available...
 			t.Fatalf("-got, +want: %v", d)
 		}
 	})
+
+	t.Logf("Deleting clustertask read-clustertask")
+	t.Run("Delete clustertask read-clustertask", func(t *testing.T) {
+		res := tkn.Run("clustertask", "delete", "read-clustertask", "-f")
```
danielhelfand, line 220:
```suggestion
		res := tkn.MustSucceed("clustertask", "delete", "read-clustertask", "-f")
```

test/e2e/clustertask/start_test.go
```diff
@@ -232,4 +214,21 @@ Waiting for logs to be available...
...
+		expected := "ClusterTasks deleted: \"read-clustertask\""
+		res.Assert(t, icmd.Expected{
+			ExitCode: 0,
```
danielhelfand, line 223: You can remove `ExitCode: 0` if `MustSucceed` is used above.

test/e2e/clustertask/start_test.go
```diff
@@ -48,16 +49,6 @@ func TestClusterTaskInteractiveStartE2E(t *testing.T) {
 	tkn, err := cli.NewTknRunner(namespace)
 	assert.NilError(t, err)
 
-	t.Run("Get list of ClusterTasks when none present", func(t *testing.T) {
```
danielhelfand, line 51: Can you explain the rationale for not needing this test? Is there another way this test can be kept to preserve testing the case where No ClusterTasks are found?

VeereshAradhya: The problem with clutertask is even if we create new namespace the tasks will be visible. Maybe we can cover this in unit test if it is already not present :) or else in any other test where we just check the individual functionality and not the complete workflow

danielhelfand: I guess what is a bit confusing to me is where the issue is occurring with this if the test itself is not failing?

VeereshAradhya: The test only passes in a cluster which does not contain any clustertask. If I want to run the test in a cluster which contains cluster tasks the test fails :|

piyush-garg: In e2e, we run on a fresh cluster every time, so I think it is fine to keep this

danielhelfand: I am wondering if the idea here is to do something downstream with these tests where the cluster has clustertasks already?

VeereshAradhya: Instead of removing the test, I am checking for an environment variable `NO_CLUSTERTASK_FOUND_TEST`, if it is set I don't run the `No ClusterTasks found` test (which is downstream) and if `NO_CLUSTERTASK_FOUND_TEST` is not set, then I will `No ClusterTasks found` test (which is upstream).

test/e2e/clustertask/start_test.go
```diff
@@ -47,47 +51,55 @@ func TestClusterTaskInteractiveStartE2E(t *testing.T) {
 	kubectl := cli.NewKubectl(namespace)
 	tkn, err := cli.NewTknRunner(namespace)
 	assert.NilError(t, err)
+	// Set environment variable NO_CLUSTERTASK_FOUND_TEST to any value to skip "No ClusterTasks found" test
+	if os.Getenv("NO_CLUSTERTASK_FOUND_TEST") == "" {
```
piyush-garg, line 55: A better name of flag like `TEST_CLUSTERTASK_LIST_EMPTY`

test/e2e/clustertask/start_test.go
```diff
@@ -47,47 +51,55 @@ func TestClusterTaskInteractiveStartE2E(t *testing.T) {
...
-			0: &builder.TaskData{
-				Name: "read-clustertask",
-			},
-		})
-		res.Assert(t, icmd.Expected{
-			ExitCode: 0,
-			Err:      icmd.None,
-			Out:      expected,
-		})
+		if os.Getenv("NO_CLUSTERTASK_FOUND_TEST") == "" {
```
piyush-garg, line 77: We don't need to skip for this test

test/e2e/clustertask/start_test.go
piyush-garg: And can we document about the flag

test/e2e/clustertask/start_test.go
VeereshAradhya: If there are no clustertasks when we run the test then expected value of `tkn ct ls` is  `read-clustertask`. But when we have other clustertasks when we run this test then the expected value of `tkn ct ls` is `read-clustertask` along with other task. So I am running test depending on environment variable

piyush-garg (commented):
/lgtm

test/e2e/clustertask/start_test.go
```diff
@@ -47,47 +51,55 @@ func TestClusterTaskInteractiveStartE2E(t *testing.T) {
...
-			0: &builder.TaskData{
-				Name: "read-clustertask",
-			},
-		})
-		res.Assert(t, icmd.Expected{
-			ExitCode: 0,
-			Err:      icmd.None,
-			Out:      expected,
-		})
+		if os.Getenv("TEST_CLUSTERTASK_LIST_EMPTY") == "" {
```
danielhelfand, line 77: Why is this check needed?

test/e2e/clustertask/start_test.go
```diff
@@ -232,4 +244,20 @@ Waiting for logs to be available...
...
+		res.Assert(t, icmd.Expected{
+			Err: icmd.None,
+			Out: expected,
+		})
+	})
+
+	t.Run(fmt.Sprintf("Check if clustertask %s got deleted", clusterTaskName), func(t *testing.T) {
+		res := tkn.Run("clustertask", "list")
+		assert.Assert(t, !strings.Contains(res.Stdout(), clusterTaskName))
+	})
```
danielhelfand, line 261: This should be a single test.

test/resources/read-file-clustertask.yaml
```diff
@@ -15,7 +15,7 @@
 apiVersion: tekton.dev/v1alpha1
 kind: ClusterTask
 metadata:
-  name: read-clustertask
+  generateName: read-clustertask-
```
danielhelfand, line 18: Why do we need to use `generateName`?

danielhelfand (changes requested):
In general, I am still a bit reluctant to add this approach upstream. I am okay with accepting this for now after comments are addressed, but I think there should be a better approach when dealing with downstream testing issues that should probably be addressed downstream.

test/e2e/clustertask/start_test.go
VeereshAradhya: If the cluster contains other clustertasks, this test fails because we are checking for only one clustertask (`read-clustertask`) in the output of ``tkn ct ls`. So adding a check if

test/resources/read-file-clustertask.yaml
VeereshAradhya: If the same set of tests are run in multiple platforms in parallel, having the same name for the cluster task causes problem.

vdemeester (commented):
I think a better approach, without skipping this, would be to list the cluster task present before running the tests and take those into account during checks (aka if there is any already, substract the number, or …)
/lgtm

VeereshAradhya:
In that case, we would be missing `No clusteratsks found` test. :(

vdemeester:
> In that case, we would be missing `No clusteratsks found` test. :(

`TEST_CLUSTERTASK_LIST_EMPTY` makes use skip the test right ? so in the end, it's doing the same thing.

VeereshAradhya:
If we have not set `TEST_CLUSTERTASK_LIST_EMPTY` the `No clusteratsks found` test will run :)

danielhelfand (approved):
/cc @piyush-garg @pradeepitm12 @chmouel

PR #1158, opened 2020-09-11
Fix slowness in executing tkn commands

### LINKED ISSUE #1153: Slow `tkn` operation

- Kubernetes version:

OpenShift 4.6

- Tekton Pipeline version:

nightly

- Operating System:

OSX

# Actual Behavior

`tkn pipeline ls` has been feeling very slow, the released or nightly version : 

i.e over 11s for a simple `tkn p ls` : 

```
% time ./bin/tkn p ls
NAME               AGE              LAST RUN   STARTED   DURATION   STATUS
pipeline-buildah   56 seconds ago   ---        ---       ---        ---
./bin/tkn p ls  0.14s user 0.07s system 1% cpu 11.329 total
```

or even longer for a `tkn p describe` : 

```
./bin/tkn p describe pipeline-buildah  0.18s user 0.09s system 1% cpu 26.565 total
```

It doesn't seem to be a latency since kubectl operation takes even less than a second : 

```
% time (k get pipeline >/dev/null && k describe pipeline pipeline-buildah >/dev/null)
( kubectl get pipeline > /dev/null && kubectl describe pipeline  > /dev/null;   0.17s user 0.05s system 22% cpu 0.984 total
```

piyush-garg:
This will fix the issue of slowness in
tkn commands. We were fetching apiGroupRes
details everytime in a api call using client-go.
Now we store the results and reuse in further
operations. So in a single tkn command execution
we fetch the details of apiGroupRes only once
and reuse in further operations api-call.

Example like for tkn pipelines ls command,
we need the pipelines and also pipelineruns
for all pipelines. We were fetching apiGroupRes
everytime like during fetching pipelines and further
pipelineruns which depends on no. of pipelines.
Now we store the result in single execution
and reuse this for all operations.

Refactoring in unit tests accordingly.

Remove redundant golden files.

Fix #1153

# Submitter Checklist

review them:

for more details._

# Release Notes

```release-note
   Fix tkn slowness issue, taking lot of time
```


pkg/actions/gvr.go
```diff
@@ -34,3 +37,12 @@ func GetGroupVersionResource(gr schema.GroupVersionResource, discovery discovery
 
 	return &gvr, nil
 }
+
+func InitializeAPIGroupRes(discovery discovery.DiscoveryInterface) error {
+	var err error
+	apiGroupRes, err = restmapper.GetAPIGroupResources(discovery)
```
danielhelfand, line 43:
```suggestion
	apiGroupRes, err := restmapper.GetAPIGroupResources(discovery)
```

pkg/actions/gvr.go
```diff
@@ -34,3 +37,12 @@ func GetGroupVersionResource(gr schema.GroupVersionResource, discovery discovery
 
 	return &gvr, nil
 }
+
+func InitializeAPIGroupRes(discovery discovery.DiscoveryInterface) error {
```
sthaha, line 41: do we need a separate function for this?

pkg/clustertask/clustertask_test.go
```diff
@@ -91,22 +92,28 @@ func TestClusterTask_List(t *testing.T) {
...
 			name:   "Multi clusterTasks",
 			client: c2,
+			params: p2,
 			want:   []string{"clustertask", "clustertask2"},
 		},
 	}
 
 	for _, tp := range testParams {
 		t.Run(tp.name, func(t *testing.T) {
+			if err := actions.InitializeAPIGroupRes(tp.params.Tekton.Discovery()); err != nil {
```
sthaha, line 114: Wouldn't this be initialised using `tp.params.Tekton.Discovery()` automatically?

pkg/actions/gvr.go
piyush-garg: we need this for unit tests to have the apiGroupRes, we need to call it from every test

pkg/clustertask/clustertask_test.go
piyush-garg: No, it was not getting initialized

danielhelfand (commented):
A question/comment:

1. For my own understanding, under what circumstances is `apiGroupRes==nil`, and how does it reset to nil?

2. This approach may also be useful for getting the apiVersion for `--dry-run` with start commands: 

https://github.com/tektoncd/cli/blob/1f0312a6f6b25a1b8a3ce8a36d4b065b3c60026f/pkg/cmd/pipeline/start.go#L859

https://github.com/tektoncd/cli/blob/1f0312a6f6b25a1b8a3ce8a36d4b065b3c60026f/pkg/cmd/task/start.go#L469

https://github.com/tektoncd/cli/blob/1f0312a6f6b25a1b8a3ce8a36d4b065b3c60026f/pkg/cmd/clustertask/start.go#L372

pkg/actions/gvr.go
piyush-garg: as `apiGroupRes` has been declared before, this is throwing errors

piyush-garg:
> A question/comment:
> 
> 1. For my own understanding, under what circumstances is `apiGroupRes==nil`, and how does it reset to nil?
> 2. This approach may also be useful for getting the apiVersion for `--dry-run` with start commands:
> 
> https://github.com/tektoncd/cli/blob/1f0312a6f6b25a1b8a3ce8a36d4b065b3c60026f/pkg/cmd/pipeline/start.go#L859
> 
> https://github.com/tektoncd/cli/blob/1f0312a6f6b25a1b8a3ce8a36d4b065b3c60026f/pkg/cmd/task/start.go#L469
> 
> https://github.com/tektoncd/cli/blob/1f0312a6f6b25a1b8a3ce8a36d4b065b3c60026f/pkg/cmd/clustertask/start.go#L372

By default the apiGroupRes is nil, so after the first call it has the value. So everytime user execute a command, apiroupRes will be nil, and after any first API call it has the value, and we don't need to fetch this value again and again for all operations.

As you will execute another command after that, it will be nil again

pkg/actions/gvr.go
```diff
@@ -20,10 +20,13 @@ import (
 	"k8s.io/client-go/restmapper"
 )
 
+var apiGroupRes []*restmapper.APIGroupResources
```
vdemeester, line 23: We *may* want to use https://pkg.go.dev/sync?tab=doc#Once for this (to make it a little bit cleaner)

pkg/actions/gvr.go
vdemeester: Any reason we need to call it from every test ? I would *like* this fix to not have to modify all the tests.

pkg/clustertask/clustertask_test.go
vdemeester: Same as @sthaha , `List` seems to call `actions.List` that calls `GetGroupVersionResource`, so I don't understand why we need to call `InitializeAPIGroupRes` before.

vdemeester (commented):
The diff seems way too big for that fix. I would think we don't need to update tests, at least not calling `InitializeAPIGroupRes` before each. Also, this feels a very good usecase for `sync.Once` :upside_down_face:

pkg/actions/gvr.go
piyush-garg: Improved according to comments

pkg/clustertask/clustertask_test.go
piyush-garg: Because it is getting initialized in running first test, then for second test it was not getting initialized as the value is already so there, so getting called from every test to have the value once.

pkg/actions/gvr.go
piyush-garg: Updated to use sync

piyush-garg:
@sthaha @danielhelfand @vdemeester @chmouel Refactored the PR according to comments

pkg/cmd/clustertriggerbinding/clustertriggerbinidng_test.go
```diff
@@ -0,0 +1,31 @@
+// Copyright © 2020 The Tekton Authors.
```
danielhelfand, line 1: Just a question. Is this related to this pr?

danielhelfand (commented):
/lgtm

pkg/cmd/clustertriggerbinding/clustertriggerbinidng_test.go
piyush-garg: I was going through tests, and I found this is missing one and there are few redundant golden files which are not used, so did all refactoring :D

pradeepitm12 (commented):
/lgtm

vdemeester (approved):
/meow

pkg/test/builder/apiresource.go
```diff
@@ -36,9 +36,9 @@ func APIResourceList(version string, kinds []string) []*metav1.APIResourceList {
 func apiresources(version string, kinds []string) []metav1.APIResource {
 	apires := make([]metav1.APIResource, 0)
 	for _, kind := range kinds {
-		namespaced := false
+		namespaced := true
 		if strings.Contains(kind, "cluster") {
-			namespaced = true
+			namespaced = false
```
chmouel, line 41: 😅  was it used anywhere?

piyush-garg: not sure, It did not make any difference anywhere

chmouel:
Looks good to me, but let's wait for @sthaha +1 before the final approve.

piyush-garg:
@chmouel It is merged now, but if @sthaha has any concerns, I will raise separate PR to address those. Please give it a review @sthaha

sthaha:
@piyush-garg I had a look at the code and I don't see an easy way to reset the cache automatically for each test  without quite some refactor. 

@piyush-garg however I was wondering if we could resync the cache using the discovery.Interface() if there is a cache miss before it errors. I believe that may solve the need to add `InitializeCache` in all tests.

Issue #1039, opened 2020-06-30
Extension execution model


vdemeester:
### Feature request

It should be possible to have a simple *extension* execution model. It would be similar to `git` or `kubectl`, in that, if a command is prefixed by `tkn-` it could be seens as a sub command.

```bash
$ ls /usr/local/bin
tkn-foo
$ tkn foo --bar
# invokirng /usr/local/bin/tkn-foo with following args
# […]
```

It would, however not be able to *shadow* existing subcommands, aka if I define a `tkn-task` and put it in my `PATH`, it wouldn't affect `tkn`.

### Use case

There is at least two *use case* that would benefit from this feature:
- experimental projects and other repositories could use this to be able to quickly enhance the `tkn` experience and gather feedback before integrating in `tkn` directly
- it would reduce the potential dependencies headaches that we may have by integrating all tekton components. We already can have this with the `triggers` integration if the dependency of pipeline we track and the one triggers track differs.

We could think of quite some use case

```bash
# the hub (https://github.com/tektoncd/hub)
$ tkn hub search kaniko
…
$ tkn hub import buildah
$ tkn hub import buildah/0.2 --as  buildah-0_2
# the result API (https://github.com/tektoncd/experimental/tree/master/results)
$ tkn result list
$ tkn result logs an-old-archived-pipeline
# the OCI experiment
$ tkn oci create --from ./tasks/my-task.yaml --as quay.io/vdemeest/foo:is-bar
$ tkn oci extract quay.io/vdemeester/foo:is-baz
$ tkn oci add …
# the dashboard (https://github.com/tektoncd/dashboard)
$ tkn dashboard deploy …
# the operator (https://github.com/tektoncd/operator)
$ tkn operator install …
```

Note that **the *distribution* part of this is not covered** and can be discussed later on. Once we have that feature we can decide what could be the distribution model (we can also decide it's not in the `tkn` responsabilies and the distribution can be handle separately — by package managers, `krew`, …).

/cc @chmouel @sthaha @ImJasonH 

/kind feature
/milestone "0.12.0"

chmouel:
👍

danielhelfand:
For clarification, this would mean every project basically supporting some project-specific cli (e.g. triggers, dashboard, operator, etc.)? Think the only thing needed would be some way to avoid overlap between projects, but in favor of the basic concept. 

We might also want to develop some type of CLI style guide to enforce some level of consistency for these projects.

vdemeester:
[...]

PR #1136, opened 2020-08-20
WIP: Attempt at adding a plugin sub-cmd

sthaha:
# Changes
This is only a POC that adds a plugin extension model like kubectl (tested only on linux)

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [🙅‍♂️  ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [🙅‍♂️  ] Run the code checkers with `make check`
- [🙅‍♂️ ] Regenerate the manpages, docs and go formatting with `make generated`
- [🙅‍♂️ ] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/cli/blob/master/CONTRIBUTING.md)
for more details._

# Release Notes

PR #1178, opened 2020-09-18
Basic execution plugin module

vdemeester:
# Changes

This adds a basic execution plugin model, very simple to be honest. Most of the work has been done by @sthaha. In a gist this is how it works

- If the specified subcommand doesn't exists, we check if a command
  with the subcommand prefixed by `tkn-` exists in the
  `$HOME/.config/tkn/pulgins` (for `foo` we look for `tkn-foo`).
- The default folder (`$HOME/.config/tkn/plugins`) can be overriden
  with `TKN_PLUGINS_DIR`.
- If the binary doesn't exists, we proceed as usual
- If the binary exists, we execute it "à-la" `exec …`, aka we give full hand to the binary.

```bash
$ ls ~/.config/tkn/plugins/tkn-*
/home/vincent/.config/tkn/plugins/tkn-bar  /home/vincent/.config/tkn/plugins/tkn-foo
$ tkn foo is bar --hello
args is bar --hello
fooo
bar
$ tkn bar bar
/usr/bin/env: ‘invalidcommand’: No such file or directory
```

What may need to be enhanced :
- <del>This needs quick & dirty tests</del>
- This doesn't pass *any* global flags (value or as arguments — `tkn` doesn't have any it seems) — any "client handling" is on the plugin binary to handle
- This doesn't take into account `help` — `tkn foo --help` would call `tkn-foo --help`, *but* `tkn help` won't show `foo` as subcommand. This can probably be taken care of in follow-ups ?
- <del>This only look into `PATH`. Maybe we want to have some "standard" </del>

Carrying #1136.
Fixes: #1039

/cc @tektoncd/cli-maintainers 

# Submitter Checklist

review them:

for more details._

# Release Notes

```release-note
`tkn` now support a basic plugin system. Any command prefixed by `tkn-` in the `PATH` can be used as a `tkn` subcommands (e.g. `tkn-foo` can be used as `tkn foo`).
```

### DIFF
--- a/cmd/tkn/main.go
+++ b/cmd/tkn/main.go
@@ -15,18 +15,77 @@
 package main
 
 import (
+	"fmt"
 	"os"
+	"path/filepath"
+	"syscall"
 
+	homedir "github.com/mitchellh/go-homedir"
+	"github.com/pkg/errors"
 	"github.com/tektoncd/cli/pkg/cli"
 	"github.com/tektoncd/cli/pkg/cmd"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
+)
+
+const (
+	pluginDirEnv = "TKN_PLUGINS_DIR"
+	pluginDir    = "~/.config/tkn/plugins"
 )
 
 func main() {
 	tp := &cli.TektonParams{}
 	tkn := cmd.Root(tp)
 
+	args := os.Args[1:]
+	cmd, _, _ := tkn.Find(args)
+	if cmd != nil && cmd == tkn && len(args) > 0 {
+		pluginCmd := "tkn-" + os.Args[1]
+		pluginDir, err := getPluginDir()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error getting plugin folder: %v", err)
+		}
+		exCmd, err := findPlugin(pluginDir, pluginCmd)
+		if err != nil {
+			// Can't find the binary in PATH, bailing to usual execution
+			goto CoreTkn
+		}
+
+		if err := syscall.Exec(exCmd, append([]string{exCmd}, os.Args[2:]...), os.Environ()); err != nil {
+			fmt.Fprintf(os.Stderr, "Command finished with error: %v", err)
+			os.Exit(127)
+		}
+		return
+	}
+
+CoreTkn:
 	if err := tkn.Execute(); err != nil {
 		os.Exit(1)
 	}
+}
+
+func getPluginDir() (string, error) {
+	dir := os.Getenv(pluginDirEnv)
+	// if TKN_PLUGINS_DIR is set, follow it
+	if dir != "" {
+		return dir, nil
+	}
+	// Respect XDG_CONFIG_HOME if set
+	if xdgHome := os.Getenv("XDG_CONFIG_HOME"); xdgHome != "" {
+		return filepath.Join(xdgHome, "tkn", "plugins"), nil
+	}
+	// Fallback to default pluginDir (~/.config/tkn/plugins)
+	return homedir.Expand(pluginDir)
+}
+
+func findPlugin(dir, cmd string) (string, error) {
+	path := filepath.Join(dir, cmd)
+	_, err := os.Stat(path)
+	if err == nil {
+		// Found in dir
+		return path, nil
+	}
+	if !os.IsNotExist(err) {
+		return "", errors.Wrap(err, fmt.Sprintf("i/o error while reading %s", path))
+	}
+	return "", err
 }
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/hinshun/vt10x v0.0.0-20180809195222-d55458df857c
 	github.com/jonboulle/clockwork v0.1.1-0.20190114141812-62fb9bc030d1
 	github.com/ktr0731/go-fuzzyfinder v0.2.0
+	github.com/mitchellh/go-homedir v1.1.0
 	github.com/pkg/errors v0.9.1
 	github.com/spf13/cobra v1.0.0
 	github.com/spf13/pflag v1.0.5
--- a/test/e2e/plugin/plugin_test.go
+++ b/test/e2e/plugin/plugin_test.go
@@ -1,0 +1,49 @@
+// +build e2e
+// Copyright © 2020 The Tekton Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"os"
+	"testing"
+
+	"github.com/tektoncd/cli/test/cli"
+	"gotest.tools/v3/assert"
+	"gotest.tools/v3/env"
+	"gotest.tools/v3/icmd"
+)
+
+func TestTknPlugin(t *testing.T) {
+	tkn, err := cli.NewTknRunner("any-namespace")
+	assert.NilError(t, err)
+	currentpath, err := os.Getwd()
+	assert.NilError(t, err)
+	defer env.Patch(t, "TKN_PLUGINS_DIR", currentpath)()
+	t.Run("Success", func(t *testing.T) {
+		tkn.MustSucceed(t, "success")
+		tkn.MustSucceed(t, "success", "with", "args")
+	})
+	t.Run("Failure", func(t *testing.T) {
+		tkn.Run("failure").Assert(t, icmd.Expected{
+			ExitCode: 12,
+		})
+		tkn.Run("failure", "with", "args").Assert(t, icmd.Expected{
+			ExitCode: 12,
+		})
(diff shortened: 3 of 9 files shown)


cmd/tkn/main.go
```diff
@@ -26,7 +29,25 @@ func main() {
...
+	if cmd != nil && cmd == tkn {
+		pluginCmd := "tkn-" + os.Args[1]
+		exCmd, err := exec.LookPath(pluginCmd)
+		if err != nil {
+			// Can't find the binary in PATH, bailing to usual execution
+			goto CoreTkn
+		}
+
+		if err := syscall.Exec(exCmd, append([]string{exCmd}, os.Args[2:]...), os.Environ()); err != nil {
+			fmt.Printf("Command finished with error: %v", err)
```
sthaha, line 43: How about `fmt.Fprintf(os.Stderr, ...`

cmd/tkn/main.go
```diff
@@ -26,7 +29,25 @@ func main() {
...
+			os.Exit(254)
```
sthaha, line 44: why 254? should it be 127? `Command not found or not in path` ? doesn't `128+<exit-code>` have a different meaning? 🤔

cmd/tkn/main.go
```diff
@@ -26,7 +29,25 @@ func main() {
 	tp := &cli.TektonParams{}
 	tkn := cmd.Root(tp)
 
-	if err := tkn.Execute(); err != nil {
+	args := os.Args[1:]
+	cmd, _, _ := tkn.Find(args)
+	if cmd != nil && cmd == tkn {
```
sthaha, line 34: we should write a  test for this shouldn't we?  ' at least a shell test ?

cmd/tkn/main.go
```diff
@@ -26,7 +29,25 @@ func main() {
...
+
+		if err := syscall.Exec(exCmd, append([]string{exCmd}, os.Args[2:]...), os.Environ()); err != nil {
+			fmt.Printf("Command finished with error: %v", err)
+			os.Exit(254)
+		}
+		return
+	}
+
+CoreTkn:
+	if err := cmd.Execute(); err != nil {
```
sthaha, line 50: won't this crash if `cmd == nil` ?

cmd/tkn/main.go
vdemeester: @sthaha yeah it's on the desc of the PR :grin:

cmd/tkn/main.go
vdemeester: Random number :sweat_smile: I'll update to 127 :stuck_out_tongue_closed_eyes:

cmd/tkn/main.go
vdemeester: @sthaha right, it will, but `cmd` is never `nil`. But this is an *artefact* of a refactoring I tried :sweat_smile: it should be `tkn.Execute()`, I'll update :sweat_smile:

cmd/tkn/main.go
```diff
@@ -26,7 +29,25 @@ func main() {
 	tp := &cli.TektonParams{}
 	tkn := cmd.Root(tp)
 
-	if err := tkn.Execute(); err != nil {
+	args := os.Args[1:]
+	cmd, _, _ := tkn.Find(args)
+	if cmd != nil && cmd == tkn {
+		pluginCmd := "tkn-" + os.Args[1]
```
SM43, line 35: If I do only `tkn` with no args then this will break here. index out of range.
Should we handle this? If there is no argument then skip this.

cmd/tkn/main.go
```diff
@@ -26,7 +29,25 @@ func main() {
 	tp := &cli.TektonParams{}
 	tkn := cmd.Root(tp)
 
-	if err := tkn.Execute(); err != nil {
+	args := os.Args[1:]
+	cmd, _, _ := tkn.Find(args)
```
SM43, line 33: Why `tkn.Find(args)` doesn't return an error when an invalid command is passed? 
I tried with another CLI `Find` returns err when the command is invalid, it happens only with `tkn`. :smile:

vdemeester: @SM43 my guess is the `suggest` part but I am not sure :sweat:

vdemeester:
This should be ready to be reviewed. Fixed some problems *and* added a small e2e test :upside_down_face:

cmd/tkn/main.go
vdemeester: fixed it by checking `len(args)` :wink:

vdemeester:
```
TestPipelineRunLogE2E: pipelinerun_test.go:55: could not get pipelinerun: conversion webhook for tekton.dev/v1beta1, Kind=PipelineRun failed: conversion failed to version v1alpha1 for type [kind=PipelineRun group=tekton.dev version=v1beta1] -  the specified field/section is not available in v1alpha1
```
:thinking:

sthaha (commented):
looks good to me except of the the unrelated CI failure.

sthaha (commented):
/lgtm

danielhelfand (commented):
A few thoughts:
* I am wondering if the default should not just look at path but something like `~/.config/tkn` to help users separate out `tkn` plugins from what else is on their PATH
* We should follow up with documentation on plugin usage
* We should consider adding management features around plugins (e.g. `tkn plugin ls` to show available plugins)

It's a good starting point though and am excited to have it.

/lgtm

vdemeester:
> A few thoughts:
> 
>     * I am wondering if the default should not just look at path but something like `~/.config/tkn` to help users separate out `tkn` plugins from what else is on their PATH

Yes, that's a good point, that would be worth deciding before merging this :upside_down_face: (not a fan of `~/.config/tkn` to store binaries but other than that, might make sense to look into a specific path instead of `$PATH`).

>     * We should follow up with documentation on plugin usage

:+1: 

>     * We should consider adding management features around plugins (e.g. `tkn plugin ls` to show available plugins)

Yes, this is a whole other discussion and should be done in a follow-up if this *experimentation* is sucessful :stuck_out_tongue:.

danielhelfand:
> Yes, that's a good point, that would be worth deciding before merging this 🙃 (not a fan of ~/.config/tkn to store binaries but other than that, might make sense to look into a specific path instead of $PATH).

Something that mirrors `kn`'s plugin model location would make sense (`~/.config/tkn/plugins`). I would be fine with moving forward with starting with `$PATH` and supporting `~/.config/tkn` or something of the like later on though. I see no reason why we can't make it work for both.

vdemeester:
> Something that mirrors `kn`'s plugin model location would make sense (`~/.config/tkn/plugins`). I would be fine with moving forward with starting with `$PATH` and supporting `~/.config/tkn` or something of the like later on though. I see no reason why we can't make it work for both.

Right :sweat_smile: I'd rather start with supporting only `~/.config/tkn/plugins` at first and add support to `$PATH` if there is a need. I'll update :wink:

vdemeester:
Updated, `$HOME/.config/tkn/plugins` is the default folder, and you can override it with `TKN_PLUGINS_DIR` environment variable.

vdemeester:
This is really simple, it doesn't support multiple folder or using the `PATH`. That can be done later on.

danielhelfand:
Rebasing around master with #1249 should resolve EventListener test failures

cmd/tkn/main.go
```diff
@@ -15,18 +15,66 @@
...
 
+	"github.com/pkg/errors"
 	"github.com/tektoncd/cli/pkg/cli"
 	"github.com/tektoncd/cli/pkg/cmd"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 )
 
+const (
+	pluginDirEnv = "TKN_PLUGINS_DIR"
+	pluginDir    = ".config/tkn/plugins"
```
danielhelfand, line 31:
```suggestion
	pluginDir    = "~/.config/tkn/plugins"
```

danielhelfand (commented):
Looks good otherwise, just a suggestion for the default plugin path.

cmd/tkn/main.go
vdemeester: Oh right, I wanted to use `homedir` to expand this :sweat_smile:

vdemeester: Updated, using `homedir.Expand` now :wink:

vdemeester:
https://github.com/tektoncd/cli/pull/1252 fixes the cross error

danielhelfand (approved):

chmouel:
Is `~/.config/tkn` the right place from a LSB distro/point of view ? it should be ~/.local/share isn't it ? But most people would be confused and would not care about FHS or LSB i guess, can we have `~/.config/tekton/plugins` tho ? in case we end up with other stuff we want that is tekton related but not tkn releated.

krew has it in `kube/plugins`

vdemeester:
> Is `~/.config/tkn` the right place from a LSB distro/point of view ? it should be ~/.local/share isn't it ? But most people would be confused and would not care about FHS or LSB i guess, can we have `~/.config/tekton/plugins` tho ? in case we end up with other stuff we want that is tekton related but not tkn releated.
> 
> krew has it in `kube/plugins`

krew has `.krew/bin` doesn't it ? I choosed `tkn` instead of `tekton`, mainly because the tool is called `tkn` :sweat_smile:. If there is another cli tool for tekton apis named `foo` in the future, what make `tkn` more inclined to use `tekton` instead of `foo` ?

I don't have strong opinion on this though, I just want us to go with a given setup and see how it goes and how it is useful.

vdemeester:
@tektoncd/cli-maintainers are we good to merge given the result of the [poll](https://doodle.com/poll/n5iit5dayrsyxcyy) ?

piyush-garg (approved):

chmouel:
/lgtm

Looking Good! 🤙🏽

Issue #1139, opened 2020-08-21
Cancelling a pipelinerun which failed with `Failed(Cancelled)` gives success message


VeereshAradhya:
# Versions and Operating System

- Kubernetes version:

  **Output of `kubectl version`:**

  ```
  Client Version: version.Info{Major:"1", Minor:"10+", GitVersion:"v1.10.0+b3b92b2", GitCommit:"b3b92b2", GitTreeState:"clean", BuildDate:"2020-07-15T09:27:21Z", GoVersion:"go1.14.3", Compiler:"gc", Platform:"linux/amd64"}
  Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.3+012b3ec", GitCommit:"012b3ec", GitTreeState:"clean", BuildDate:"2020-07-24T07:23:10Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}
  ```

- Tekton Pipeline version:

  **Output of `tkn version` or `kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'`**

  ```
  $ tkn version
  Client version: 0.11.0
  Pipeline version: v0.14.3
  Triggers version: v0.6.1
  ```

- Operating System:

  ```
  Linux
  ```

# Expected Behavior
`tkn` should throw error

# Actual Behavior
`tkn` is showing success message

# Steps to Reproduce the Problem

1. Create a pipeline with a single task
2. Start the pipeline 
3. Cancel the taskrun

# Additional Info
Command execution logs
```
[varadhya@localhost testing]$ tkn p start pipeline-test-timeout 
? Value for param `first-param` of type `string`? check
PipelineRun started: pipeline-test-timeout-run-c8ksq

In order to track the PipelineRun progress run:
tkn pipelinerun logs pipeline-test-timeout-run-c8ksq -f -n veeresh-testing-non-admin
[varadhya@localhost testing]$ 
[varadhya@localhost testing]$ 
[varadhya@localhost testing]$ tkn pr ls
NAME                              STARTED          DURATION   STATUS
pipeline-test-timeout-run-c8ksq   20 seconds ago   ---        Running
[varadhya@localhost testing]$ 
[varadhya@localhost testing]$ 
[varadhya@localhost testing]$ tkn tr ls
NAME                                                 STARTED          DURATION   STATUS
[...]

PR #1185, opened 2020-09-25
Cancelling a pipelinerun which failed with `Failed(Cancelled)`  throws an error

PuneetPunamiya:
Signed-off-by: Puneet Punamiya <ppunamiy@redhat.com>

# Changes

# Submitter Checklist

review them:

for more details._

# Release Notes

### DIFF
--- a/pkg/cmd/pipelinerun/cancel.go
+++ b/pkg/cmd/pipelinerun/cancel.go
@@ -29,6 +29,7 @@ var (
 	succeeded   = formatted.ColorStatus("Succeeded")
 	failed      = formatted.ColorStatus("Failed")
 	prCancelled = formatted.ColorStatus("Cancelled") + "(PipelineRunCancelled)"
+	cancelled   = formatted.ColorStatus("Cancelled") + "(Cancelled)"
 	prTimeout   = formatted.ColorStatus("Failed") + "(" + v1beta1.PipelineRunReasonTimedOut.String() + ")"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/cmd/pipelinerun/cancel.go
```diff
@@ -75,7 +76,7 @@ func cancelPipelineRun(p cli.Params, s *cli.Stream, prName string) error {
 	}
 
 	prCond := formatted.Condition(pr.Status.Conditions)
```
piyush-garg, line 78: Is not better to don't format the condition here and do the decision bases on Condition Status, if the Condition Status is True or False then don't do the cancellation, if the condition status in unknown then do the cancellation

cc @vdemeester @danielhelfand

piyush-garg:
Also, should we add the case for `Cancelled` here https://github.com/tektoncd/cli/blob/master/pkg/formatted/k8s.go#L72

pkg/cmd/pipelinerun/cancel.go
danielhelfand: Yes, I think the original reason for using formatting here was a quick fix for the problem. Maybe what we can do here, since we are approaching the release, is keep this approach as is for now but open a separate issue to refactor based on status.

vdemeester: @piyush-garg agreed. Depending on how easy/hard it is to modify this, I would go with @danielhelfand suggestion :upside_down_face:

piyush-garg: Yes, let's create a refactor issue for this and handle later @PuneetPunamiya

danielhelfand (approved):
Related issue is #1139

pradeepitm12 (commented):
/lgtm

Issue #1111, opened 2020-08-07
Enable auto selection of the pipeline for `describe` command if only one pipeline is present


Divyansh42:
### Feature request

The describe command should auto select the pipeline if only one pipeline is present.

### Use case

User will not have to select the pipeline if only one pipeline is present.

PR #1187, opened 2020-09-26
Enable auto-select support in pipelineDescribe if only one pipeline is present

Divyansh42:
# Changes

When only one `pipeline` is present then instead of prompting to user to select the
pipeline at the time of running the `pipeline describe` command then pipeline will
be `autoselected` and the output will be displayed.

Closes: #1111 

Signed-off-by: Divyansh42 <diagrawa@redhat.com>

# Submitter Checklist

review them:

for more details._

# Release Notes

```release-note
Enable auto-select support in Pipeline describe command if only one is present
```

### DIFF
--- a/pkg/cmd/pipeline/describe.go
+++ b/pkg/cmd/pipeline/describe.go
@@ -17,7 +17,6 @@ package pipeline
 import (
 	"fmt"
 	"io"
-	"os"
 	"sort"
 	"strings"
 	"text/tabwriter"
@@ -131,9 +130,17 @@ func describeCommand(p cli.Params) *cobra.Command {
 			}
 
 			if len(args) == 0 {
-				err = askPipelineName(opts, p)
+				pipelineNames, err := pipeline.GetAllPipelineNames(p)
 				if err != nil {
 					return err
+				}
+				if len(pipelineNames) == 1 {
+					opts.PipelineName = pipelineNames[0]
+				} else {
+					err = askPipelineName(opts, pipelineNames)
+					if err != nil {
+						return err
+					}
 				}
 			} else {
 				opts.PipelineName = args[0]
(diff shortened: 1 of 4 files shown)


pkg/cmd/pipeline/describe_test.go
```diff
@@ -117,6 +117,47 @@ func TestPipelineDescribe_empty(t *testing.T) {
 	golden.Assert(t, got, fmt.Sprintf("%s.golden", t.Name()))
 }
 
+func TestPipelineDescribe_WithoutNameOfOnlyOnePipelinePresent(t *testing.T) {
+	clock := clockwork.NewFakeClock()
+
+	pipelines := []*v1alpha1.Pipeline{
+		tb.Pipeline("pipeline",
+			tb.PipelineNamespace("ns"),
+			// created  5 minutes back
```
vinamra28, line 126: Please try to avoid the use of `tb` or test builders when writing unit tests as they will be deprecated in the future. You can use structs instead like you did in the `v1beta1` test.

pkg/cmd/pipeline/describe.go
```diff
@@ -131,10 +131,18 @@ func describeCommand(p cli.Params) *cobra.Command {
...
 			if len(args) == 0 {
-				err = askPipelineName(opts, p)
+				pipelineNames, err := pipeline.GetAllPipelineNames(p)
 				if err != nil {
 					return err
 				}
+				if len(pipelineNames) == 1 {
+					opts.PipelineName = pipelineNames[0]
+				} else {
+					err = askPipelineName(opts, p)
```
vinamra28, line 141: Maybe we could pass the list of names to `askPipelineName` as an arg by adding a `[]string` param?
In `askPipelineName`, you could check if the list is empty to determine whether to make the call here to avoid doing it twice in some cases.

vinamra28 (changes requested):

vinamra28 (approved):
/lgtm

piyush-garg (approved):

PR #1189, opened 2020-09-28
Refactor PipelineRun and TaskRun Cancel Error Based on Condition Status

### LINKED ISSUE #1188: Refactor PipelineRun and TaskRun Cancel to Handle Cancelling Already Completed Run Using Status
As described in #1185, in the case where a PipelineRun or TaskRun has already completed, `cancel` commands should display an error instead stating the run can't be cancelled since it has completed (either by success or failure). Instead of using the current validation approaches for [PipelineRun](https://github.com/tektoncd/cli/blob/7168e571b304ed9715562054f88b1d37fbe2d5a2/pkg/cmd/pipelinerun/cancel.go#L79) and [TaskRun](https://github.com/tektoncd/cli/blob/7168e571b304ed9715562054f88b1d37fbe2d5a2/pkg/cmd/taskrun/cancel.go#L76), a simpler approach would be to decide whether to error out in this scenario [based on the condition status of the run](https://github.com/tektoncd/cli/pull/1185#discussion_r495055342).

As noted in #1185,  if the Condition Status is True or False then display the error. If the condition status in unknown then proceed with the cancellation.

danielhelfand:
Closes #1188 

Instead of validating based on the result of `formatted.Condition`, the validation should check whether the condition status of the run is `ConditionUnknown`, which indicates the PipelineRun or TaskRun is still running and can be cancelled. There needed to be some minor tweaks to unit tests which had condition statuses for running that were incorrect. 

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/cmd/pipelinerun/cancel_test.go
+++ b/pkg/cmd/pipelinerun/cancel_test.go
@@ -74,17 +74,59 @@ func Test_cancel_pipelinerun(t *testing.T) {
 	prName := "test-pipeline-run-123"
 
 	prs := []*v1alpha1.PipelineRun{
-		tb.PipelineRun(prName,
-			tb.PipelineRunNamespace("ns"),
-			tb.PipelineRunLabel("tekton.dev/pipeline", "pipelineName"),
-			tb.PipelineRunSpec("pipelineName",
-				tb.PipelineRunServiceAccountName("test-sa"),
-				tb.PipelineRunResourceBinding("git-repo", tb.PipelineResourceBindingRef("some-repo")),
-				tb.PipelineRunResourceBinding("build-image", tb.PipelineResourceBindingRef("some-image")),
-				tb.PipelineRunParam("pipeline-param-1", "somethingmorefun"),
-				tb.PipelineRunParam("rev-param", "revision1"),
-			),
-		),
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      prName,
+				Namespace: "ns",
+				Labels:    map[string]string{"tekton.dev/pipeline": "pipelineName"},
+			},
+			Spec: v1alpha1.PipelineRunSpec{
+				PipelineRef: &v1alpha1.PipelineRef{
+					Name: "pipelineName",
+				},
+				ServiceAccountName: "test-sa",
+				Resources: []v1alpha1.PipelineResourceBinding{
+					{
+						Name: "git-repo",
+						ResourceRef: &v1alpha1.PipelineResourceRef{
+							Name: "some-repo",
+						},
+					},
+					{
+						Name: "build-image",
+						ResourceRef: &v1alpha1.PipelineResourceRef{
+							Name: "some-image",
+						},
+					},
+				},
+				Params: []v1alpha1.Param{
+					{
+						Name: "pipeline-param-1",
+						Value: v1alpha1.ArrayOrString{
+							Type:      v1alpha1.ParamTypeString,
+							StringVal: "somethingmorefun",
+						},
+					},
+					{
+						Name: "somethingmorefun",
+						Value: v1alpha1.ArrayOrString{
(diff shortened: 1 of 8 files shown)


pkg/cmd/taskrun/cancel_test.go
```diff
@@ -62,17 +62,28 @@ func TestTaskRunCancel(t *testing.T) {
...
+			Spec: v1alpha1.TaskRunSpec{
+				TaskRef: &v1alpha1.TaskRef{
+					Name: "failure-task",
+				},
+			},
+			Status: v1alpha1.TaskRunStatus{
+				Status: duckv1beta1.Status{
+					Conditions: duckv1beta1.Conditions{
+						{
+							Status: corev1.ConditionUnknown,
```
danielhelfand, line 80: This should be changed to Unknown since it needs to make it through the validation of whether or not the TaskRun has finished or not.

pkg/cmd/taskrun/cancel_test.go
```diff
@@ -311,7 +322,7 @@ func TestTaskRunCancel_v1beta1(t *testing.T) {
 				Status: duckv1beta1.Status{
 					Conditions: duckv1beta1.Conditions{
 						{
-							Status: corev1.ConditionTrue,
+							Status: corev1.ConditionUnknown,
```
danielhelfand, line 325: This should be changed to Unknown since it needs to make it through the validation of whether or not the TaskRun has finished or not.

pkg/cmd/taskrun/cancel_test.go
```diff
@@ -242,7 +253,7 @@ func TestTaskRunCancel_v1beta1(t *testing.T) {
...
```
danielhelfand, line 256: This should be changed since the TaskRun is in a running state, so its Condition status should be unknown.

pkg/cmd/pipelinerun/cancel_test.go
```diff
@@ -74,17 +74,59 @@ func Test_cancel_pipelinerun(t *testing.T) {
...
+							StringVal: "revision1",
+						},
+					},
+				},
+			},
+			Status: v1alpha1.PipelineRunStatus{
+				Status: duckv1beta1.Status{
+					Conditions: duckv1beta1.Conditions{
+						{
+							Status: corev1.ConditionUnknown,
```
danielhelfand, line 123: This should be changed to Unknown since it needs to make it through the validation of whether or not the PipelineRun has finished or not.

pkg/cmd/pipelinerun/cancel_test.go
```diff
@@ -369,6 +411,16 @@ func Test_cancel_pipelinerun_v1beta1(t *testing.T) {
 					},
 				},
 			},
+			Status: v1alpha1.PipelineRunStatus{
```
danielhelfand, line 414: Adding status here since the test is for a successful cancel, which means the PipelineRun should be in a running state.

test/e2e/pipeline/pipeline_test.go
```diff
@@ -284,10 +284,6 @@ Waiting for logs to be available...
 	kubectl.MustSucceed(t, "create", "-f", helper.GetResourcePath("pipeline-with-workspace.yaml"))
 
 	t.Run("Start PipelineRun with --workspace and volumeClaimTemplate", func(t *testing.T) {
-		if tkn.CheckVersion("Pipeline", "v0.10.2") {
```
danielhelfand, line 287: It seems this runs appropriately without the version check. Are we still running these tests against a cluster with v0.10.2 installed?

danielhelfand: Addressed by https://github.com/tektoncd/plumbing/pull/596. Will add this back.

piyush-garg (commented):
/lgtm

pradeepitm12 (approved):
/lgtm

PR #1196, opened 2020-10-02
Update OSX instruction to install from brew 🍻

chmouel:
# Changes

Update OSX instructions to point to the official brew repo.

We still keep the upload to tektoncd/tools since it's easy to maintain (it's part of goreleaser) 
and if there is an issue merging brew update we can still point user at it in the future.

# Submitter Checklist

review them:

# Release Notes

```release-note
Brew is available from the official brew repository. You still can install tektoncd cli on OSX from the `tektoncd/tools` tap since both would be supported for the foreseeable future. 
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,11 +16,10 @@ Download the latest binary executable for your operating system:
 
 * Mac OS X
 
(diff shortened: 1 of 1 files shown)


vdemeester (approved):
/lgtm

PR #1227, opened 2020-10-29
Do Not Delete ClusterTask TaskRuns with --task Flag

danielhelfand:
When running `tkn tr delete` with the `--task` option, if a ClusterTask exists with the same name, the ClusterTask's TaskRuns will also be deleted. This is because of TaskRuns for ClusterTasks having the `tekton.dev/task` label, which [will be removed in January 2021](https://github.com/tektoncd/pipeline/issues/2533#issuecomment-623454245).

For now, the workaround is to filter the TaskRuns by the TaskRefKind, ignoring the run for the deletion process if its kind is `ClusterTask`. A follow up to this pr should be an option added to `tkn tr delete` to delete TaskRuns associated with a ClusterTask. 

# Submitter Checklist

# Release Notes

```release-note
Prevent tkn tr delete --task from deleting TaskRuns associated with ClusterTasks
```

### DIFF
--- a/pkg/task/tasklastrun.go
+++ b/pkg/task/tasklastrun.go
@@ -19,7 +19,6 @@ import (
 
 	"github.com/tektoncd/cli/pkg/cli"
 	trlist "github.com/tektoncd/cli/pkg/taskrun/list"
-	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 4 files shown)


pkg/task/tasklastrun.go
```diff
@@ -65,14 +64,3 @@ func FilterByRef(taskruns []v1beta1.TaskRun, kind string) []v1beta1.TaskRun {
 	}
 	return filtered
 }
-
-// this will filter the taskkrun which have reference to Task
```
danielhelfand, line 69: This isn't used anywhere in the CLI, so I removed it. Let me know if there are considerations around it I am not aware of.

piyush-garg: I am also not sure about this.

piyush-garg (approved):

pkg/task/tasklastrun.go
danielhelfand: /cc @pradeepitm12 Not sure if you are more familiar with this?

PR #1229, opened 2020-10-30
WIP: Add TaskRun Restart Command

### LINKED ISSUE #1091: Better UX for restarting taskrun/pipelinerun
### Feature request

Have an easy way to restart taskruns or pipelineruns

### Use case

A user can use the feature `--use-taskrun` to `tkn task` or the pipeline equivalent. That's not an obvious wayfor the user who wants to restart a taskrun/pipelinerun for something as common as restarting a pipelinerun/taskrun

### UI Example

Something like this : 

`tkn taskrun restart taskrun-abcd`

would do the same thing as 

`tkn task start taskrun --use-taskrun taskrun-abcd`

same goes for `pipelinerun`

### LINKED ISSUE #1119: Start a task from a taskrun with embedded task
### Feature request

I would like to be able to re-run a taskrun with embedded task spec.
Today this is not possible:
```
$ tkn task start --use-taskrun plumbing-yamllint-pr-complete-zkg94
Error: either a Task name or a --filename argument must be supplied
```

The `TaskRun` is defined here: https://github.com/tektoncd/plumbing/blob/bde34da7abdec4bd7f63a607d56828b5bdba5833/tekton/resources/ci/github-template.yaml#L26

### Use case

This is useful to re-run taskruns in Tekton CI.

### UI Example

I see two options:
- allow for the task name to be missing in the `--use-taskrun` command
- implement this as `tkn taskrun rerun <taskrun-name>`

In either case, fail if there is no embedded spec or the referenced task cannot be found.

danielhelfand:
Closes #1091 
Closes #1119 

This pull request adds the ability to create a new TaskRun based on a previous TaskRun. This is similar to `tkn task start --use-taskrun` with the key difference being that it will allow users the ability to create a new TaskRun without needing to have a TaskRef. This will allow support for starting TaskRuns that have TaskSpecs defined (#1119) and also allow users a simpler experience instead having to respecify several options associated with starting a Task.

```
tkn taskrun restart foo-12345
```

Some things to discuss:
* The name `restart` may be confusing as this command actually creates a new run and doesn't actually re-execute anything. Should it actually be just start or something else?
* Currently this implementation requires the TaskRun used to have the `generateName` property specified. For those who want to use a TaskRun without `generateName`, the user can use the `--prefix-name` option to specify a `generateName`.
* The current approach requires a TaskRun to already exist, but is there a better workflow as far as being able to start a TaskRun from a file definition or in a declarative way?

# Submitter Checklist

review them:

# Release Notes

```release-note
Add tkn taskrun restart command to create a new TaskRun using old TaskRun runtime values
```


chmouel:
@danielhelfand is this ready for review ?

danielhelfand:
> @danielhelfand is this ready for review ?

I am open to feedback on it, but I think last I left off with this I was wondering if it maybe made more sense to have a whole `tkn taskrun start` command versus only restart. I would prefer an approach here that supports what was being added in this pr along with a way to create new TaskRuns that are not from previous runs.

vdemeester:
> * The name `restart` may be confusing as this command actually creates a new run and doesn't actually re-execute anything. Should it actually be just start or something else?
> * Currently this implementation requires the TaskRun used to have the `generateName` property specified. For those who want to use a TaskRun without `generateName`, the user can use the `--prefix-name` option to specify a `generateName`.
> * The current approach requires a TaskRun to already exist, but is there a better workflow as far as being able to start a TaskRun from a file definition or in a declarative way?

@danielhelfand good questions. I feel if we were to go with `start` we could use a `from` (flag or notion) so that you could use a previous taskrun or a task to refer from a task. With restart I wonder how it would look like.

PR #1242, opened 2020-11-10
Allow --prefix-name and --timeout to be Overrode when using --last or --use-run Options

### LINKED ISSUE #1240: Override --timeout Value when Using --last/--use-taskrun/--use-pipelinerun

- Kubernetes version:

N/A

- Tekton Pipeline version:

N/A

- Operating System:

N/A

# Expected Behavior

When using the `--last`, `--use-taskrun`, and `--use-pipelinerun` with `start` commands, users should have the ability to override the value of `--timeout` using the option as shown below:

```
tkn t start taskName --last --timeout 1s
```

The above command would result in a TaskRun that uses all the previous runtime values of the last available TaskRun but also specifies a new `--timeout` that should be `1s`. If a previous timeout had been part of the last TaskRun, it should be replaced by the new `--timeout` value.

# Actual Behavior

The old `--timeout` value is still applied from the older run. 

1. Create a TaskRun from a Task (Any Task should work) using `tkn task start`
2. Create a new TaskRun using `tkn task start` with `--last` or `--use-taskrun` and add the `--timeout` value.
3. Observe that the timeout value is not the same as what was specified using the `--timeout` option by using `tkn tr desc --last` or `kubectl describe tr <taskrunname>`

# Additional Info

As part of working on this issue, it might be helpful to also assess whether other runtime values (e.g. params, workspaces, etc.) are also able to be overrode when using `--last`.

It should also be noted that this will apply to ALL `start` commands: `tkn clustertask`, `tkn task`, `tkn pipeline`. This fix for this issue should be applied to all to close this issue.

danielhelfand:
Closes #1240 

This pull request moves where `--prefix-name` and `--timeout` options are set in order to allow these values to be set when used with `--last` or `--use-taskrun` or `--use-pipelinerun` options.

# Submitter Checklist

# Release Notes

```release-note
Allow --prefix-name and --timeout to be set with --last and --use options
```

### DIFF
--- a/pkg/cmd/clustertask/start.go
+++ b/pkg/cmd/clustertask/start.go
@@ -210,20 +210,6 @@ func startClusterTask(opt startOptions, args []string) error {
 		return err
 	}
 
-	if opt.TimeOut != "" {
-		timeoutDuration, err := time.ParseDuration(opt.TimeOut)
-		if err != nil {
-			return err
-		}
-		tr.Spec.Timeout = &metav1.Duration{Duration: timeoutDuration}
-	}
-
-	if opt.PrefixName == "" {
-		tr.ObjectMeta.GenerateName = ctname + "-run-"
-	} else {
-		tr.ObjectMeta.GenerateName = opt.PrefixName + "-"
-	}
-
 	if opt.Last || opt.UseTaskRun != "" {
 		taskRunOpts := options.TaskRunOpts{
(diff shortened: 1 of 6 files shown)


piyush-garg (approved):

chmouel (approved):

chmouel:
/lgtm

Looking Good! 🤙🏽

Issue #1256, opened 2020-11-17
Debian package from ppa is not reporting properly the released version


chmouel:
```
chmouel@thebond 🐱 ~% dpkg -l tektoncd-cli|tee
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name           Version      Architecture Description
+++-==============-============-============-====================================================
ii  tektoncd-cli   0.13.1-0     amd64        A command line interface for interacting with Tekton
chmouel@thebond 🐱 ~% tkn version
Client version: 2020-10-21T10:47:53Z
```

piyush-garg:
Fixed with #1257

PR #1257, opened 2020-11-17
📦 Use VERSION file debian/versiion for debian package if we find it

chmouel:
Use debian/VERSION when building debian package

    It seems that when building on launchpad, we cannot detect the version
    by doing a curl. Perhaps they restrict network access or github API get
    restricted?

    We can't use just VERSION at the root dir since dpkg-source doesnt
    include it, but we can do that inside the debian/ dir so let's just do
    that.

Fixes bug #1256

```release-note
NONE
```

### DIFF
--- a/tekton/debbuild/container/buildpackage.sh
+++ b/tekton/debbuild/container/buildpackage.sh
@@ -37,8 +37,8 @@ cd tektoncd-cli-${version}
 # Make it easy for devs
 [[ -d /debian ]] && { rm -rf debian && cp -a /debian . ; }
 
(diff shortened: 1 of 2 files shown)


vdemeester (approved):

chmouel:
This change should work, tested it by uploading a package directly, see log (detected as 0.14.0) : 

https://launchpadlibrarian.net/507542670/buildlog_ubuntu-groovy-amd64.tektoncd-cli_0.14.0-6_BUILDING.txt.gz

/hold cancel

PR #1268, opened 2020-12-16
Drop cobra fork and rework the shell completion system

chmouel:
# Changes

Adapted the completion we had previously to the new completion system from
cobra 1.0.

I started to use the client-go directly but since there is no caching, the
<Tab><Tab> is not very usuable since super slow. So using the kuebctl binrary directly instead
(I can't use tkn since caching is not there as well) to get the
completions.

This is how we were doing it previously (using kubectl or tkn) and
completions should be much faster. And bonus, we have some proper completion for powershell/fish and others added for free. (not tested).

It doesn't pickup the completion on `tkn hub`, so we wil need to figure out on how to do this for commands that gets pulled up by deps.

Note: this PR has two diffsets on purpose, to make it easier to review/track
history. The split is between updating the new library and the second split to
add all completions to all commands.

# Submitter Checklist

review them:

# Release Notes

```release-note
The cobra fork we were using previously has been dropped and all completions
has been modified to use latest cobra shell completion system.
```


chmouel:
There is as way to get inline completion for flags, i.e: --service-account <TAB> to complete the serviceaccount as we had before, this is the example from the helm sources : 

https://github.com/helm/helm/blob/e71d38b414eb92912a87e6d1e751e88ee8fc57ea/cmd/helm/get_notes.go#L62-L71

Will make a further patch later on to address thoses as this will be another big one..

Ready for review this one to make it easier...

/hold cancel

chmouel:
We probably going to have to figure how to do this with `tkn hub` too...

vdemeester:
cc @mattmoor

pradeepitm12 (approved):
/lgtm

PR #1279, opened 2021-01-09
Support --clustertask Option for tkn tr delete

### LINKED ISSUE #1228: Support --clustertask Option for tkn tr delete
### Feature request

Similar to the `tkn tr delete --task` option, which deletes TaskRuns associated with a Task in a namespace, it would also be helpful to allow for the deletion of TaskRuns for a ClusterTask. Since TaskRuns are namespaced but ClusterTasks have a global scope, my thought is to only delete TaskRuns from a ClusterTask in a particular namespace. 

### Use case

There is no way to delete TaskRuns from a ClusterTask with `tkn tr delete`.

### UI Example

```
tkn tr delete --clustertask foo -n ns
```

vinamra28:
# Changes

Currently `tkn tr delete` doesn't support `--clustertask` flag. Due to this
if the user wants to delete the `taskruns` associated with a `clustertask`
was not possible.

This PR add `--clustertask` as flag and now user can delete all the
`TaskRun` from a particular namespace associated with a `ClusterTask`.
The command will filter the taskruns by label ie `tekton.dev/task` or
`tekton.dev/clusterTask`. Also if the user passes `--task` and
`--clustertask` together then it will give an error.

closes #1228

Signed-off-by: vinamra28 <vinjain@redhat.com>
# Submitter Checklist

review them:

for more details._

# Release Notes
```release-note
Add support for --clustertask option in tkn tr delete which allows to delete all the taskruns associated with a particular clustertask
```

### DIFF
--- a/pkg/cmd/taskrun/delete.go
+++ b/pkg/cmd/taskrun/delete.go
@@ -34,8 +34,14 @@ import (
 	cliopts "k8s.io/cli-runtime/pkg/genericclioptions"
 )
 
+type deleteOptions struct {
+	ClusterTaskName string
+	TaskName        string
+}
+
 func deleteCommand(p cli.Params) *cobra.Command {
-	opts := &options.DeleteOptions{Resource: "TaskRun", ForceDelete: false, ParentResource: "Task", DeleteAllNs: false}
+	opts := &options.DeleteOptions{Resource: "TaskRun", ForceDelete: false, DeleteAllNs: false}
+	deleteOpts := &deleteOptions{}
 	f := cliopts.NewPrintFlags("delete")
 	eg := `Delete TaskRuns with names 'foo' and 'bar' in namespace 'quux':
 
@@ -64,6 +70,18 @@ or
 				Err: cmd.OutOrStderr(),
 			}
 
+			if deleteOpts.TaskName != "" && deleteOpts.ClusterTaskName != "" {
+				return fmt.Errorf("cannot use --task and --clustertask option together")
+			}
+
+			if deleteOpts.ClusterTaskName != "" {
+				opts.ParentResource = "ClusterTask"
+				opts.ParentResourceName = deleteOpts.ClusterTaskName
+			} else {
+				opts.ParentResource = "Task"
+				opts.ParentResourceName = deleteOpts.TaskName
+			}
+
 			if opts.Keep < 0 {
 				return fmt.Errorf("keep option should not be lower than 0")
 			}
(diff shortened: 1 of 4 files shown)


pradeepitm12 (approved):

pkg/cmd/taskrun/delete.go
```diff
@@ -64,6 +70,22 @@ or
 				Err: cmd.OutOrStderr(),
 			}
 
+			switch {
+			case deleteOpts.ClusterTaskName != "":
+				opts.ParentResource = "ClusterTask"
+				opts.ParentResourceName = deleteOpts.ClusterTaskName
+			case deleteOpts.TaskName != "":
```
piyush-garg, line 77: we don't need this case

pkg/cmd/taskrun/delete.go
```diff
@@ -64,6 +70,22 @@ or
...
+				opts.ParentResource = "Task"
+				opts.ParentResourceName = deleteOpts.TaskName
+			default:
+				opts.ParentResource = "Task"
+				opts.ParentResourceName = deleteOpts.TaskName
+			}
+
+			if deleteOpts.TaskName != "" && deleteOpts.ClusterTaskName != "" {
```
piyush-garg, line 85: this case should be before switch

pkg/cmd/taskrun/delete.go
```diff
@@ -143,16 +167,25 @@ func deleteTaskRuns(s *cli.Stream, p cli.Params, trNames []string, opts *options
...
 		lOpts := metav1.ListOptions{
-			LabelSelector: fmt.Sprintf("tekton.dev/task=%s", taskName),
+			LabelSelector: fmt.Sprintf("tekton.dev/%s=%s", label, taskName),
 		}
 		trs, err := trlist.TaskRuns(cs, lOpts, p.Namespace())
 		if err != nil {
 			return nil, err
 		}
-		trs.Items = taskpkg.FilterByRef(trs.Items, string(v1beta1.NamespacedTaskKind))
+		if kind == "ClusterTask" {
```
piyush-garg, line 184: we need the filter only if the kind is Task

vinamra28: We need this for `ClusterTask` as well right since we need to filter the `TaskRuns` by `ClusterTask` also?

piyush-garg: No, we don't need it because taskruns associated with cluster tasks can be filtered by tekton.dev/clustertask and this will not be there for taskrun started by the task

vinamra28: ohh..okay...got it :+1:

piyush-garg (approved):
/lgtm

Issue #1264, opened 2020-11-24
Display version of task in tkn task ls


sm43:
### Feature request

It would be helpful to know which version of the task is installed. Currently, to know the version 
user has to see the YAML of the task.

### Use case

hub cli will soon have a feature to install/upgrade a task. so, to know which version is currently installed would
be helpful.

In the catalog, the version for a Task is defined as a label `app.kubernetes.io/version`

/kind feature

PuneetPunamiya:
Is this output fine for the above issue 
```
 ./tkn task list
NAME   VERSION   DESCRIPTION              AGE
az     0.1       This task performs ...   2 days ago
```

cc @SM43  @piyush-garg  @danielhelfand

piyush-garg:
> Is this output fine for the above issue
> 
> ```
>  ./tkn task list
> 
> NAME   VERSION   DESCRIPTION              AGE
> az     0.1       This task performs ...   2 days ago
> ```
> 
> cc @SM43 @piyush-garg @danielhelfand

This looks good, also handle the case if there is no version label/annotation

PuneetPunamiya:
> > Is this output fine for the above issue
> > ```
> >  ./tkn task list
> > 
> > NAME   VERSION   DESCRIPTION              AGE
> > az     0.1       This task performs ...   2 days ago
> > ```
> > 
> > 
> > cc @SM43 @piyush-garg @danielhelfand
> 
> This looks good, also handle the case if there is no version label/annotation

Yes but according to TEP a resource should have version

piyush-garg:
> > > Is this output fine for the above issue
> > > ```
> > >  ./tkn task list
> > > 
> > > NAME   VERSION   DESCRIPTION              AGE
> > > az     0.1       This task performs ...   2 days ago
> > > ```
> > > 
> > > 
> > > cc @SM43 @piyush-garg @danielhelfand
> > 
> > 
> > This looks good, also handle the case if there is no version label/annotation
> 
> Yes but according to TEP a resource should have version

But a user can create its own task or can install task from other sources

piyush-garg:
Also, I think we should add the same for ClusterTask also

PuneetPunamiya:
> > > > Is this output fine for the above issue
> > > > ```
> > > >  ./tkn task list
> > > > 
> > > > NAME   VERSION   DESCRIPTION              AGE
> > > > az     0.1       This task performs ...   2 days ago
> > > > ```
> > > > 
> > > > 
[...]

PR #1285, opened 2021-01-19
Adds version field while listing all the tasks

PuneetPunamiya:
- This patch list the task with the version of the task

   For e.g.

       ./tkn task list
       NAME                DESCRIPTION              VERSION   AGE
       ansible-tower-cli   Ansible-tower-cli t...   0.1       21 hours ago
       buildah                                      ---       20 hours ago 
      
Partially closes: #1264 

Signed-off-by: Puneet Punamiya <ppunamiy@redhat.com>

# Changes

# Submitter Checklist

review them:

for more details._

# Release Notes

### DIFF
--- a/pkg/cmd/task/list.go
+++ b/pkg/cmd/task/list.go
@@ -36,18 +36,18 @@ No Tasks found
 {{ else -}}
 {{- if not $.NoHeaders -}}
 {{- if $.AllNamespaces -}}
-NAMESPACE	NAME	DESCRIPTION	AGE
+NAMESPACE	NAME	DESCRIPTION	VERSION	AGE
 {{ else -}}
-NAME	DESCRIPTION	AGE
-{{ end -}} 
+NAME	DESCRIPTION	VERSION	AGE
+{{ end -}}
 {{- end -}}
 {{- range $_, $t := .Tasks.Items }}{{- if $t }}
 {{- if $.AllNamespaces -}}
-{{ $t.Namespace }}	{{ $t.Name }}	{{ formatDesc $t.Spec.Description }}	{{ formatAge $t.CreationTimestamp $.Time }}
-{{ else -}} 
-{{ $t.Name }}	{{ formatDesc $t.Spec.Description }}	{{ formatAge $t.CreationTimestamp $.Time }}
+{{ $t.Namespace }}	{{ $t.Name }}	{{ formatDesc $t.Spec.Description }}	{{ formatVersion $t.Labels }}	{{ formatAge $t.CreationTimestamp $.Time }}
+{{ else -}}
+{{ $t.Name }}	{{ formatDesc $t.Spec.Description }}	{{ formatVersion $t.Labels }}	{{ formatAge $t.CreationTimestamp $.Time }}
 {{ end }}{{- end }}{{- end }}
-{{- end -}} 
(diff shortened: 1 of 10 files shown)


pkg/cmd/task/list.go
```diff
@@ -36,18 +36,18 @@ No Tasks found
...
+{{ end -}}
 {{- end -}}
 {{- range $_, $t := .Tasks.Items }}{{- if $t }}
 {{- if $.AllNamespaces -}}
-{{ $t.Namespace }}	{{ $t.Name }}	{{ formatDesc $t.Spec.Description }}	{{ formatAge $t.CreationTimestamp $.Time }}
-{{ else -}} 
-{{ $t.Name }}	{{ formatDesc $t.Spec.Description }}	{{ formatAge $t.CreationTimestamp $.Time }}
+{{ $t.Namespace }}	{{ $t.Name }}	{{ formatDesc $t.Spec.Description }}	{{ formatVersion $t.Labels }}	{{ formatAge $t.CreationTimestamp $.Time }}
+{{ else -}}
+{{ $t.Name }}	{{ formatDesc $t.Spec.Description }}	{{ formatVersion $t.Labels }}	{{ formatAge $t.CreationTimestamp $.Time }}
```
vinamra28, line 48: can we also add a test in which a task has a version?

pkg/formatted/version.go
```diff
@@ -0,0 +1,22 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package formatted
+
+func FormatVersion(t map[string]string) string {
```
vinamra28, line 17: can we have a better variable instead of `t`?

test/builder/builder.go
```diff
@@ -295,8 +295,8 @@ func ListAllTasksOutput(t *testing.T, cs *framework.Clients, td map[int]interfac
 	t.Helper()
 	const (
 		emptyMsg = "No tasks found"
-		header   = "NAME\tDESCRIPTION\tAGE"
-		body     = "%s\t%s\t%s\n"
+		header   = "NAME\tDESCRIPTION\tVERSION\tAGE"
+		body     = "%s\t%s\t%s\t%s\n"
```
vinamra28, line 299: can we update the e2e tests also to have a version?

chmouel:
Most tasks that are not coming from hub (which i assume is most of the tasks user wrote) would not have the version field which would then always shows "---"

I am not sure I am too keen on this "---" showing all the time, is there a better way ?

PuneetPunamiya:
@chmouel  Thank you for the review, would it be fine if we show nothing instead of "---" as for the description as well we don't show anything if there is no description

chmouel:
i guess my issue is that the version label is not a top level API field while description it is i.e: users would hardly would know that we can have task versioning it's just a convention kind of thing.

So in practice only the tasks from the official catalog are going to have it. So for most people this is not going to be any useful and only show empty....

PuneetPunamiya:
Creating a different pr to add version field in describe command and not in list command
/close

Issue #1282, opened 2021-01-15
tkn completion foo exit=0


NissesSenap:
# Versions and Operating System

- Kubernetes version:

  **Output of `kubectl version`:**

```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.0", GitCommit:"ffd68360997854d442e2ad2f40b099f5198b6471", GitTreeState:"clean", BuildDate:"2020-11-18T13:35:49Z", GoVersion:"go1.15.0", Compiler:"gc", Platform:"linux/amd64"}
```

- Tekton Pipeline version:

  **Output of `tkn version` or `kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'`**

  ```
  Client version: 0.15.0

  ```

- Operating System:

  ```
  rhel 8.3
  ```

# Expected Behavior
tkn completion foo

Error: requires valid option got foo, expected zsh or bash

echo $?
1
# Actual Behavior
tkn completion foo

echo $?
0
# Steps to Reproduce the Problem

1. tkn completion foo
2. echo $?
3.

# Additional Info

rupalibehera:
We can close this issue now, the fix is merged.

chmouel:
thanks!

/close

NissesSenap:
Thanks for the quick response as always ☺️

PR #1287, opened 2021-01-20
tkn completion command will bail out if something unexpected is passed  

rupalibehera:
# Changes

- tkn completion command will bail out if something else is passed apart from expected args
- forced the number of args to be 1, passing multiple in valid options was returning 0 for eg: ( tkn completion f b)
- Modified unit test as required
issue https://github.com/tektoncd/cli/issues/1282

# Submitter Checklist

review them:

for more details._

# Release Notes

### DIFF
--- a/pkg/cmd/completion/completion.go
+++ b/pkg/cmd/completion/completion.go
@@ -27,6 +27,8 @@ interactive completion
 Supported Shells:
 	- bash
 	- zsh
+	- fish
+	- powershell
 `
 	eg = `To load completions:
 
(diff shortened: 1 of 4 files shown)


pkg/cmd/completion/completion.go
```diff
@@ -72,7 +75,7 @@ func Command() *cobra.Command {
 		Annotations: map[string]string{
 			"commandType": "utility",
 		},
-		Args: cobra.MinimumNArgs(1),
+		Args: cobra.ExactArgs(1),
```
piyush-garg, line 78: we can use `ValidArgs []string` here, we don't need default then

rupalibehera: Actually I was thinking if we use `Args: cobra.ExactValidArgs(1)` then we don't require the default block in the switch case and also explicit check of the len(args) before switch.

vdemeester: @rupalibehera what does the `ExactValidArgs` do ? aka how does it know which value are valid ?

rupalibehera: it checks if the input is present in the `ValidArgs: []string{"bash", "zsh", "fish", "powershell"}` and also check if the number of parameter is correct, so we don't have to explicitly check in the code.

chmouel:
/lgtm
Looks good to me, thanks! 👍

pkg/cmd/completion/completion.go
vdemeester: Oh nice, I like that 😉

chmouel:
FYI we usually would rather have the contributor rebasing their multiple commits as one before merging so that makes it easier to browse the github log!

rupalibehera:
> FYI we usually would rather have the contributor rebasing their multiple commits as one before merging so that makes it easier to browse the github log!

Yep Just did that :)

vdemeester (approved):
/meow

piyush-garg (approved):
/lgtm

/hold cancel

chmouel:
Good job @rupalibehera on your first tekton contribution  👍🏽

PR #1312, opened 2021-02-25
Introduce a tkn bundle push command to produce Tekton bundles

pierretasci:
# Changes

Introduces a `tkn bundle push` command to produce a valid Tekton bundle.

This change is part of TEP#31: https://github.com/tektoncd/community/blob/main/teps/0031-tekton-bundles-cli.md

This adds a `bundle` top level command and a `bundle push` subcommand which takes file or `stdin` input (similar to `kubectl apply -f -`) to produce a Tekton bundle from a set of valid YAML or JSON Tekton objects.

It supports:
- multiple objects in one file
- stdin input
- fully-qualified Tekton Bundle references
- bearer token or basic auth

# Submitter Checklist

review them:

for more details._

# Release Notes

```release-note
Adds a `tkn bundle push` command to generate valid Tekton Bundles
```


pkg/bundle/builder.go
```diff
@@ -0,0 +1,81 @@
...
+		}); err != nil {
+			return nil, err
+		}
+	}
+	return img, nil
+}
+
+// Return the ObjectMetadata.Name field which every resource should have.
+func getObjectName(obj runtime.Object) string {
+	return reflect.Indirect(reflect.ValueOf(obj)).FieldByName("ObjectMeta").FieldByName("Name").String()
```
vdemeester, line 80: This line scares me 😹

vdemeester (commented):
Question (on the design, without reviewing code) : 
- Is the `tkn push` additive or does it replace the full bundle ? as, can I incrementally (multiple commands) create my bundle, or should it be in a one-time operation ?

pkg/bundle/builder.go
pierretasci: You're right. I found a lib in apimachinery that does this for us so I will switch to using that.

pierretasci:
> * Is the `tkn push` additive or does it replace the full bundle ? as, can I incrementally (multiple commands) create my bundle, or should it be in a one-time operation ?

The design called for a single command with no "builder" like features so this is a one-time operation which will replace bundles with the same name so long as you have write access to the repository.

vdemeester (approved):

chmouel:
/lgtm

let's iterate from there, thanks

PR #1344, opened 2021-04-06
Fixes the bundle push and list commands to conform to the bundle spec.

pierretasci:
# Changes

The original implementation did not comply with the Tekton bundle spec which explcitely says the `kind` must be lowercased. We were forcing it upper-cased which is not legal. This PR fixes that bug and:
- allows the user to `list` with any of the uppercase/lowercase and/or singular/plural forms of a `kind` to make using this easier
- the push will set the correct annotation on the bundle object

To make this problem easier to diagnose, this change will be paired with a change to the controller to make the errors more obvious.

*Tested*
- ran the push, list locally and used a locally pushed image in a minikube cluster with `tekton:latest`

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/docs/cmd/tkn_bundle.md
+++ b/docs/cmd/tkn_bundle.md
@@ -24,6 +24,6 @@ Manage Tekton Bundles
 ### SEE ALSO
 
 * [tkn](tkn.md)	 - CLI for tekton pipelines
-* [tkn bundle list](tkn_bundle_list.md)	 - List a Tekton bundle's contents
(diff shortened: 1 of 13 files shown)


vinamra28 (commented):
@pierretasci can you please run `make generated` as there are some changes in the generated docs.
LGTM

chmouel:
same comment as @vinamra28 

/lgtm

vinamra28 (commented):
/lgtm

PR #1384, opened 2021-06-03
Replace maze.io/x/duration with its redirect

chmouel:
The redirection seems flaky, getting this at the moment :

```
% make test
Running unit tests...
go: github.com/tektoncd/hub/api@v0.0.0-20210517094448-c032b766a83c requires
	maze.io/x/duration@v0.0.0-20160924141736-faac084b6075: unrecognized import path "maze.io/x/duration": parse https://maze.io/x/duration?go-get=1: no go-import meta tags ()
make: *** [test-unit] Error 1
```

Trying to curl -iv on https://maze.io/x/duration?go-get=1 only gets a empty 200
with no redirection.

Sometime it may works but since it's flaky let's pin it (and it hasn't been
updated since 2016 anyway)

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -44,4 +44,5 @@ replace (
 	// https://github.com/kubernetes-sigs/kustomize/issues/1500
 	github.com/go-openapi/spec => github.com/go-openapi/spec v0.19.3
 	github.com/kr/pty => github.com/creack/pty v1.1.10
+	maze.io/x/duration => git.maze.io/go/duration v0.0.0-20160924141736-faac084b6075
 )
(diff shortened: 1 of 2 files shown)


piyush-garg:
Should we raise this on Hub CLI and use that here with changes

vdemeester (approved):

chmouel:
I haven't done it on hub, do you still need that pr @piyush-garg ?

piyush-garg:
> I haven't done it on hub, do you still need that pr @piyush-garg ?

Will go with this now for release, we can do the updated other way later, as it will require hub dep bump also

Issue #1388, opened 2021-06-08
prune resource  in cli


pradeepitm12:
### Feature request 
A way in tkn prune/delete resources with more options.
Maybe we can add a new `prune` command with multiple pruning  option 
or 
add more flags and param to pre-existing `delete` command 
### Use case
With tekton pipeline very soon user have hundreds of taskruns/pipelineruns in their cluster, user would like to prune some of it.
With this feature user will have an option to create a cron job with tkn cli and prune resource as required.

### UI Example

tkn taskrun prune/delete <--keep-younger-than=96h --keep-max=100 ... >

chmouel:
Ah I didn't realise if you were working on this, feel free to improve on #1392 after it gets merged..

vdemeester:
@chmouel @pradeepitm12 can we close this one ?

PR #1392, opened 2021-06-10
Add --keep-since flag to pipelinerun delete

chmouel:
# Changes

Add a `--keep-since=Minutes` flag to `tkn pipelinerun delete`.

This allows deleting all pipeline but the one younger than X minutes.

# Submitter Checklist

review them:

- [x] Commit messages follow [commit message best practices](https://github.com

# Release Notes

```release-note
add option --keep-since to tkn pr delete --all to allow keeping some prs that are younger than X minutes when deleting all pipelineruns. 
```

### DIFF
--- a/docs/cmd/tkn_pipelinerun_delete.md
+++ b/docs/cmd/tkn_pipelinerun_delete.md
@@ -33,6 +33,7 @@ or
   -f, --force                         Whether to force deletion (default: false)
   -h, --help                          help for delete
       --keep int                      Keep n most recent number of PipelineRuns
+      --keep-since int                When deleting all PipelineRuns keep the ones that has been completed since n minutes
   -o, --output string                 Output format. One of: json|yaml|name|go-template|go-template-file|template|templatefile|jsonpath|jsonpath-as-json|jsonpath-file.
   -p, --pipeline string               The name of a Pipeline whose PipelineRuns should be deleted (does not delete the Pipeline)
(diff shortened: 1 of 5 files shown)


vdemeester (approved):

vinamra28:
This PR might address #1388

chmouel:
maybe --keep-since is a better flag name 🤔

pradeepitm12:
> maybe --keep-since is a better flag name thinking

@chmouel  PERFECT :+1:

pradeepitm12:
How about renaming existing **--keep -> --keep-max** ?

chmouel:
> How about renaming existing **--keep -> --keep-max** ?

Renaming existing is not *free* backward compatibility and all, we may break a lot of users.. (altho we can do aliases but i don't think it's worth the pain)

vdemeester:
> > How about renaming existing **--keep -> --keep-max** ?
> 
> Renaming existing is not _free_ backward compatibility and all, we may break a lot of users.. (altho we can do aliases but i don't think it's worth the pain)

We could alias it ? But even if we want to do that, it should be in a different PR.

chmouel:
> We could alias it ? But even if we want to do that, it should be in a different PR.

I let this one up as an exercise to the reader 😏

pradeepitm12 (commented):
/lgtm

Issue #1395, opened 2021-06-14
Client version command doesn't work on 0.19.0 release


electrocucaracha:
# Versions and Operating System
- Tekton Pipeline version:
`0.19.0`

- Operating System:
```
NAME="Ubuntu"
VERSION="16.04.7 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.7 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial
```

# Expected Behavior

```bash
$ tkn version
Client version: 0.19.0
```

# Actual Behavior

```bash
$ tkn version
```
# Steps to Reproduce the Problem

1. Download [tarball](https://github.com/tektoncd/cli/releases/download/v0.19.0/tkn_0.19.0_Linux_x86_64.tar.gz)
2. Extract the binary
3. Execute `version` command

# Additional Info

tanwald:
Same here on Fedora Server:
```
$ tkn version --component client && cat /etc/os-release | sed -n 2p && uname -r && echo $TERM
VERSION="34 (Server Edition)"
5.12.12-300.fc34.x86_64
xterm-256color
```
but not on Fedora Workstation:
```
$ tkn version --component client && cat /etc/os-release | sed -n 2p && uname -r && echo $TERM
0.19.0
VERSION="34 (Workstation Edition)"
5.12.12-300.fc34.x86_64
xterm-256color
```

pradeepitm12:
Hey @anshulvermapatel,
Thanks for taking this up.
As we are planning for a v0.19.1 release this is something we would like to get in.
Just wanted to check if you are still on it?

vdemeester:
@tanwald both installed through the RPMs ? 🤔

pradeepitm12:
Fixed : https://github.com/tektoncd/cli/pull/1390
/close

tanwald:
> @tanwald both installed through the RPMs ? 🤔 
No, I have a script that compares my version with github releases and installs the binary. That script failed on the server edition because it couldn't get the installed version any longer.

anshulvermapatel:
Hi everyone, 

I apologize for replying late.
[...]

PR #1408, opened 2021-06-29
Fix client version output with --component option

anshulvermapatel:
Resolves [#1395](https://github.com/tektoncd/cli/issues/1395)

Resolved an issue where the client version with  `--component` option was being
printed same as the output of client version without adding `--component`

# Changes

# Submitter Checklist

review them:

for more details._

# Release Notes

<!--
Does your PR contain User facing changes?

If so, briefly describe them here so we can include this description in the
release notes for the next release!

For pull requests with a release note:

```release-note
Your release note here
```

For pull requests that require additional action from users switching to the new release, include the string "action required" (case insensitive) in the release note:

```release-note
action required: your release note here
```

For pull requests that don't need to be mentioned at release time, use the `/release-note-none` Prow command to add the `release-note-none` label to the PR. You can also write the string "NONE" as a release note in your PR description:


chmouel:
I think we need to test in this command, or move the logic block out of Command in a package in pkg/version for example to better make it testable.

pkg/cmd/version/version.go
```diff
@@ -114,7 +114,11 @@ func Command(p cli.Params) *cobra.Command {
 					fmt.Fprintf(cmd.OutOrStdout(), "Invalid component value\n")
 				}
 			} else {
-				fmt.Fprintf(cmd.OutOrStdout(), "Client version: %s\n", clientVersion)
+				if component == "" {
+					fmt.Fprintf(cmd.OutOrStdout(), "Client version: %s\n", clientVersion)
+				} else {
+					fmt.Fprintf(cmd.OutOrStdout(), "%s\n", clientVersion)
```
vdemeester, line 120: I don't think this is right though, because as is, `tkn version --component=pipeline` with a error connecting kubernetes, it will print the client version without an error, which doesn't make sense to me. 

We should print the client version **if** the component is `client`, otherwise we should either error out or print `unknown`.

vdemeester (commented):
One comment 👼🏼

pkg/cmd/version/version.go
anshulvermapatel: Yes, my bad.
I will correct it.

anshulvermapatel:
@chmouel , made the tests for this better .. still kept the block in cmd itself as it was a small change. Please do suggest once your review.

vdemeester (approved):

pkg/cmd/version/version_test.go
```diff
@@ -91,21 +91,84 @@ func TestVersionGood(t *testing.T) {
...
+				}
+				// To add multiple deployments in a particular namespace
+				for _, v := range tp.deployment {
+					if _, err := cls.Kube.AppsV1().Deployments(tp.namespace).Create(context.Background(), v, metav1.CreateOptions{}); err != nil {
+						t.Errorf("failed to create deployment: %v", err)
+					}
+				}
+				got, _ := test.ExecuteCommand(version, "version", "-n", "test", "--component", c)
+				fmt.Println(t.Name())
+				golden.Assert(t, got, fmt.Sprintf("%s-%s.golden", t.Name(), c))
```
vinamra28, line 167: instead of generating golden files can't we just do assert?

vdemeester: I tend to agree with @vinamra28, could be a follow-up though 😉

vdemeester (approved):

pkg/cmd/version/version_test.go
vinamra28: sure. I'll track this in an issue and merge this one :slightly_smiling_face:

PR #1433, opened 2021-08-10
add param for only return pipelinerun name

### LINKED ISSUE #1426: Add a flag to 'tkn pipeline start` to return PipelineRun name
### Feature request

A flag in which the `tkn pipeline start` return's just the name of the PipelineRun

### Use case

Currently trying to make a basic smoke test pipeline that will run through and call an existing pipeline multiple times in sequence with different parameters. Currently when you run the command you get a message that is nice for a user where it gives you a command for following the PipelineRun's log. 

But it would be nice if we could include a flag that would modify the output to just return the name of the PipelineRun so I could have a subsequent command for following the logs without having to pull the name using a regex expression to extract it.

liangyuanpeng:
# Changes

Add param for only return pipelinerun name.
fixed #1426 

# Submitter Checklist

review them:

for more details._

# Release Notes


piyush-garg:
Is this PR to have a flag to return the name of pipelinerun, when we start the pipeline? 

Can we think of a better flag name than `--return-name` ?

cc @vdemeester @chmouel @vinamra28

vinamra28:
> Is this PR to have a flag to return the name of pipelinerun, when we start the pipeline?
> 
> Can we think of a better flag name than `--return-name` ?
> 
> cc @vdemeester @chmouel @vinamra28

maybe `--output-pipeline-name`  :thinking:

chmouel:
we  already have -o what about -o=name ? like on kubectl ?

liangyuanpeng:
> we already have -o what about -o=name ? like on kubectl ?

OK, So we should be change `--return-name` to `-o`.

chmouel:
> OK, So we should be change --return-name to -o.

just to clarify it's `-o name` not just -o

there is already a -o yaml or -o json we just extend it with -o name

liangyuanpeng:
> This is great thanks, is there a way to get a test for this ? With golden hopefully,

Absolutely, Thanks for your patience !

piyush-garg:
@liangyuanpeng Can you please add tests for this and also squash the commits

liangyuanpeng:
Sorry for my late, i lose this issue, now i'm come back. @piyush-garg , Thanks for your remind.

> Can you please add tests for this and also squash the commits

Yes, i will.

chmouel:
any chance @liangyuanpeng to update this, there is quite a bit of use case that depends on it including for #1539 
if you like I can make a new PR based on your commit and add tests etc..

Issue #1407, opened 2021-06-29
Add --keep-since flag to taskrun delete


pradeepitm12:
### Feature request
add option --keep-since to tkn tr delete --all to allow keeping some trs that are younger than X minutes when deleting all taskruns. 

### Use case
this can help to make pruning much easier

chmouel:
Can we add --label at the same time please?

mbharatk:
Sure @chmouel will add that too.

PR #1435, opened 2021-08-11
Add --keep-since flag to taskrun delete

digglife:
# Changes

Add `--keep-since` flag to `taskrun` delete. Keep the ones that has been completed since n minutes when deleting all TaskRuns. 

This PR is basically a copy of #1392.

Issue: #1407 #1388 

# Submitter Checklist

review them:

# Release Notes

```release-note
add option --keep-since to tkn tr delete --all to allow keeping some trs that are younger than X minutes when deleting all taskruns. 
```

### DIFF
--- a/pkg/cmd/taskrun/delete.go
+++ b/pkg/cmd/taskrun/delete.go
@@ -17,6 +17,7 @@ package taskrun
 import (
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/tektoncd/cli/pkg/formatted"
 	taskpkg "github.com/tektoncd/cli/pkg/task"
@@ -90,6 +91,14 @@ or
 				opts.DeleteAllNs = true
 			}
 
+			if opts.Keep > 0 && opts.KeepSince > 0 {
+				return fmt.Errorf("cannot mix --keep and --keep-since options")
+			}
+
+			if opts.KeepSince > 0 && !opts.DeleteAllNs {
+				return fmt.Errorf("--keep-since option can only be used with --all")
+			}
+
 			if err := opts.CheckOptions(s, args, p.Namespace()); err != nil {
 				return err
 			}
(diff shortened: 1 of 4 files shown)


vdemeester (approved):
/cc @chmouel

pkg/cmd/taskrun/delete.go
```diff
@@ -90,6 +91,14 @@ or
 				opts.DeleteAllNs = true
 			}
 
+			if opts.Keep > 0 && opts.KeepSince > 0 {
+				return fmt.Errorf("cannot mix --keep and --keep-since options")
```
piyush-garg, line 95: should we add a check that --keep-since should not be less than 0?

pkg/cmd/taskrun/delete.go
```diff
@@ -180,20 +196,42 @@ func taskRunLister(p cli.Params, keep int, kind string, cs *cli.Clients) func(st
...
+func allTaskRunNames(cs *cli.Clients, keep, since int, ns string) ([]string, []string, error) {
+	var todelete, tokeep []string
+
 	taskRuns, err := trlist.TaskRuns(cs, metav1.ListOptions{}, ns)
 	if err != nil {
-		return nil, err
+		return todelete, tokeep, err
+	}
+
+	if since > 0 {
```
piyush-garg, line 212: cant we do like

```
if since > 0 {
		return keepTaskRunsByAge(taskRuns, since), nil
}
return keepTaskRunsByNumber(taskRuns, keep), nil

digglife: This is cleaner but the method is supposed to return 3 values.

digglife: Adding an extra nil to the return value of `keepTaskRunsByNumber` and `keepTaskRunsByAge` is doable though.

pradeepitm12 (approved):
/lgtm

PR #1436, opened 2021-08-12
Add --ignore-running flag with pipelinrun and taskrun delete commands.

liangyuanpeng:
# Changes

Fixed #1432 

Add --ignore-running flag with pipelinrun and taskrun delete commands.  

Ignore the running taskrun/pipelinerun when run delete command.

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [x] Run the code checkers with `make check`
- [x] Regenerate the manpages, docs and go formatting with `make generated`
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/cli/blob/master/CONTRIBUTING.md)
for more details._

# Release Notes

PR #1443, opened 2021-09-02
Delete runs using resource filter and keep-since

vinamra28:
# Changes

Deleting the Runs by filtering them with the help of their parent
resource and longer than x minutes. Sample UX

```bash=
$ tkn pr delete -p foobar --keep-since x
$ tkn tr delete -t foobar --keep-since x
$ tkn tr delete --clustertask foobar --keep-since x
```

closes #1442 
# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [X] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [X] Run the code checkers with `make check`
- [X] Regenerate the manpages, docs and go formatting with `make generated`
- [X] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/cli/blob/master/CONTRIBUTING.md)
for more details._

# Release Notes
```release-note
Delete PipelineRun or TaskRun by using --parent-resource and --keep-since together
```

PR #1445, opened 2021-09-13
Adds ignore-running flag to pr,tr delete cmd

pradeepitm12:
This is copy of #1436
Adds ignore-running to pipelinerun and taskrun delete command
this ignores all the running and pending resources while delete

# Changes

# Submitter Checklist

review them:

for more details._

# Release Notes


pradeepitm12:
cc: @vinamra28

pkg/cmd/pipelinerun/delete_test.go
```diff
@@ -239,7 +239,7 @@ func TestPipelineRunDelete(t *testing.T) {
 			input:       seeds[0].pipelineClient,
 			inputStream: strings.NewReader("y"),
 			wantError:   false,
-			want:        "Are you sure you want to delete all PipelineRuns related to Pipeline \"pipeline\" (y/n): All PipelineRuns associated with Pipeline \"pipeline\" deleted in namespace \"ns\"\n",
+			want:        "Are you sure you want to delete all PipelineRuns related to Pipeline \"pipeline\" (y/n): All PipelineRuns associated with Pipeline \"pipeline\" deleted in namespace \"ns\" ignores running or pending\n",
```
pradeepitm12, line 242: should we really consider adding `ignores running or pending`?

pradeepitm12:
/hold
while testing below fails 
```
./tkn pr delete  --ignore-running false --all -f
Error: --all flag should not have any arguments or flags specified with it
```
and 
```
./tkn pr delete  --ignore-running=false --all -f
All PipelineRuns deleted in namespace "default"
```

let's get this merge https://github.com/tektoncd/cli/pull/1443, as it takes care of some of the flags and its combination
cc: @piyush-garg @vinamra28

piyush-garg:
Do we need to mention `ignores running or pending` in output? I don't think so that we need to explicitly mention that. That can be flag description but not in output after success.

pradeepitm12:
> Do we need to mention `ignores running or pending` in output? I don't think so that we need to explicitly mention that. That can be flag description but not in output after success.

I should have written this with more description. I too feel the same. 
> should we really consider adding ignores running or pending?
The only thing is to ignore the running tr/pr is by default enabled. So even if the user does not specify the flag running ones will be ignored.

pradeepitm12:
I would really be happy to remove those extra details.

pkg/cmd/pipelinerun/delete_test.go
```diff
@@ -266,7 +266,7 @@ func TestPipelineRunDelete(t *testing.T) {
 			input:       seeds[4].pipelineClient,
 			inputStream: nil,
 			wantError:   false,
-			want:        "All but 2 PipelineRuns deleted in namespace \"ns\"\n",
+			want:        "All but 2 PipelineRuns(Completed) deleted in namespace \"ns\"\n",
```
pradeepitm12, line 269: @piyush-garg @vinamra28 
how about this  `(Completed)`

vinamra28: @pradeepitm12 there can be a case where the pipeline is in error state so should we show the same for that as well?

piyush-garg (approved):

pkg/cmd/pipelinerun/delete_test.go
piyush-garg: Completed mean both to me - either success or failure, but it has completed running

piyush-garg: cc @vdemeester @chmouel Any thoughts for this?

vinamra28:
LGTM

pkg/cmd/pipelinerun/delete_test.go
chmouel: Maybe "Finished" but it's nitpick...

either I am not so fond of having another switch and instead would rather have a filter capability...

i.e:

```
--ignore-filter running,label=key,anotherlabel=another=key,failed,completed
```

which would ignore all of this

we would need to include a `--list-filter` too.

piyush-garg: As discussed in WG yesterday, is it fine to with current implementation and refactor later based on other use cases cc @vdemeester @chmouel

PR #1467, opened 2021-10-22
Go compile darwin-arm64 in makefile for m1 macs

mathew-fleisch:
In this simple change I copied the existing format for cross-compiling to different system architectures to output binaries for m1 macs (Darwin-aarch64/arm64). It looks like the rest of the automation will pick this change up automatically in the next release. Please let me know if there is anything else I can do for this PR

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,6 +57,7 @@ arm:
 .PHONY: arm64
 arm64:
 	GOOS=linux GOARCH=arm64 go build -mod=vendor $(LDFLAGS) -o bin/tkn-linux-arm64 ./cmd/tkn
+	GOOS=darwin GOARCH=arm64 go build -mod=vendor $(LDFLAGS) -o bin/tkn-darwin-arm64 ./cmd/tkn
 
 .PHONY: s390x
 s390x:


chmouel:
goreleaser is not using the makefile so that would not get passed over,

it should already been generated from goreleaser since we have arm64 https://github.com/tektoncd/cli/blob/main/.goreleaser.yml

maybe we are using a too old goreleaser version? https://github.com/tektoncd/catalog/tree/main/task/goreleaser/0.2/

last time i updated that task it was end of december so it's not that old

i think a lot of people (including me on m1) are just using brew for this (but that's not a "solution" to force on users)

mathew-fleisch:
done 😄

piyush-garg:
As chmouel mentioned, the makefile is not getting used in release and `[tkn_0.22.0_Darwin_arm64.tar.gz](https://github.com/tektoncd/cli/releases/download/v0.22.0/tkn_0.22.0_Darwin_arm64.tar.gz)` also getting generated in releases

chmouel:
instead you need [universal binaries](https://goreleaser.com/customization/universalbinaries/) to generate binary compatible for m1 mac and x86 , see how i do it on pac here 

https://github.com/openshift-pipelines/pipelines-as-code/blob/main/.goreleaser.yml#L101

chmouel:
I have submitted #1528 for this

piyush-garg:
@chmouel As you have pushed the other change, its fine to merge this makefile change

chmouel:
Yep sgtm

PR #1530, opened 2022-03-14
Remove go11modules export instructions in README

chmouel:
it's been not needed since around go11 (since auto) and on by default
since go16

see https://maelvls.dev/go111module-everywhere/

## Release Note
```release-note
NONE
```
Signed-off-by: Chmouel Boudjnah <chmouel@redhat.com>


chmouel:
hub certificate is down so we are gettings thing like : 

image

🤦🏻

vdemeester (approved):

PR #1535, opened 2022-03-17
Add available plugins to the root help

chmouel:
# Changes

Add installed plugins to the root help. We define our own litle function
and reuse our already custom template to use it to show plugins.

Not trying to get the first line of the command to make it a pretty help
with description of the plugins but since we don't have a standardize
format around plugins for that.

We can perhaps standardize plugins who wants to implement it on a flag like : 

--short-help

and if we can run the command with that flag then maybe we add that short help to the plugin description.

but maybe that would bring more issues (ie: if it's a random script called tkn-foo and that tkn-foo does a lot of things the user would not want to be run).

anyoooooo at least the command name is a start...

cc @concaf

# Submitter Checklist

review them:

for more details._

# Release Notes

```release-note
plugins are not added to the root help command.
```

### DIFF
--- a/cmd/docs/generate.go
+++ b/cmd/docs/generate.go
@@ -132,6 +132,11 @@ func main() {
 	if err != nil {
 		fmt.Fprintln(os.Stderr, err.Error())
 	}
+	// Override PATH to make sure we don't get the plugins in PATH
+	oldEnv := os.Getenv("PATH")
+	os.Setenv("PATH", "")
+	defer os.Setenv("PATH", oldEnv)
+
 	fmt.Printf("Project root: %s\n", opts.source)
 	fmt.Printf("Generating yaml files into %s\n", opts.target)
 	if err := generateCliYaml(opts); err != nil {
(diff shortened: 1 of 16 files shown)


vdemeester (approved):

concaf (approved):

chmouel:
blog post describing the change :D (sorry for the shameless plug) https://blog.chmouel.com/2022/03/18/show-plugins-list-to-a-cli-when-using-gos-cobra-library/

pradeepitm12 (commented):
> LGTM

cmd/docs/generate.go
```diff
@@ -132,6 +132,11 @@ func main() {
 	if err != nil {
 		fmt.Fprintln(os.Stderr, err.Error())
 	}
+	// Override PATH to make sure we don't get the plugins in PATH
+	oldEnv := os.Getenv("PATH")
+	os.Setenv("PATH", "")
+	defer os.Setenv("PATH", oldEnv)
```
vinamra28, line 138: same here

chmouel:
@vinamra28  I refactored most of the plugins functions to pkg/plugins/ , we can't just reuse directly because they do different things but they share some code now,

vinamra28 (approved):
lgtm

PR #1708, opened 2022-09-08
Fix incomplete TaskRuns being deleted when --ignore-running=true

### LINKED ISSUE #1707: Incomplete TaskRuns deleted when --ignore-running=true

- Kubernetes version:

  **Output of `kubectl version`:**

  ```
  v1.21.14-gke.700
  ```

- Tekton Pipeline version:

  **Output of `tkn version` or `kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'`**

  ```
  v0.27.3
  ```

- Operating System:

  ```
  Alpine Linux v3.11
  ```

# Expected Behavior

Running the following command should avoid deleting TaskRuns that have not completed yet (because they are running or will run in the near future):

```
tkn tr delete -n some-namespace --keep 25 --ignore-running --force
```

# Actual Behavior

TaskRuns experiencing `ExceededResourceQuota` are unexpectedly removed, even though `Pending` TaskRuns are kept.

1. Set a low pod resource quota on your namespace
2. Fill up your TaskRun queue with lots of work
3. Wait for TaskRuns to experience `ExceededResourceQuota`
4. Run `tkn tr delete --ignore-running ...`
5. Watch as those not-yet-running TaskRuns get removed

# Additional Info

I have `tkn tr delete` running periodically to help minimize the total number of TaskRun objects in my cluster, as I don't need them for more than a few minutes after they have finished running.  Although `--ignore-running=true` seems like the correct way to do this, it has the unfortunate side effect of also removing TaskRuns that haven't run yet because we've temporarily hit our pod resource quota.

I suppose `--ignore-running=true` technically works in the strictest sense, as TaskRuns waiting on available quota are not actually in a "running" state.  But given that `--ignore-running=true` preserves `Pending` TaskRuns as seen here...

https://github.com/tektoncd/cli/blob/8aa0ead4a5b8de4d012cc40b486ee41c6d55971d/pkg/cmd/taskrun/delete.go#L293-L308

... I believe the actual intent of this flag is to only delete TaskRuns that have actually completed, and this behavior does not perfectly align with that understanding.

# Proposed Solution

Perhaps instead of relying on the `reason`, we could instead check whether `completionTime` is set?  According to the docs, this seems to only be set once a TaskRun has completed for any reason (including success, failed steps, validation errors, cancellation, timeouts, etc).  Those are the exact types of TaskRuns I don't want to keep around in my cluster.

Alternatively, we could expand the 'allowlist' of reasons to include `ExceededResourceQuota` and others, but that seems less robust.

I'd love to contribute to Tekton, so if you agree that a change should be made, please let me know which approach is preferred and I'll submit a PR accordingly :)

colinodell:
# Changes

Fixes #1707 by replacing the logic used to determine if a TaskRun is "running".

Prior to this PR, we used an allowlist of reasons like `Running`, `Pending`, and `Started`.  This did not account for [some states](https://tekton.dev/docs/pipelines/taskruns/#monitoring-execution-status) where the TaskRun is still running, like when cancellation has been requested (but not completed), or when the TaskRun encounters a non-permanent error and is still running (like `ExceededResourceQuota`).

This PR fixes this by instead checking for the presence of a `completionTime` which would indicate that the TaskRun is no longer in a running or about-to-run state.

Additionally, to help confirm these changes work as expected in the test suite, output messages that previously said something like `All TaskRuns were deleted` were updated to include the actual number of TaskRuns deleted.

# Submitter Checklist

review them:

for more details._

# Release Notes

```release-note
- Fix --ignore-running=true unexpectedly deleting some TaskRuns in incomplete states
- Ensure TaskRun deletion command always shows the number of deleted objects
```

### DIFF
--- a/pkg/cmd/taskrun/delete.go
+++ b/pkg/cmd/taskrun/delete.go
@@ -221,7 +221,7 @@ func deleteTaskRuns(s *cli.Stream, p cli.Params, trNames []string, opts *options
 			case opts.KeepSince > 0 && !opts.IgnoreRunning:
 				fmt.Fprintf(s.Out, "All but %d expired TaskRuns associated with %q %q deleted in namespace %q\n", numberOfDeletedTr, opts.ParentResource, opts.ParentResourceName, p.Namespace())
 			case opts.ParentResourceName != "" && !opts.IgnoreRunning:
-				fmt.Fprintf(s.Out, "All TaskRuns associated with %s %q deleted in namespace %q\n", opts.ParentResource, opts.ParentResourceName, p.Namespace())
+				fmt.Fprintf(s.Out, "All %d TaskRuns associated with %s %q deleted in namespace %q\n", numberOfDeletedTr, opts.ParentResource, opts.ParentResourceName, p.Namespace())
 			case opts.Keep > 0 && opts.KeepSince > 0:
 				fmt.Fprintf(s.Out, "%d TaskRuns(Completed) associated with %s %q has been deleted in namespace %q\n", numberOfDeletedTr, opts.ParentResource, opts.ParentResourceName, p.Namespace())
 			case opts.Keep > 0:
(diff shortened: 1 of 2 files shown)


vdemeester (commented):
/ok-to-test
/cc @pradeepitm12 @piyush-garg

piyush-garg:
@colinodell PR looks good, can you please squash the commits into 1

/lgtm

colinodell:
Thanks! I've squashed the commits as requested :)

pradeepitm12 (approved):
/lgtm

Issue #1738, opened 2022-10-07
replacements in .goreleaser.yml preventing CI automation


T0MASD:
Hi, I'm raising this to review https://github.com/tektoncd/cli/blame/main/.goreleaser.yml#L21 the replacements:
```
    darwin: Darwin
    linux: Linux
    windows: Windows
    386: i386
    amd64: x86_64
```
This is preventing url generation on CI builds like I have attempted here: 
https://github.com/tektoncd/plumbing/pull/1223/files#diff-c2315bb01874a1f74a8b6e38a5612b842595974e893145a06365f4db4826f574R20

`GOARCH=$(go env GOARCH) \
    GOOS=$(go env GOOS) \
    && wget -O- https://github.com/tektoncd/cli/releases/download/v${TKN_VERSION}/tkn_${TKN_VERSION}_${GOOS}_${GOARCH}.tar.gz | tar zxf - -C /usr/local/bin`

Is there a reason to make those replacements? What are the options to reliably generate multi arch (arm, x86 etc.) downloads for CI builds?

vdemeester:
cc @chmouel

chmouel:
@vdemeester i think you are the one who added this first from a copy pasta i guess 🙃 

I don't think there is a reason but whatever we do we may be careful changing binary names for distro/people's scripts out there grabbing new versions via regexp on release page...

chmouel:
maybe grab the scripts via the api  instead ?

T0MASD:
Thanks for your reply @chmouel, it makes sense to keep the existing format where binaries are named following convention in `replacements`. There's a one issue though, replacements are missing for arm architecture causing the lookup method (uname -m) that is used on x86 arch to fail. please see https://github.com/tektoncd/plumbing/blob/main/tekton/images/tkn/Dockerfile#L19
I will propose a change in MR for review

PR #1745, opened 2022-10-10
Update .goreleaser.yml

T0MASD:
setup replacement for arm64 to aarch64 to be in line with replacements for x86 arch

this fixes https://github.com/tektoncd/plumbing/issues/1222 and https://github.com/tektoncd/cli/issues/1738 and https://github.com/tektoncd/pipeline/issues/5568

this is potentially breaking change as build arm binaries will use different url format

/hold

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -24,6 +24,7 @@ archives:
     windows: Windows
     386: i386
     amd64: x86_64
+    arm64: aarch64
   format_overrides:
     - goos: windows
       format: zip


vdemeester:
@T0MASD do you know if there is a way to have both ? Both `arm64` and `aarch64` in the release pages. If not, we can warn users that starting with v0.28 the ARM url will change.

chmouel:
aarch64 is the standard name isnt it ?

on my rpi4 running fedora on arm64 i get aarch64 : 

% uname -r
5.19.13-300.fc37.aarch64

and my macbook call itself arm64:

```
% uname -a
Darwin snitch 21.6.0 Darwin Kernel Version 21.6.0: Wed Aug 10 14:28:35 PDT 2022; root:xnu-8020.141.5~2/RELEASE_ARM64_T8101 arm64 arm Darwin
```

altho both report "arm64" as uname -m so it probably the correct fix....

I don't think tektoncd cli on arm64[1] was super popular to be worried about backward compatibilty but your call @T0MASD or @vdemeester 

[1] mac users would use brew or tkn_0.27.0_Darwin_all.tar.gz binary

/lgtm

vinamra28 (approved):
@T0MASD feel free to remove the hold if there are no more changes

PR #2127, opened 2023-09-17
Bump goreleaser version and optimize get-version

vinamra28:
# Changes

With go version v1.20.7, goreleaser version also need to be upgraded to v1.15.2.

Also in get-version task we were running two steps and both the steps were using same image, hence merged both into a single step. Also the task was using goreleaser image which is more than 300M in size, instead using alpine/git which is ~21M

Also use tkn hub install to install tasks from catalog so that we always use the latest one available

----

With goreleaser version v1.14.0, archives.replacements and
nfpms.replacements have been deprecated and we can leverage their
respective name_template and file_name_template. Also the `replacements`
property has been removed in v1.19.0.

Updating goreleaser.yaml as per the suggested approach specified at
https://goreleaser.com/deprecations/#archivesreplacements

Signed-off-by: vinamra28 <jvinamra776@gmail.com>

/kind misc

# Submitter Checklist

review them:

- [NA] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [NA] Run the code checkers with `make check`
- [NA] Regenerate the manpages, docs and go formatting with `make generated`

for more details._

# Release Notes
```release-note
NONE
```

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -17,14 +17,14 @@ builds:
   ldflags:
   - -w -X github.com/tektoncd/cli/pkg/cmd/version.clientVersion={{.Version}}
 archives:
-- name_template: "{{ .Binary }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}"
-  replacements:
-    darwin: Darwin
-    linux: Linux
-    windows: Windows
-    386: i386
-    amd64: x86_64
-    arm64: aarch64
+- name_template: >-
+    {{- .Binary }}_
+    {{- .Version }}_
+    {{- title .Os }}_
+    {{- if eq .Arch "amd64" }}x86_64
+    {{- else if eq .Arch "386" }}i386
+    {{- else if eq .Arch "arm64" }}aarch64
+    {{- else }}{{ .Arch }}{{ end }}
   format_overrides:
     - goos: windows
       format: zip
@@ -41,11 +41,10 @@ changelog:
     - Merge pull request
     - Merge branch
 release:
-  prerelease: true
+  prerelease: "true"
 brews:
-  -
-    name: tektoncd-cli
-    tap:
+  - name: tektoncd-cli
+    repository:
       owner: tektoncd
       name: homebrew-tools
     folder: Formula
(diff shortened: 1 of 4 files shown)


.goreleaser.yml
```diff
@@ -17,14 +17,14 @@ builds:
...
-    darwin: Darwin
-    linux: Linux
-    windows: Windows
-    386: i386
-    amd64: x86_64
-    arm64: aarch64
+- name_template: >-
+    {{- .Binary }}_
+    {{- .Version }}_
+    {{- title .Os }}_
```
piyush-garg, line 23: do we need to do title here ?

.goreleaser.yml
```diff
@@ -17,14 +17,14 @@ builds:
...
+    {{- if eq .Arch "amd64" }}x86_64
```
piyush-garg, line 24: are all the cases handled ?

piyush-garg: i.e.
```
{{- else if eq .Arch "darwin" }}MacOS
{{- else if eq .Arch "linux" }}Linux
{{- else if eq .Arch "windows" }}Windows
```

.goreleaser.yml
```diff
@@ -62,7 +61,15 @@ brews:
       (zsh_completion/"_tkn").write output
       prefix.install_metafiles
 nfpms:
-  - file_name_template: "tektoncd-cli-{{.Version}}_{{.Os}}-{{.Arch}}"
+  - file_name_template: >-
+      tektoncd-cli-
+      {{- .Version }}_
+      {{- title .Os }}-
```
piyush-garg, line 67: same as above

.goreleaser.yml
```diff
@@ -62,7 +61,15 @@ brews:
...
+      {{- if eq .Arch "amd64" }}64bit
+      {{- else if eq .Arch "386" }}32bit
+      {{- else if eq .Arch "arm" }}ARM
+      {{- else if eq .Arch "arm64" }}ARM64
```
piyush-garg, line 71: same as above
```
{{- else if eq .Arch "darwin" }}MacOS
    {{- else if eq .Arch "linux" }}Linux
    {{- else if eq .Arch "windows" }}Windows
```

.goreleaser.yml
vinamra28: since Os value can be one of `darwin`, `linux` or `windows` and we had a replacements for the same
```
darwin: Darwin
linux: Linux
windows: Windows
``` 
so in order to replicate the same, added `title .Os`, do let me know if we are fine with having them in lower case

.goreleaser.yml
vinamra28: I believe these are not possible values for Arch, right?

Issue #1474, opened 2021-12-14
Create taskrun/pipelinerun from oci bundle


vdemeester:
### Feature request

It should be possible to create a `TaskRun` (using `tkn task start`) from a oci bundle.
Today, we can create a `TaskRun` from an in-cluster `Task` definition. If oci bundles are enabled on a Pipeline instance, it's possible to create a `TaskRun` that uses the oci bundle image reference as `Task` definition reference. It should be possible to do that from the cli directly

The same should be possible for `PipelineRun` as well

### Use case

Use oci bundle from the cli 👼🏼 

### UI Example

```
$ tkn task start oci://quay.io/vdemeest/my-task:0.1 -p foo=bar …
```

vdemeester:
Closing as I think it got implemented by https://github.com/tektoncd/cli/pull/2130.

PR #2130, opened 2023-09-21
Use oci image to run a task 

williamlfish:
# Changes

update the `task start` cmd to also allow an oci bundle with a task name. 
[Issue link here](https://github.com/tektoncd/cli/issues/1474)

create an argument for `--image or -i` that can pull an oci image and run a task from that bundle, still takes a positional argument as the task name

# Submitter Checklist

review them:

for more details._

# Release Notes


williamlfish:
@vdemeester the issue mentions creating a pipelinerun as well, since the logic is essentially the same, it might make sense to just bundle that logic and build the resource off the argument's and a "kind" param. I can add that to this pr, but was worried it would get much bigger, and wanted a chance to get approval on the approach. I can update this pr to do both and with more general func exported from somewhere ( not sure exactly where yet 😅 ) or just add that on a subsequent pr and update this cmd on the switch  ( see comment on the code ). lemme know your thoughts. thanks!

pkg/cmd/task/start.go
```diff
@@ -277,28 +300,59 @@ func startTask(opt startOptions, args []string) error {
 			Namespace: opt.cliparams.Namespace(),
 		},
 	}
-
 	var tname string
-	if len(args) > 0 {
+
+	switch {
```
williamlfish, line 305: here is where i think we can generalize the logic for a task and a pipeline? @vdemeester

pkg/cmd/task/start.go
```diff
@@ -18,12 +18,17 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"fmt"
-	"net/http"
 	"os"
 	"strings"
 	"time"
 
+	"github.com/google/go-containerregistry/pkg/name"
```
piyush-garg, line 25: not goimported

vdemeester: @williamlfish it should "auto-install" it in the `.bin` folder on the `make lint` commands (or whatever the target name is). At least that's what it's supposed to do 😛

williamlfish: @vdemeester  I updated `make lint` to do that with the latest commit. I'm not sure what version or what binary `yamllint` is supposed to be, but we should have that auto install too, for now i just moved to to the end so it fails there if someone does not have it installed 🤷‍♀️

piyush-garg:
lgtm

vdemeester (approved):
/meow

PR #2133, opened 2023-09-22
Reproducible/deterministic bundles

### LINKED ISSUE #2132: Reproducible/deterministic bundles
### Feature request

Currently if I run `tkn bundle push` with the same set files I will get two different images (different layer and manifest digests). This is due to the bundle builder [setting the time to the current time](https://github.com/tektoncd/cli/blob/e7befef75e491cc1ac13b6fd1a61b10aee74480d/pkg/bundle/builder.go#L88) and because I might specify different ordering when providing the files.
It would be beneficial for caching and provenance to have the option to build the bundle that is reproducible, i.e. deterministic based on the inputs.

I propose adding a command line parameter (e.g. `--ctime`) to set the created at time and to sort the layers of the manifest based on their digest by default. I think this should lead to reproducible/deterministic bundle builds.

Happy to contribute a pull request for this.

### Use case

Some argumentation for reproducible builds can be found on https://reproducible-builds.org/. It would also be useful for caching that the image is unchanged unless the files composing the image have changed.

### UI Example

For example if I run:

```shell
$ tkn bundle push -f one.yaml -f two.yaml registry.io/repository/bundle:latest --ctime 1970-01-01
$ tkn bundle push -f two.yaml -f one.yaml registry.io/repository/bundle:latest --ctime 1970-01-01
```

The `registry.io/repository/bundle:latest` would be unchanged by the second invocation.

zregvart:
# Changes

Sorts the bundle image layers according to the digest of the layer's content and adds the `--ctime` parameter to `tkn bundle push` to allow setting the created time in the bundle image config.

# Submitter Checklist

review them:

for more details._

# Release Notes

```release-note
Tekton bundle images are reproducible with --ctime provided to tkn bundle push
```

Fixes #2132

### DIFF
--- a/docs/cmd/tkn_bundle_push.md
+++ b/docs/cmd/tkn_bundle_push.md
@@ -30,6 +30,7 @@ Input:
 
 ```
       --annotate strings         OCI Manifest annotation in the form of key=value to be added to the OCI image. Can be provided multiple times to add multiple annotations.
+      --ctime string             YYYY-MM-DD, YYYY-MM-DDTHH:MM:SS or RFC3339 formatted created time to set, defaults to current time. In non RFC3339 syntax dates are in UTC timezone.
   -f, --filenames strings        List of fully-qualified file paths containing YAML or JSON defined Tekton objects to include in this bundle
   -h, --help                     help for push
       --remote-bearer string     A Bearer token to authenticate against the repository
--- a/docs/man/man1/tkn-bundle-push.1
+++ b/docs/man/man1/tkn-bundle-push.1
@@ -46,6 +46,10 @@ Input:
     OCI Manifest annotation in the form of key=value to be added to the OCI image. Can be provided multiple times to add multiple annotations.
 
 .PP
+\fB\-\-ctime\fP=""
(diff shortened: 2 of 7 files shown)


vdemeester (approved):

chmouel:
can it be the default? do we need a flag for different behavior?

zregvart:
> can it be the default? do we need a flag for different behavior?

@chmouel do you mean set the created time to a predefined constant (say 1970-01-01?) instead of current time? Should be easy to do, I'm guessing there was a reason to setting it to the current time previously and I didn't want to change that.

vdemeester:
> > can it be the default? do we need a flag for different behavior?
> 
> @chmouel do you mean set the created time to a predefined constant (say 1970-01-01?) instead of current time? Should be easy to do, I'm guessing there was a reason to setting it to the current time previously and I didn't want to change that.

I think I tend to agree with @chmouel, we could have it enabled by default. `tkn bundles` is an experimental set of subcommands, so we can "break" usage a little bit (even if, in that case, it shouldn't really break anything).

zregvart:
I created a followup in #2135

Issue #2135, opened 2023-09-27
Build deterministic/reproducible bundles by default


zregvart:
### Feature request

With #2132 the `tkn bundle push` command will create a deterministic/reproducible OCI image if `--ctime` parameter is provided and set to the same value.

A [comment](https://github.com/tektoncd/cli/pull/2133#issuecomment-1734404610) on the pull request implementing that suggested that building deterministic/reproducible OCI image should be the default.

Given that the default created time could be set to a fixed value by default, say 1970-01-01T00:00:00Z.

### Use case

It helps to be a bit more correct by default

### UI Example

Instead of `tkn bundle push --ctime '1970-01-01T00:00:00Z' -f...` one could use `tkn bundle push -f...`.

PR #2137, opened 2023-09-28
Support for `SOURCE_DATE_EPOCH` in `tkn bundle push`

### LINKED ISSUE #2136: Support `SOURCE_DATE_EPOCH` in `tkn bundle push`
### Feature request

Getting the created time from `SOURCE_DATE_EPOCH` is an encouraged standard, see https://reproducible-builds.org/specs/source-date-epoch/. `tkn bundle push` should support it.

### Use case

This helps in wider build pipelines when folk define `SOURCE_DATE_EPOCH` for multiple build tools.

### UI Example

With this, the `--ctime` would get set to 1970-01-01T00:00:00Z:

```sh
$ export SOURCE_DATE_EPOCH=0
$ tkn bundle push -f ...
```

zregvart:
# Changes

First commit is a refactoring commit to move the code from `PreRunE` to `parseArgsAndFlags`, this is something that in full honesty should have been done on #2133. This way is easier to test the implementation.

Second commit introduces support for `SOURCE_TIME_EPOCH` environment variable in `tkn bundle push`.

Fixes #2136 

# Submitter Checklist

review them:

for more details._

# Release Notes

```release-note
Created time in OCI Image Configuration in `tkn bundle push` can be provided using the SOURCE_TIME_EPOCH environment variable
```

### DIFF
--- a/pkg/cmd/bundle/push.go
+++ b/pkg/cmd/bundle/push.go
@@ -14,6 +14,7 @@
 package bundle
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -22,6 +23,7 @@ import (
 	"time"
 
 	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/jonboulle/clockwork"
 	"github.com/spf13/cobra"
 	"github.com/tektoncd/cli/pkg/bundle"
 	"github.com/tektoncd/cli/pkg/cli"
@@ -92,7 +94,7 @@ Created time:
 				Err: cmd.OutOrStderr(),
 			}
 
-			return opts.Run(args)
+			return opts.Run(cmd.Context(), args)
 		},
 	}
 	c.Flags().StringSliceVarP(&opts.bundleContentPaths, "filenames", "f", []string{}, "List of fully-qualified file paths containing YAML or JSON defined Tekton objects to include in this bundle")
@@ -105,7 +107,7 @@ Created time:
 
 // Reads the positional arguments and the `-f` flag to fill in the `bunldeContents` parameter with all of the raw Tekton
 // contents.
-func (p *pushOptions) parseArgsAndFlags(args []string) (err error) {
+func (p *pushOptions) parseArgsAndFlags(ctx context.Context, args []string) (err error) {
 	p.ref, _ = name.ParseReference(args[0], name.StrictValidation, name.Insecure)
 
 	// If there are file paths specified, then read them and include their contents.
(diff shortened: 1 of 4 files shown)


chmouel:
any chance you can explain why this is needed ? and what is this needed to be a flag ? is it expected to be used by a large number of users?

i am afraid having more and more flags in CLI with no clear explanation of what it does will just end up confusing and a burden for the maintainers.

zregvart:
> any chance you can explain why this is needed ? and what is this needed to be a flag ? is it expected to be used by a large number of users?

This is not adding another flag but allowing the existing flag (`--ctime`) to be specified with the `SOURCE_DATE_EPOCH` environment variable. This is in accordance with the [specification](https://reproducible-builds.org/specs/source-date-epoch/) from reproducible builds project.

> i am afraid having more and more flags in CLI with no clear explanation of what it does will just end up confusing and a burden for the maintainers.

Oh, for sure, but we're not adding one here :)

chmouel:
> Oh, for sure, but we're not adding one here :)

ah i misread the PR sorry about that, this is fine by me if it's following the spec

/ok-to-test

chmouel:
but another thing, what about making ctime the default? as discussed on another issue i don't believe there is breakage that could happen by changing the behaviour...

zregvart:
> but another thing, what about making ctime the default? as discussed on another issue i don't believe there is breakage that could happen by changing the behaviour...

Yeah, I plan on doing that, chose this to do first because of the refactoring that I did here. The followup in #2135 should be just a couple of lines following this PR :)

zregvart:
I see the tests fail, I'm going to investigate a bit, most likely something to do with the time zones...

zregvart:
Took me a minute to realize that the [test runner is setting](https://github.com/tektoncd/plumbing/blob/686a1d369c9f93cbf8303aa1765ac5a9699944b5/tekton/images/test-runner/runner.sh#L87C1-L88) `SOURCE_DATE_EPOCH`, and that interfered with the tests.

pkg/cmd/bundle/push_test.go
```diff
@@ -50,6 +50,13 @@ var (
 	}
 )
 
+func init() {
+	// fixed time to ease testing
+	now = func() time.Time {
+		return time.Date(2023, 9, 22, 1, 2, 3, 0, time.UTC)
```
chmouel, line 56: we usually use fakeclock/clockwork for that

chmouel:
/lgtm

Thanks!

pkg/cmd/bundle/push.go
```diff
@@ -22,6 +23,7 @@ import (
 	"time"
 
 	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/jonboulle/clockwork"
```
chmouel, line 26: thanks addressing other part of the code

PR #2139, opened 2023-10-09
Build deterministic bundles by default

### LINKED ISSUE #2135: Build deterministic/reproducible bundles by default

zregvart:
# Changes

Sets the default created time to 1970-01-01T00:00:00Z (Unix epoch) instead of the current time.

# Submitter Checklist

review them:

# Release Notes

```release-note
Default created time of Tekton OCI bundles is set to Unix epoch (1970-01-01T00:00:00Z)
```

Fixes #2135

### DIFF
--- a/docs/cmd/tkn_bundle_push.md
+++ b/docs/cmd/tkn_bundle_push.md
@@ -26,7 +26,7 @@ Input:
 	Valid input in any form is valid Tekton YAML or JSON with a fully-specified "apiVersion" and "kind". To pass multiple objects in a single input, use "---" separators in YAML or a top-level "[]" in JSON.
 
 Created time:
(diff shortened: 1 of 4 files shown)


vdemeester (approved):
/ok-to-test
/cc @chmouel

PR #2216, opened 2024-01-17
support v1 for sign command

Yongxuanzhang:
# Changes

# Submitter Checklist

review them:

for more details._

# Release Notes


docs/cmd/tkn_pipeline_sign.md
```diff
@@ -36,6 +36,7 @@ or using kms
   -o, --output string                 Output format. One of: (json, yaml, name, go-template, go-template-file, template, templatefile, jsonpath, jsonpath-as-json, jsonpath-file).
       --show-managed-fields           If true, keep the managedFields when printing objects in JSON or YAML format.
       --template string               Template string or path to template file to use when -o=go-template, -o=go-template-file. The template format is golang templates [http://golang.org/pkg/text/template/#pkg-overview].
+  -v, --version string                apiVersion of the Pipeline to be signed (default "v1")
```
chmouel, line 39: usually on all clis, -v mean --verbose or to see the version of the tool, not to set version 

what about having --pipeline-version, this probably not going to be used often and at least until v1beta1 is completely deprecated./..

Yongxuanzhang: Thanks that's a great suggestion! 😄

pkg/cmd/pipeline/sign.go
```diff
@@ -70,7 +73,13 @@ or using kms
 				return err
 			}
 
-			crd := &v1beta1.Pipeline{}
+			var crd metav1.Object
```
piyush-garg, line 76: We are reading the file provided by user as input, we can get the info about version from file content. Why we need to take the input from user?

Also --pipeline-version flag does not look make sense for task commands

chmouel: naming is hard 😅 I think the idea is to convey which tektoncd/pipeline version we want... 

but yeah that may be confusing?

any other suggestions? --crd-version perhaps ?

Yongxuanzhang: Thanks for the suggestion, but I don't know what would be the best way to do this. 
How to read the version info from the file content? Scan the string? How to get the version `v1` for this case?
e.g.
```yaml
# apiVersion: v1beta1
apiVersion: v1
...
```

Yongxuanzhang: --pipeline-version means (Tekton) Pipeline Version, it was suggested in the previous comment of this PR. Maybe rename it to --api-version?Wdyt?

vdemeester:
@Yongxuanzhang seems like it needs another rebase 🙏🏼

Yongxuanzhang:
> @Yongxuanzhang seems like it needs another rebase 🙏🏼

Just rebased, sorry for the delay 🙏 
I was focusing on something else in the past months

piyush-garg:
> > @Yongxuanzhang seems like it needs another rebase 🙏🏼
> 
> Just rebased, sorry for the delay 🙏 I was focusing on something else in the past months

Did you try with some approach for reading it through yaml ?

Yongxuanzhang:
> > > @Yongxuanzhang seems like it needs another rebase 🙏🏼
> > 
> > 
> > Just rebased, sorry for the delay 🙏 I was focusing on something else in the past months
> 
> Did you try with some approach for reading it through yaml ?

Sorry I cannot think of a way which is not error prone. :(
Like I mentioned before, detecting the content from file can lead to unexpected errors.

Since it is an experiment feature, why don't we start with the easier one then iterate?

piyush-garg:
Hey @Yongxuanzhang 

Coming back to this after a long, you can try something like this which we are already using CLI code

```
func getApiVersion(pipelineLocation string, httpClient http.Client) (string, error) {
	b, err := file.LoadFileContent(httpClient, pipelineLocation, file.IsYamlFile(), fmt.Errorf("invalid file format for %s: .yaml or .yml file extension and format required", pipelineLocation))
	if err != nil {
		return "", err
	}
	m := map[string]interface{}{}
	err = yaml.UnmarshalStrict(b, &m)
	if err != nil {
		return "nil", err
	}
	return fmt.Sprint(m["apiVersion"]), nil
}
```

We have this file load function at https://github.com/tektoncd/cli/blob/main/pkg/file/file.go

pkg/cmd/pipeline/sign.go
piyush-garg: Replied here https://github.com/tektoncd/cli/pull/2216#issuecomment-2182139865

It does omit the commented one than in file

vdemeester:
Closing — the chains command has been deprecated (#2534) and is scheduled for removal, making this PR obsolete.

PR #2373, opened 2024-08-13
Re-use printer for multiple elements

### LINKED ISSUE #2091: Tkn bundle list split yamls
### Feature request

It would be a useful feature for us when we use tkn bundle list on an OCI image, that contains multiple tasks/pipelines the output would be a valid merged yaml file. Right now it does not separate the yamls with ---. Therefor if we use tkn bundle list, then pipe it to kubectl apply it only applies the last task/pipeline, not all of them.

### Use case

If it would be implemented we could use bundle to bundle a Pipeline with all of its Tasks into one OCI Image, then we could version control our Pipelines and could easily deploy them from CI with one tkn bundle list piped into kubectl apply.

lcarva:
# Changes

Some printers contain an internal state. The YAMLPrinter, for example, tracks how many objects have been printed. When it prints an object, and it is not the first one to be printed, it prints the YAML document separator, `---`. If a new printer is used when printing each object, then the YAMLPrinter always considers that it is printing the first object. This causes the YAML objects to be muddled together.

This commit makes it so a printer is re-used when printing multiple objects. For conformity, this is also done when a single object is expected to be printed, e.g. get or describe.

Fixes #2091

# Submitter Checklist

review them:

for more details._

# Release Notes

```release-note
Fix printing of multiple YAML documents
```


pkg/cmd/bundle/list.go
```diff
@@ -113,8 +112,13 @@ Caching:
 				Err: cmd.OutOrStderr(),
 			}
 
+			p, err := f.ToPrinter()
```
piyush-garg, line 115: can we keep this also inside `PrintObjects` function, it is getting duplicated at multiple places

pkg/cmd/triggertemplate/list.go
```diff
@@ -87,7 +86,11 @@ or
 			}
 
 			if output != "" {
-				return printer.PrintObject(stream.Out, tts, f)
+				p, err := f.ToPrinter()
```
piyush-garg, line 89: can we keep this also as function like before, it is getting duplicated at multiple places

pkg/cmd/bundle/list.go
lcarva: Notice how `PrintObjects` was called for each item in the list, individually. This means a new instance of the printer is created for each call which causes the issue in the first place. I tried to setup some sort of caching where the printer would only be created once per output format. The unit tests revealed that to be a problematic pattern as it forces a global state between different commands.

Any suggestions?

piyush-garg: ok this makes sense, for now will go with this

pkg/cmd/triggertemplate/list.go
piyush-garg: this comment is invalid based on above comment reply

vinamra28 (approved):

PR #2433, opened 2024-11-18
Use io.ReadFull to read the bundle content

vdemeester:
# Changes

The io.Reader documentation says:

> Read reads up to len(p) bytes into p. It returns the number of bytes
> read (0 <= n <= len(p)) and any error encountered. ... If some data is
> available but not len(p) bytes, Read conventionally returns what is
> available instead of waiting for more.

Read is not guaranteed to fill the data argument. Use io.ReadFull to
fill the buffer.

In some cases (a relatively big bundle), the bundle content was not
completely read and it would fail to parse. Using `io.ReadFull` fixes
the issue.

*Note:* because bundles are cached, one will need to clean their cache
bundle (`~/.tekton/bundles`) to make sure it re-fetches correct
content.

/kind bug

# Submitter Checklist

review them:

for more details._

# Release Notes

```release-note
Fix an issue on `bundle list` command with relatively big bundles that couldn't be parsed (truncated data)
```

### DIFF
--- a/pkg/bundle/reader.go
+++ b/pkg/bundle/reader.go
@@ -116,7 +116,7 @@ func readTarLayer(l v1.Layer) ([]byte, error) {
 	}
 
 	contents := make([]byte, header.Size)
-	if _, err := treader.Read(contents); err != nil && err != io.EOF {
+	if _, err := io.ReadFull(treader, contents); err != nil && err != io.EOF {
 		// We only allow 1 resource per layer so this tar bundle should have one and only one file.
 		return nil, fmt.Errorf("failed to read tar bundle: %w", err)
 	}


vdemeester:
cc @chmouel @piyush-garg @PuneetPunamiya 
We will need to cherry-pick this to previous LTS releases.

vdemeester:
This is related to https://github.com/tektoncd/cli/pull/2433

arewm (approved):
This seems like a reasonable change since the issue was indicative of not reading the entire blob.

pkg/bundle/reader.go
```diff
@@ -116,7 +116,7 @@ func readTarLayer(l v1.Layer) ([]byte, error) {
 	}
 
 	contents := make([]byte, header.Size)
-	if _, err := treader.Read(contents); err != nil && err != io.EOF {
+	if _, err := io.ReadFull(treader, contents); err != nil && err != io.EOF {
```
chmouel, line 119: that's not the same tho, you are going to read the whole content in memeory instead of stream it now....

arewm: The extracted content would have to fit into an etcd, chunk? So there should be a maximum size of the blob, right?

If we want to prevent this from getting too big, we could potentially add validation into `tkn bundle push` to warn/fail if the size of the extracted content is too big.

vdemeester: But yes, there is definitely a follow-up to be done here, but this does fixes some existing issue.

arewm: https://github.com/tektoncd/cli/issues/2434

zregvart: Might want to have guard on `header.Size` not being over 1.5MB (https://etcd.io/docs/v3.5/dev-guide/limit/), might introduce a denial of service via crafted compressed layer (zip bomb)

vinamra28 (approved):

PR #2450, opened 2024-12-20
Fix log prefix display inconsistency

icloudnote:
# Changes

The format displayed on tekton-dashboard is task displayName + step, but the format displayed on pipelinerun is name + step

# Submitter Checklist

review them:

for more details._

# Release Notes


chmouel:
I am afraid this may be breaking people that rely on that output...

chmouel:
maybe a flag like --long to display more info and have the displayname into it?

icloudnote:
> maybe a flag like --long to display more info and have the displayname into it?

good idea

icloudnote:
> maybe a flag like --long to display more info and have the displayname into it?

I have added the `--display-name` flag to format the output logs.

icloudnote:
> Add a new long flag `--display-name` to show pipelinerun logs, when `--display-name=true` , show logs with task display name.

I have add new flag `--display-name` to show pipelinerun logs.

chmouel:
what do you think about --long instead as we can add more verbose output in the future and not necessary have to add a flag for every kind of detailed output

icloudnote:
> what do you think about --long instead as we can add more verbose output in the future and not necessary have to add a flag for every kind of detailed output

For more detailed output, I personally feel that it may not be a good idea to put it in the prefix. After all, prefixes are often used to quickly locate problems and query logs of related steps. The format of `[task_name, step_name]` is sufficient.

The log format of using the `[display_name: step_name]` prefix is ​​also to make the log look more concise and locate problems more quickly. It is bad that pipeline run does not display Display_Name, but Tekton Dashboard displays Display_Name, which looks very good.

waveywaves:
Hi @icloudnote. @chmouel has a point, is it possible to rework this for `--long`. Also the contents of the PR are not tied to a specific an issue so I was thinking that we can update it and make this PR for `--long` and not `--display-name`.

PR #3040, opened 2026-07-19
ci(.github/workflows): enable Codecov coverage reporting

enarha:
Add .codecov.yaml with coverage thresholds and an ignore list for non-source directories, following the same change made in tekton pruner. Upload coverage to Codecov via OIDC alongside the existing artifact upload and PR-comment step.

Assisted-by: Claude Sonnet 5 (via Claude Code)

/kind misc

# Changes

# Submitter Checklist

review them:

for more details._

# Release Notes

```release-note
NONE
```

### DIFF
--- a/.codecov.yaml
+++ b/.codecov.yaml
@@ -1,0 +1,17 @@
+coverage:
+  # Commit status https://docs.codecov.io/docs/commit-status are used
+  # to block PR based on coverage threshold.
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 5%
+    patch:
+      # Disable the coverage threshold of the patch, so that PRs are
+      # only failing because of overall project coverage threshold.
+      # See https://docs.codecov.io/docs/commit-status#disabling-a-status.
+      default: false
+
+ignore:
+  - "vendor/**"
+  - "third_party/**"
--- a/.github/workflows/go-coverage.yml
+++ b/.github/workflows/go-coverage.yml
@@ -1,0 +1,54 @@
+name: Go coverage
+
+permissions:
+  contents: read
+
+on:
+  push:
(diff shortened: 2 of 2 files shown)


.github/workflows/go-coverage.yml
```diff
@@ -0,0 +1,54 @@
+name: Go coverage
+
+permissions:
+  contents: read
+
+on:
```
divyansh42, line 6: @enarha are we not targeting to run this on the PR?

divyansh42 (approved):
Thanks @enarha 
I just added one comment for the clarification.

.github/workflows/go-coverage.yml
enarha: Not at the moment. From the upstream discussion, we agreed to start with on push only (so we do not add comments on the PRs or block PRs). Later we'll add the on PR. For repository which already had on PR enabled, I left it as is.

.codecov.yaml
```diff
@@ -0,0 +1,17 @@
...
+      default:
+        target: auto
+        threshold: 5%
+    patch:
+      # Disable the coverage threshold of the patch, so that PRs are
+      # only failing because of overall project coverage threshold.
+      # See https://docs.codecov.io/docs/commit-status#disabling-a-status.
+      default: false
+
+ignore:
```
pratap0007, line 15: should we ignore  docs, hack, and test directories?

divyansh42: @pratap0007 I think it is already ignored by codecov, as it only checks for .go files.
@enarha Please correct if my understanding is not correct.

enarha: Yes, because what codecov is consuming is the output of the golang tooling.

PR #3068, opened 2026-07-27
feat: add ndjson output format with field selection to list commands

### LINKED ISSUE #2853: Add field selection and NDJSON streaming for list commands
## Summary

AI agents pay per token and lose reasoning capacity with irrelevant output. List commands should support field selection and streaming to minimize context window usage.

### Field Selection
```bash
# Only return specific fields
tkn pipelinerun list --fields name,status,startTime --output json
```

### NDJSON Streaming
```bash
# One JSON object per line, streamable
tkn pipelinerun list --output ndjson
```

This allows agents to process results incrementally without buffering large arrays, and to request only the fields they need.

## Acceptance Criteria

- [ ] `--fields` flag on list commands to select output fields
- [ ] `--output ndjson` for newline-delimited JSON (one object per line)
- [ ] Works with all resource list commands

/kind feature

vikashkumar2020:
# Changes
Fixes: https://github.com/tektoncd/cli/issues/2853
Add `--output ndjson` and `--fields` to every `tkn * list` command so that
automation scripts and AI agents can consume resource lists incrementally and
request only the fields they need.

### `--output ndjson` — newline-delimited JSON streaming

Instead of a single JSON array, each resource is emitted as one JSON object
per line (NDJSON / JSON Lines). Consumers can begin processing the first record
before the last one arrives, and never need to buffer an entire response in
memory.

```bash
tkn pipelinerun list --output ndjson
tkn taskrun list    --output ndjson -n staging
tkn pipeline list   --output ndjson -A
```

### `--fields` — dot-path field projection

Restricts each NDJSON object to the requested fields, specified as a
comma-separated list of dot-separated paths. Shared path prefixes are merged
into a single nested object; unknown paths are silently ignored.

```bash
# Only name + status condition for every PipelineRun
tkn pipelinerun list --output ndjson \
  --fields metadata.name,status.conditions

# Combine with --all-namespaces
tkn taskrun list --output ndjson -A \
  --fields metadata.name,metadata.namespace,status.startTime
```

### Implementation

New helper in `pkg/formatted/ndjson.go`:

| Symbol | Purpose |
|--------|---------|
| `PrintNDJSON(w, obj, fields)` | Converts any `runtime.Object` list to NDJSON via `runtime.DefaultUnstructuredConverter`, then streams one line per item |
| `pickFields(src, fields)` | Dot-path projection — builds a new map with only the requested paths |
| `getNestedField` / `setNestedField` | Recursive helpers for reading/writing dot-separated paths |

Each of the nine `list` commands gains:

* `Fields []string` in its `ListOptions` / `listOptions` struct
* `--fields` flag wired to that slice
* A `case output == "ndjson":` branch (in `switch {}` form to satisfy `gocritic`) that calls `formatted.PrintNDJSON`

The `--output` flag is the existing `cliopts.NewPrintFlags` flag — no new top-level flag required.

### Commands covered

| Command | Package |
|---------|---------|
| `tkn pipeline list` | `pkg/cmd/pipeline` |
| `tkn pipelinerun list` | `pkg/cmd/pipelinerun` |
| `tkn task list` | `pkg/cmd/task` |
| `tkn taskrun list` | `pkg/cmd/taskrun` |
| `tkn customrun list` | `pkg/cmd/customrun` |
| `tkn eventlistener list` | `pkg/cmd/eventlistener` |
| `tkn triggertemplate list` | `pkg/cmd/triggertemplate` |
| `tkn triggerbinding list` | `pkg/cmd/triggerbinding` |
| `tkn clustertriggerbinding list` | `pkg/cmd/clustertriggerbinding` |

(`tkn bundle list` reads from an OCI registry, not a Kubernetes API, and is
intentionally excluded.)

### Tests

`pkg/formatted/ndjson_test.go` covers:

- All-fields output produces one valid JSON object per list item
- Multi-field dot-path projection includes only the requested keys
- Top-level field selection (no dot) works
- Unknown paths are silently ignored; known paths are still present
- Empty list produces empty output

# Submitter Checklist

review them:

for more details._

# Release Notes

```release-note
All `tkn * list` commands now support `--output ndjson` (one JSON object per
line, streamable) and `--fields` (comma-separated dot-path projection, e.g.
`metadata.name,status.conditions`) to reduce token usage for automation and
AI-agent workflows.
```

### DIFF
--- a/pkg/cmd/clustertriggerbinding/list.go
+++ b/pkg/cmd/clustertriggerbinding/list.go
@@ -34,6 +34,7 @@ const (
 
 type listOptions struct {
 	NoHeaders bool
+	Fields    []string
 }
 
 func listCommand(p cli.Params) *cobra.Command {
@@ -78,7 +79,10 @@ or
 				Err: cmd.OutOrStderr(),
 			}
 
-			if output == "name" && tbs != nil {
+			switch {
+			case output == "ndjson":
+				return formatted.PrintNDJSON(stream.Out, tbs, opts.Fields)
+			case output == "name" && tbs != nil:
 				w := cmd.OutOrStdout()
 				for _, pr := range tbs.Items {
 					_, err := fmt.Fprintf(w, "clustertriggerbinding.triggers.tekton.dev/%s\n", pr.Name)
@@ -87,7 +91,7 @@ or
 					}
 				}
 				return nil
-			} else if output != "" {
+			case output != "":
 				p, err := f.ToPrinter()
 				if err != nil {
 					return err
@@ -105,6 +109,7 @@ or
 
 	f.AddFlags(c)
 	c.Flags().BoolVar(&opts.NoHeaders, "no-headers", opts.NoHeaders, "do not print column headers with output (default print column headers with output)")
+	c.Flags().StringSliceVar(&opts.Fields, "fields", opts.Fields, "Comma-separated list of fields to include in output (e.g. metadata.name,status.startTime); only used with --output ndjson")
 	return c
 }
 
--- a/pkg/cmd/clustertriggerbinding/list.go
+++ b/pkg/cmd/clustertriggerbinding/list.go
@@ -59,6 +59,15 @@ or
 		},
 		Example: eg,
 		RunE: func(cmd *cobra.Command, _ []string) error {
+			output, err := cmd.LocalFlags().GetString("output")
+			if err != nil {
+				return errors.New("output option not set properly")
+			}
+
+			if len(opts.Fields) > 0 && output != "ndjson" {
+				return fmt.Errorf("--fields is only supported with --output ndjson")
+			}
+
 			cs, err := p.Clients()
 			if err != nil {
 				return err
@@ -69,18 +78,13 @@ or
 				return fmt.Errorf("failed to list ClusterTriggerBindings: %v", err)
 			}
 
-			output, err := cmd.LocalFlags().GetString("output")
-			if err != nil {
-				return errors.New("output option not set properly")
-			}
-
 			stream := &cli.Stream{
 				Out: cmd.OutOrStdout(),
 				Err: cmd.OutOrStderr(),
 			}
 
 			switch {
-			case output == "ndjson":
+			case output == "ndjson" && tbs != nil:
 				return formatted.PrintNDJSON(stream.Out, tbs, opts.Fields)
 			case output == "name" && tbs != nil:
 				w := cmd.OutOrStdout()
--- a/pkg/cmd/pipeline/list.go
+++ b/pkg/cmd/pipeline/list.go
@@ -77,14 +77,18 @@ func listCommand(p cli.Params) *cobra.Command {
 		},
 		SilenceUsage: true,
 		RunE: func(cmd *cobra.Command, _ []string) error {
-			cs, err := p.Clients()
-			if err != nil {
-				return err
-			}
-
 			output, err := cmd.LocalFlags().GetString("output")
 			if err != nil {
 				return fmt.Errorf("output option not set properly: %v", err)
+			}
+
+			if len(opts.Fields) > 0 && output != "ndjson" {
+				return fmt.Errorf("--fields is only supported with --output ndjson")
+			}
+
+			cs, err := p.Clients()
+			if err != nil {
+				return err
 			}
 
 			ns := p.Namespace()
--- a/pkg/cmd/pipeline/list.go
+++ b/pkg/cmd/pipeline/list.go
@@ -61,6 +61,7 @@ NAME	AGE	LAST RUN	STARTED	DURATION	STATUS
 type ListOptions struct {
 	AllNamespaces bool
 	NoHeaders     bool
+	Fields        []string
 }
 
 func listCommand(p cli.Params) *cobra.Command {
@@ -91,7 +92,13 @@ func listCommand(p cli.Params) *cobra.Command {
 				ns = ""
 			}
 
-			if output != "" {
+			if output == "ndjson" {
+				var pl *v1.PipelineList
+				if err := actions.ListV1(pipelineGroupResource, cs, metav1.ListOptions{}, ns, &pl); err != nil {
+					return fmt.Errorf("failed to list Pipelines from namespace %s: %v", ns, err)
+				}
+				return formatted.PrintNDJSON(cmd.OutOrStdout(), pl, opts.Fields)
+			} else if output != "" {
 				p, err := f.ToPrinter()
 				if err != nil {
 					return err
@@ -108,6 +115,7 @@ func listCommand(p cli.Params) *cobra.Command {
 	f.AddFlags(c)
 	c.Flags().BoolVarP(&opts.AllNamespaces, "all-namespaces", "A", opts.AllNamespaces, "list Pipelines from all namespaces")
 	c.Flags().BoolVarP(&opts.NoHeaders, "no-headers", "", opts.NoHeaders, "do not print column headers with output (default print column headers with output)")
+	c.Flags().StringSliceVar(&opts.Fields, "fields", opts.Fields, "Comma-separated list of fields to include in output (e.g. metadata.name,status.startTime); only used with --output ndjson")
 
 	return c
 }
--- a/pkg/cmd/pipeline/list.go
+++ b/pkg/cmd/pipeline/list.go
@@ -95,7 +95,11 @@ func listCommand(p cli.Params) *cobra.Command {
 			if output == "ndjson" {
 				var pl *v1.PipelineList
 				if err := actions.ListV1(pipelineGroupResource, cs, metav1.ListOptions{}, ns, &pl); err != nil {
-					return fmt.Errorf("failed to list Pipelines from namespace %s: %v", ns, err)
+					scope := fmt.Sprintf("namespace %q", ns)
+					if ns == "" {
+						scope = "all namespaces"
+					}
+					return fmt.Errorf("failed to list Pipelines from %s: %w", scope, err)
 				}
 				return formatted.PrintNDJSON(cmd.OutOrStdout(), pl, opts.Fields)
 			} else if output != "" {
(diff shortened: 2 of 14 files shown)


pkg/cmd/task/list.go
```diff
@@ -80,7 +81,13 @@ func listCommand(p cli.Params) *cobra.Command {
 				ns = ""
 			}
 
-			if output != "" {
+			if output == "ndjson" {
+				var tl *v1.TaskList
+				if err := actions.ListV1(taskGroupResource, cs, metav1.ListOptions{}, ns, &tl); err != nil {
+					return fmt.Errorf("failed to list Tasks from namespace %s: %v", ns, err)
+				}
```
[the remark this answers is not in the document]
vikashkumar2020: updated

vikashkumar2020:
@divyansh42 @pradeepitm12 Can you please take a look at PR, i have fixed the review comments

pkg/cmd/pipeline/list.go
```diff
@@ -91,7 +92,13 @@ func listCommand(p cli.Params) *cobra.Command {
 				ns = ""
 			}
 
-			if output != "" {
+			if output == "ndjson" {
+				var pl *v1.PipelineList
+				if err := actions.ListV1(pipelineGroupResource, cs, metav1.ListOptions{}, ns, &pl); err != nil {
+					return fmt.Errorf("failed to list Pipelines from namespace %s: %v", ns, err)
+				}
```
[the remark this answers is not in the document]
vikashkumar2020: updated

pkg/cmd/clustertriggerbinding/list.go
```diff
@@ -78,7 +79,10 @@ or
 				Err: cmd.OutOrStderr(),
 			}
 
-			if output == "name" && tbs != nil {
+			switch {
+			case output == "ndjson":
```
divyansh42: Unlike other commands in this PR (e.g., `pipelinerun`, `taskrun`, `customrun`) which guard with `&& prs != nil` or similar, this branch passes `tbs` directly without a nil check. If `tbs` is nil, `runtime.DefaultUnstructuredConverter.ToUnstructured(nil)` will panic.
Can we to something like:

```go
case output == "ndjson" && tbs != nil:
    return formatted.PrintNDJSON(stream.Out, tbs, opts.Fields)
```

pkg/cmd/triggerbinding/list.go
```diff
@@ -87,7 +88,10 @@ or
 				Err: cmd.OutOrStderr(),
 			}
 
-			if output == "name" && tbs != nil {
+			switch {
+			case output == "ndjson":
```
divyansh42: This should also guard against `tbs` being nil similar to `clustertriggerbinding`

pkg/cmd/task/list.go
```diff
@@ -80,7 +81,17 @@ func listCommand(p cli.Params) *cobra.Command {
 				ns = ""
 			}
 
-			if output != "" {
+			if output == "ndjson" {
+				var tl *v1.TaskList
+				if err := actions.ListV1(taskGroupResource, cs, metav1.ListOptions{}, ns, &tl); err != nil {
```
divyansh42, line 90: Same concern as above — this makes a second API call instead of reusing the already-fetched list. Can we align with the approach used in the other commands.

pkg/formatted/ndjson.go
```diff
@@ -0,0 +1,117 @@
...
+	if !ok {
+		return nil
+	}
+	return getNestedField(child, parts[1])
+}
+
+// setNestedField sets a value in dst at the given dot-separated path,
+// creating intermediate maps as needed and merging with existing maps.
+func setNestedField(dst map[string]any, val any, path string) {
+	if val == nil {
```
divyansh42: When a field has a legitimate `nil`/`null` value (e.g., `status.completionTime` for a running PipelineRun), this logic drops it entirely from the output. For consumers (especially AI agents parsing the response), there's a semantic difference between:
- `"completionTime": null` → field exists, value is explicitly null (run not finished)
- field absent → field was not requested or does not exist

Can we distinguish between "path not found in source" (should be dropped) vs. "path found but value is nil" (should be emitted as `null`). 

One approach can be:

```go
type fieldResult struct {
    found bool
    value any
}
func getNestedField(src map[string]any, path string) fieldResult { ... }
func setNestedField(dst map[string]any, result fieldResult, path string) {
    if !result.found {
        return
    }
    // proceed even if result.value is nil
    ...
}
```

pkg/formatted/ndjson.go
```diff
@@ -0,0 +1,117 @@
+// Copyright © 2024 The Tekton Authors.
```
divyansh42: This file was written in 2026, the copyright should reflect that:

```suggestion
// Copyright © 2026 The Tekton Authors.
```

pkg/formatted/ndjson_test.go
```diff
@@ -0,0 +1,179 @@
+// Copyright © 2024 The Tekton Authors.
```
divyansh42:
```suggestion
// Copyright © 2026 The Tekton Authors.
```

pkg/cmd/pipeline/list.go
```diff
@@ -108,6 +119,7 @@ func listCommand(p cli.Params) *cobra.Command {
 	f.AddFlags(c)
 	c.Flags().BoolVarP(&opts.AllNamespaces, "all-namespaces", "A", opts.AllNamespaces, "list Pipelines from all namespaces")
 	c.Flags().BoolVarP(&opts.NoHeaders, "no-headers", "", opts.NoHeaders, "do not print column headers with output (default print column headers with output)")
+	c.Flags().StringSliceVar(&opts.Fields, "fields", opts.Fields, "Comma-separated list of fields to include in output (e.g. metadata.name,status.startTime); only used with --output ndjson")
```
divyansh42, line 126: If a user passes `--fields metadata.name` without `--output ndjson` (or with `--output json`), the flag is silently ignored. This could lead to confusion. Can we add validation in the `RunE` function:

```go
if len(opts.Fields) > 0 && output != "ndjson" {
    return fmt.Errorf("--fields is only supported with --output ndjson")
}
```

This should apply to all list commands.

divyansh42 (changes requested):
Thanks @vikashkumar2020. I have added few suggestion, please let me know your thoughts on the same.

Also, the test file only covers the `PrintNDJSON` helper function. There are no unit tests verifying:
- The `--output ndjson` branch is correctly reached in each command's `RunE`
- `--fields` is properly parsed and passed through the options struct
- Error handling in commands that make their own `ListV1` call (pipeline, task)

The existing test infrastructure for other commands (e.g., `pipelinerun/list_test.go`) uses fake clients — adding similar tests for the ndjson path would guard against regression in the wiring logic.

pkg/cmd/clustertriggerbinding/list.go
vikashkumar2020: Changed to case output == "ndjson" && tbs != nil:

pkg/cmd/triggerbinding/list.go
vikashkumar2020: fixed

pkg/cmd/task/list.go
vikashkumar2020: Kept the single ListV1 call in the ndjson branch - the command doesn't pre-fetch for the non-ndjson table path and it delegates to printTaskDetails which does its own fetch, the ndjson branch already has exactly one call and is correct

pkg/formatted/ndjson.go
vikashkumar2020: added fieldResult{found bool, value any}, updated getNestedField to return it, and setNestedField to write even nil values when found=true

pkg/formatted/ndjson.go
vikashkumar2020: fixed

pkg/formatted/ndjson_test.go
vikashkumar2020: fixed

pkg/cmd/pipeline/list.go
vikashkumar2020: Added if len(opts.Fields) > 0 && output != "ndjson" { return error } to all 8 list commands: pipeline, task, pipelinerun, taskrun, customrun, clustertriggerbinding, triggerbinding, triggertemplate, eventlistener

pkg/formatted/ndjson.go
```diff
@@ -0,0 +1,117 @@
...
+
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// PrintNDJSON serialises each item of a Kubernetes list object as a single
+// JSON line (NDJSON / JSON Lines).  When fields is non-empty only those
+// dot-separated paths are included in each output object.
+func PrintNDJSON(w io.Writer, obj runtime.Object, fields []string) error {
+	// Convert the list to unstructured so we can work with raw map[string]any.
+	raw, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
```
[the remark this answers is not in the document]
vikashkumar2020: Replaced with meta.EachListItem — each item is converted and marshalled individually

pkg/cmd/pipeline/list.go
```diff
@@ -91,7 +92,17 @@ func listCommand(p cli.Params) *cobra.Command {
 				ns = ""
 			}
 
-			if output != "" {
+			if output == "ndjson" {
+				var pl *v1.PipelineList
+				if err := actions.ListV1(pipelineGroupResource, cs, metav1.ListOptions{}, ns, &pl); err != nil {
+					scope := fmt.Sprintf("namespace %q", ns)
+					if ns == "" {
```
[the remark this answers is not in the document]
vikashkumar2020: Added TestListPipelineRuns_ndjson, TestListPipelines_ndjson, and TestListTasks_ndjson, each with three sub-tests: valid JSON lines output, --fields narrowing, and --fields without ndjson error

vikashkumar2020:
@divyansh42 addressed the comments

Issue #2850, opened 2026-05-12
Add consistent --output json support to all tkn commands


vdemeester:
## Summary

Not all `tkn` commands support `--output json`. This makes it unreliable for AI agents and scripts that need machine-readable output.

## Current State

- `start --dry-run` supports `--output json/yaml`
- `list` commands use Go template tables, some support `--output` via `PrintObjects`
- `describe` commands have custom formatted output, some support `--output` via `PrintObjectV1`
- `logs`, `delete`, `cancel` have no structured output

## Proposal

1. Add `--output json` and `--output yaml` to **every** command that produces output
2. Auto-detect non-TTY stdout and default to JSON (opt-out with `--output table`)
3. Support `TKN_OUTPUT_FORMAT=json` environment variable for session-wide default
4. Ensure all JSON output goes to stdout, all status/progress messages to stderr

## Commands to update

- [ ] `pipeline list/describe/start/delete`
- [ ] `pipelinerun list/describe/delete/cancel/logs`
- [ ] `task list/describe/start/delete`
- [ ] `taskrun list/describe/delete/cancel/logs`
- [ ] `customrun list/describe/delete`
- [ ] `bundle list/push`
- [ ] `eventlistener list/describe`
- [ ] `triggerbinding list/describe`
- [ ] `triggertemplate list/describe`
- [ ] `clustertriggerbinding list/describe`

/kind feature

PR #3086, opened 2026-08-03
feat: add JSON output to taskrun delete

Debashich:
# Changes

Part of #2850

Add consistent `-o/--output json` support to `taskrun delete`, including machine-readable output of successfully deleted TaskRuns. Interactive confirmation prompts are suppressed when JSON output is requested.

Extend the shared `deleter` package with accessors for successful and related deletions, with corresponding unit tests.

Add handling for TaskRuns associated with running PipelineRuns in JSON output mode, requiring `--force` for deletion.

Add validation for unsupported output formats.

# Sample Output

```json
{
  "deleted": [
    "taskrun-1",
    "taskrun-2"
  ]
}
```

# Submitter Checklist

review them:

* [x] Includes [[tests](https://github.com/tektoncd/community/blob/master/standards.md#principles)](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
* [ ] Run the code checkers with `make check`
* [ ] Regenerate the manpages, docs and go formatting with `make generated`
* [x] Commit messages follow [[commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

*See [[the contribution guide](https://github.com/tektoncd/cli/blob/master/CONTRIBUTING.md)](https://github.com/tektoncd/cli/blob/master/CONTRIBUTING.md)
for more details.*

# Release Notes

```release-note
Add JSON output support to `tkn taskrun delete`.
```

### DIFF
--- a/pkg/cmd/taskrun/delete.go
+++ b/pkg/cmd/taskrun/delete.go
@@ -15,6 +15,7 @@
 package taskrun
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
@@ -92,6 +93,11 @@ or
 				Err: cmd.OutOrStderr(),
 			}
 
+			output, err := cmd.LocalFlags().GetString("output")
+			if err != nil {
+				return err
+			}
+
 			if deleteOpts.TaskName != "" {
 				opts.ParentResource = "Task"
 				opts.ParentResourceName = deleteOpts.TaskName
@@ -122,11 +128,15 @@ or
 				return errs
 			}
 
-			if err := opts.CheckOptions(s, availableTrs, p.Namespace()); err != nil {
-				return err
-			}
+			checkStreams := s
+ 			if output == "json" {
+ 				checkStreams = &cli.Stream{In: strings.NewReader("y\n"), Out: &strings.Builder{}, Err: s.Err}
+ 			}
+ 			if err := opts.CheckOptions(checkStreams, availableTrs, p.Namespace()); err != nil {
+ 				return err
+ 			}
 
-			if err := deleteTaskRuns(s, p, availableTrs, opts); err != nil {
+			if err := deleteTaskRuns(s, p, availableTrs, opts, output); err != nil {
 				return err
 			}
 			return errs
@@ -144,7 +154,7 @@ or
 	return c
 }
 
-func deleteTaskRuns(s *cli.Stream, p cli.Params, trNames []string, opts *options.DeleteOptions) error {
+func deleteTaskRuns(s *cli.Stream, p cli.Params, trNames []string, opts *options.DeleteOptions, output string) error {
 	var numberOfDeletedTr, numberOfKeptTr int
 	cs, err := p.Clients()
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/cmd/taskrun/delete.go
```diff
@@ -212,16 +226,29 @@ func deleteTaskRuns(s *cli.Stream, p cli.Params, trNames []string, opts *options
...
+			}
 		}
 		d.DeleteRelated([]string{opts.ParentResourceName})
 	}
 
+	if output == "json" {
+		result := struct {
+			Deleted []string `json:"deleted"`
+		}{
+			Deleted: append(append([]string(nil), d.SuccessfulRelatedDeletes()...), d.SuccessfulDeletes()...),
```
divyansh42, line 246: Even outside the `--task --keep` no-op case above, whenever both `SuccessfulRelatedDeletes()` and `SuccessfulDeletes()` return nil (via `append([]string(nil), nil...)`), the result is a nil slice and `json.Encoder` serializes it as `null`, not `[]`. For consistent machine-readable output, guard against nil:

```go
deleted := append(d.SuccessfulRelatedDeletes(), d.SuccessfulDeletes()...)
if deleted == nil {
    deleted = []string{}
}
```

pkg/cmd/taskrun/delete.go
```diff
@@ -177,9 +187,13 @@ func deleteTaskRuns(s *cli.Stream, p cli.Params, trNames []string, opts *options
...
 			if !prFinished && opts.ForceDelete {
-				fmt.Fprintf(s.Out, "warning: Taskrun %s related pipelinerun still running.\n", tr.Name)
+				if output == "json" {
+					fmt.Fprintf(s.Err, "warning: Taskrun %s related pipelinerun still running.\n", tr.Name)
+				} else {
+					fmt.Fprintf(s.Out, "warning: Taskrun %s related pipelinerun still running.\n", tr.Name)
+				}
 			}
-			if !prFinished && !opts.ForceDelete {
+			if !prFinished && !opts.ForceDelete && output != "json" {
```
divyansh42, line 196: When `-o json` is used without `--force`, the confirmation prompt is suppressed but the TaskRun is unconditionally appended to `processedTrNames` and deleted. This is an implicit force-delete semantic — a running PipelineRun's TaskRun is deleted without consent or any signal in the JSON output. Deleting such a TaskRun causes the PipelineRun controller to restart it, which is a destructive side-effect that scripts/agents can't anticipate.

Since `-o json` is a new feature targeting machine consumers (issue #2850), the behavior should be strict by default. Require `--force` when the TaskRun is owned by a running PipelineRun:

```go
if !prFinished && !opts.ForceDelete && output == "json" {
    return fmt.Errorf("taskrun %s is owned by a running PipelineRun; use --force to delete", tr.Name)
}
```

pkg/cmd/taskrun/delete.go
```diff
@@ -92,6 +93,11 @@ or
 				Err: cmd.OutOrStderr(),
 			}
 
+			output, err := cmd.LocalFlags().GetString("output")
```
divyansh42, line 96: If a user passes `-o yaml` or `-o table`, the command silently falls through to human-readable text output with no error. Add validation:

```go
if output != "" && output != "json" {
    return fmt.Errorf("unsupported output format %q; supported formats: json", output)
}
```

divyansh42 (changes requested):

divyansh42:
@Debashich do you have any updates on this PR?

divyansh42:
@Debashich could you please add release notes in the PR description and also add a sample output.

Debashich:
@divyansh42  Quick update on what happened after your requested changes:

I implemented all 3 review changes, added the release note + sample output, removed the generic output flags so only json is advertised, ran gofmt, regenerated the docs, and pushed the updated commit.

During testing, one existing TaskRun delete unit test was failing due to a fixture/expected-output mismatch. I reproduced the same failure on the base branch, so I confirmed it wasn't caused by this PR, then fixed the test fixture and amended the commit.

After the latest push, build, lint, unit tests, generated code, multi-arch, CodeQL and k8s-plus-one E2E are passing. The only remaining failure is k8s-oldest E2E.

I checked the full E2E logs, and the failure is actually TestPipelineInteractiveStartE2E, which times out after 10 minutes in the go-expect interactive test path. The PipelineRun itself starts successfully before the test hangs. The TaskRun/Pipeline deletion E2Es are passing.

So I haven't changed anything in the E2E tests/infrastructure since the remaining failure appears unrelated to this PR.

PR #3130, opened 2026-08-11
feat : add consistent exit codes

### LINKED ISSUE #2855: Define consistent exit codes for all tkn commands
## Summary

AI agents rely on exit codes to determine success/failure without parsing output. Currently tkn exit codes are inconsistent and undocumented.

## Proposal

Define and document a consistent exit code scheme:

| Code | Meaning |
|------|---------|
| 0 | Success |
| 1 | General error / command failure |
| 2 | Resource not found |
| 3 | Invalid input / validation error |
| 4 | Timeout |
| 5 | Unauthorized / forbidden |

## Acceptance Criteria

- [ ] Audit all commands for current exit code behavior
- [ ] Implement consistent exit codes across all commands
- [ ] Document exit codes in `tkn --help` or a dedicated doc
- [ ] Ensure errors on stderr include structured info when `--output json` is set (e.g. `{"error": "not found", "code": 2}`)

/kind feature

vikashkumar2020:
# Changes

Fixes : https://github.com/tektoncd/cli/issues/2855
Implemented consistent exit codes across tkn. Added a new pkg/exitcode package defining codes 0–5 (success, general error, not found, invalid input, timeout, unauthorized) with a typed error that propagates through the call stack. All Kubernetes API calls in pkg/actions now classify errors at the boundary via exitcode.FromAPIError(). The root command wires up exitcode.CodeFrom() to drive os.Exit, and adds structured {"error":"…","code":N} JSON output to stderr when --output json is set. A reference doc at docs/exit-codes.md covers the full scheme with examples.
# Submitter Checklist

review them:

for more details._

# Release Notes

```release-note
The consistent exit codes with typed errors, k8s API error classification, and structured JSON error output for --output json is added
```

### DIFF
--- a/docs/exit-codes.md
+++ b/docs/exit-codes.md
@@ -1,0 +1,80 @@
+# tkn Exit Codes
+
+`tkn` uses a consistent set of exit codes so that scripts and CI systems can
+detect success or failure without parsing command output.
+
+## Exit Code Table
+
+| Code | Constant       | Meaning                                    |
+|------|----------------|--------------------------------------------|
+| `0`  | `Success`      | The command completed successfully.        |
+| `1`  | `GeneralError` | Unclassified error or command failure.     |
+| `2`  | `NotFound`     | The requested resource does not exist.     |
+| `3`  | `InvalidInput` | Invalid flag, parameter, or input value.   |
+| `4`  | `Timeout`      | The operation exceeded its deadline.       |
+| `5`  | `Unauthorized` | The request was unauthorized or forbidden. |
+
+Exit code `127` is reserved for plugin execution failures.
+
+## Exit Code `2` – Resource Not Found
+
+`tkn` returns `2` whenever a Kubernetes API call returns an HTTP 404 (Not
+Found). For example:
+
+```bash
+tkn taskrun describe my-missing-run -n default
+# → Error: taskruns.tekton.dev "my-missing-run" not found
+echo $?   # 2
+```
+
+## Exit Code `5` – Unauthorized / Forbidden
+
+`tkn` returns `5` when the server responds with HTTP 401 or 403:
+
+```bash
+tkn pipeline list -n restricted-ns
+# → Error: pipelines.tekton.dev is forbidden: ...
+echo $?   # 5
+```
+
+## Structured Errors with `--output json`
+
+When `--output json` is passed to any command that supports it, errors are
+written to **stderr** as a JSON object instead of a plain-text message:
+
+```bash
+tkn pipelinerun describe missing-run --output json 2>err.json
+cat err.json
+# {"error":"pipelineruns.tekton.dev \"missing-run\" not found","code":2}
+echo $?   # 2
+```
+
+This allows programmatic consumers to parse both the error message and the
+category code without inspecting the human-readable output.
+
+## `--exit-with-error` and PipelineRun logs
+
+`tkn pipelinerun logs --exit-with-error` exits with the PipelineRun's Unix
+status after streaming logs:
+
+| PipelineRun state          | Exit code |
+|----------------------------|-----------|
+| Succeeded                  | `0`       |
+| Failed                     | `1`       |
+| No conditions yet          | `1`       |
+
+> **Note:** The "no conditions" case returns `1` (general error) rather than
+> `2` (not found) because the PipelineRun object exists — it simply has not
+> been evaluated yet.
+
+## Using Exit Codes in Scripts
+
+```bash
+tkn task describe my-task -n default
+case $? in
+  0) echo "Found" ;;
+  2) echo "Task does not exist" ;;
+  5) echo "Permission denied" ;;
+  *) echo "Unexpected error" ;;
+esac
+```
--- a/pkg/cmd/pipelinerun/logs.go
+++ b/pkg/cmd/pipelinerun/logs.go
@@ -136,7 +136,10 @@ func Run(opts *options.LogOptions) error {
 
 func prStatusToUnixStatus(pr *tektonv1.PipelineRun) int {
 	if len(pr.Status.Conditions) == 0 {
-		return 2
+		// PipelineRun has no conditions yet; treat as a general failure
+		// so the caller can distinguish from a successful run (0) without
+		// conflicting with the "resource not found" exit code (2).
+		return 1
 	}
 	if pr.Status.Conditions[0].Status == corev1.ConditionFalse {
 		return 1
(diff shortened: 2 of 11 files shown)


divyansh42:
@vikashkumar2020 thanks for the PR!
Could you please add release notes and link it to the corresponding GitHub issue?

vikashkumar2020:
> @vikashkumar2020 thanks for the PR! Could you please add release notes and link it to the corresponding GitHub issue?

added into description

pkg/exitcode/exitcode.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// FromAPIError converts a Kubernetes API error into an Error with the
+// appropriate exit code.  If err is not a k8s status error it is returned
+// unchanged.
+func FromAPIError(err error) error {
+	if err == nil {
+		return nil
+	}
+	switch {
```
divyansh42, line 73: The `InvalidInput` constant is declared but nothing in this PR ever emits exit code 3. `FromAPIError` doesn't map `k8serrors.IsInvalid()` or `k8serrors.IsBadRequest()` to it, and Cobra flag-parsing errors also bypass this classification entirely.

We can probably do
```go
case k8serrors.IsInvalid(err), k8serrors.IsBadRequest(err):
    return &Error{Code: InvalidInput, Message: err.Error()}
```

pkg/exitcode/exitcode.go
```diff
@@ -0,0 +1,96 @@
...
+	Unauthorized = 5
+)
+
+// Error is an error that carries a specific exit code.
+type Error struct {
+	Code    int
+	Message string
+}
+
+func (e *Error) Error() string {
```
divyansh42, line 57: `FromAPIError` creates a new `*exitcode.Error` with only the string message, discarding the original Kubernetes `*StatusError`. Any downstream code that calls `k8serrors.IsNotFound(err)` or `errors.As(err, &statusErr)` after a `pkg/actions` call will now get `false` because the original error is no longer in the chain.

Can we store the original and implement `Unwrap()`:

```go
type Error struct {
    Code    int
    Message string
    Err     error
}
func (e *Error) Error() string { return e.Message }
func (e *Error) Unwrap() error { return e.Err }
```

And update `FromAPIError` to preserve the original:
```go
return &Error{Code: NotFound, Message: err.Error(), Err: err}
```
This preserves both the exit code semantics and the original error chain

pkg/cmd/root.go
```diff
@@ -122,6 +125,26 @@ func Root(p cli.Params) *cobra.Command {
 	return cmd
 }
 
+// PrintError writes err to errW.  When the resolved --output flag on cmd is
+// "json", the error is serialised as {"error":"<message>","code":<n>}.
+// Otherwise the standard "Error: <message>\n" format is used.
+func PrintError(cmd *cobra.Command, err error, errW *os.File) {
+	outputFlag, _ := cmd.Flags().GetString("output")
```
divyansh42, line 132: `PrintError` calls `cmd.Flags().GetString("output")` where `cmd` is the root `tkn` command (passed from `cmd/tkn/main.go`). However, the `--output` flag is defined on individual subcommands, not on the root. `GetString("output")` on the root will return `""` (and a non-nil error which is silently discarded), so the JSON error path will never trigger.

Consider resolving the executed subcommand's flags instead, or propagating the output format via a persistent pre-run hook that stores it somewhere accessible at error-handling time.

pkg/exitcode/exitcode.go
```diff
@@ -0,0 +1,96 @@
...
+func FromAPIError(err error) error {
+	if err == nil {
+		return nil
+	}
+	switch {
+	case k8serrors.IsNotFound(err):
+		return &Error{Code: NotFound, Message: err.Error()}
+	case k8serrors.IsUnauthorized(err), k8serrors.IsForbidden(err):
+		return &Error{Code: Unauthorized, Message: err.Error()}
+	case k8serrors.IsTimeout(err):
```
divyansh42, line 78: `k8serrors.IsTimeout()` only matches HTTP 504 from the API server. Context deadline exceeded errors (`context.DeadlineExceeded`) and client-side request timeouts won't produce exit code 4.

Can we add this before the `default` case

```go
case errors.Is(err, context.DeadlineExceeded):
    return &Error{Code: Timeout, Message: err.Error(), Err: err}
```

pkg/cmd/pipelinerun/logs.go
```diff
@@ -136,7 +136,10 @@ func Run(opts *options.LogOptions) error {
 
 func prStatusToUnixStatus(pr *tektonv1.PipelineRun) int {
 	if len(pr.Status.Conditions) == 0 {
-		return 2
+		// PipelineRun has no conditions yet; treat as a general failure
+		// so the caller can distinguish from a successful run (0) without
+		// conflicting with the "resource not found" exit code (2).
+		return 1
```
divyansh42, line 142: Changing the "no conditions yet" exit code from `2` to `1` is semantically correct under the new scheme, but it's a breaking change for scripts relying on the previous behavior. The release note should explicitly mention this: "The exit code for PipelineRuns with no conditions has changed from 2 to 1."

docs/exit-codes.md
```diff
@@ -0,0 +1,80 @@
+# tkn Exit Codes
+
+`tkn` uses a consistent set of exit codes so that scripts and CI systems can
+detect success or failure without parsing command output.
+
+## Exit Code Table
+
+| Code | Constant       | Meaning                                    |
+|------|----------------|--------------------------------------------|
+| `0`  | `Success`      | The command completed successfully.        |
```
divyansh42, line 10: In Bash, exit code 2 conventionally means "incorrect usage / invalid arguments." This PR uses 2 for "not found" instead. This isn't necessarily wrong (kubectl doesn't differentiate at all, and many CLIs define their own schemes), but it's worth documenting this divergence explicitly in `docs/exit-codes.md` so users don't confuse it with the shell convention.

divyansh42 (changes requested):
The PR wraps errors in `pkg/actions` (get, list, delete, patch) but many commands produce errors through other paths: direct client calls in trigger/bundle commands, flag validation errors in `RunE`, etc. The issue asks for consistent exit codes "across all commands."

This is acceptable as a first pass, but please file a follow-up issue tracking the remaining commands so this doesn't fall through the cracks.

divyansh42:
One more minor comment:
The new files have `Copyright © 2024` but this PR is authored in 2026. Please update to `2026`.

PR #3201, opened 2026-09-07
fix: prevent unbounded tar layer allocation

anwesha-palit-redhat:
# Changes

Previously, `readTarLayer` used the tar header's declared size directly in `make([]byte, header.Size)`, allowing a malicious bundle to specify an arbitrarily large size and potentially cause the CLI to exhaust available memory and crash.

Updated `readTarLayer` to use a bounded reader with a 1.5 MiB maximum layer size, preventing attacker-controlled tar metadata or oversized layer contents from causing excessive memory allocation. Added regression coverage to verify that layers exceeding the maximum size are rejected.

Signed-off-by: Anwesha Palit [apalit@redhat.com]

Assisted-by: Claude Opus 4.6

# Submitter Checklist

review them:

for more details._

# Release Notes
```release-note
NONE
```

### DIFF
--- a/pkg/bundle/reader.go
+++ b/pkg/bundle/reader.go
@@ -15,6 +15,9 @@ import (
 // Tekton bundle. The `version`, `kind`, and `name` fields map 1:1 with the same named fields in the Tekton Bundle spec.
 type ObjectVisitor func(version, kind, name string, element runtime.Object, raw []byte)
 
+// setting a limit reader to avoid reading more than 1.5 MiB from the layer
+const MaxLayerSize = 15 << 20 / 10
+
 // List will call visitor for every single layer in the img.
 func List(img v1.Image, visitor ObjectVisitor) error {
 	manifest, err := img.Manifest()
@@ -110,15 +113,16 @@ func readTarLayer(l v1.Layer) ([]byte, error) {
 
 	// If the user bundled this up as a tar file then we need to untar it.
 	treader := tar.NewReader(rc)
-	header, err := treader.Next()
-	if err != nil {
+	if _, err := treader.Next(); err != nil {
 		return nil, fmt.Errorf("layer is not a tarball")
 	}
-
-	contents := make([]byte, header.Size)
-	if _, err := io.ReadFull(treader, contents); err != nil && err != io.EOF {
-		// We only allow 1 resource per layer so this tar bundle should have one and only one file.
+	limited := io.LimitReader(treader, MaxLayerSize+1)
+	contents, err := io.ReadAll(limited)
+	if err != nil {
 		return nil, fmt.Errorf("failed to read tar bundle: %w", err)
+	}
+	if int64(len(contents)) > MaxLayerSize {
+		return nil, fmt.Errorf("layer exceeds maximum allowed size of %d bytes", MaxLayerSize)
 	}
 	return contents, nil
 }
--- a/pkg/bundle/builder_test.go
+++ b/pkg/bundle/builder_test.go
@@ -131,14 +131,13 @@ func TestBuildTektonBundle(t *testing.T) {
 
 	// If the user bundled this up as a tar file then we need to untar it.
 	treader := tar.NewReader(rc)
-	header, err := treader.Next()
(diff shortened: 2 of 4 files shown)


pkg/bundle/reader.go
```diff
@@ -110,15 +110,18 @@ func readTarLayer(l v1.Layer) ([]byte, error) {
...
-	if err != nil {
+	if _, err := treader.Next(); err != nil {
 		return nil, fmt.Errorf("layer is not a tarball")
 	}
-
-	contents := make([]byte, header.Size)
-	if _, err := io.ReadFull(treader, contents); err != nil && err != io.EOF {
-		// We only allow 1 resource per layer so this tar bundle should have one and only one file.
+	// setting a limit reader to avoid reading more than 10 MiB from the layer
+	const maxLayerSize = 10 << 20
```
savitaashture: can you mention a comment on choosing this value

anwesha-palit-redhat: it was mentioned in the jira example https://redhat.atlassian.net/browse/SRVKP-13858 
should i update it ?

divyansh42 (approved):
Thanks @anwesha-palit-redhat 
Overall it looks good, I just have a minor comment regarding the test files.

`maxLayerSize` is hardcoded as `10 << 20` in three test files separately from
  the constant in `reader.go`. If the limit in `reader.go` changes,
  the test copies won't catch the drift — no compiler error will surface it.
  Consider exporting the constant (e.g. `MaxLayerSize`) from `reader.go` so
  tests can reference it directly instead of duplicating the value.

pkg/bundle/reader.go
```diff
@@ -15,6 +15,9 @@ import (
 // Tekton bundle. The `version`, `kind`, and `name` fields map 1:1 with the same named fields in the Tekton Bundle spec.
 type ObjectVisitor func(version, kind, name string, element runtime.Object, raw []byte)
 
+// setting a limit reader to avoid reading more than 10 MiB from the layer
```
anwesha-palit-redhat: @divyansh42 @savitaashture 
should i replace this with this comment ?

`We are currently setting the maximum header size to 10MiB, we can update it later if required`

pkg/bundle/reader.go
savitaashture: that was an example

may be can you check once what can be the default value 🤔

anwesha-palit-redhat: where should i check this ?

anwesha-palit-redhat: i could not find any info in tekton which suggests a reasonable header size 😞 for api

anwesha-palit-redhat: @divyansh42 any suggestions please ?

anwesha-palit-redhat: @savitaashture 
https://pkg.go.dev/net/http#Server 
image

not sure if this is relevant here, it is 1MB Go's library, so can we take this as a reference and reduce it to 1 ?

divyansh42: We can probably set this to 1.5 MiB (1536 << 10)

It is tied directly to the Kubernetes etcd object size limit,  no valid Tekton resource can exceed it since Kubernetes would reject it at admission anyway.

pkg/bundle/reader.go
```diff
@@ -15,6 +15,9 @@ import (
 // Tekton bundle. The `version`, `kind`, and `name` fields map 1:1 with the same named fields in the Tekton Bundle spec.
 type ObjectVisitor func(version, kind, name string, element runtime.Object, raw []byte)
 
+// setting a limit reader to avoid reading more than 1.5 MiB from the layer
+const MaxLayerSize = 15 << 20 / 10
```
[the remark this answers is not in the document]
divyansh42: @anwesha-palit-redhat could you please modify the PR description?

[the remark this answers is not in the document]
anwesha-palit-redhat: PR description or the comment ? 🤔

savitaashture:
LGTM

